PR #11, opened 2026-01-27
Dankrad's zones review

dankrad:
## Summary

Code review fixes for zone contracts.

## Changes

- **ZoneMessenger**: Fix misleading 'transient' comment (clashes with EVM transient storage terminology) → changed to 'only valid within a callback'
- **ZoneOutbox**: Add `MAX_GAS_FEE_RATE` constant ($1 per gas) to prevent fee calculation overflow
- **ZoneOutbox**: Add `GasFeeRateTooHigh` error and bounds check in `setWithdrawalFees`  
- **ZonePortal**: Document `sequencerPubkey` purpose (future message encryption for deposits)

### DIFF
--- a/.github/workflows/docs-specs.yml
+++ b/.github/workflows/docs-specs.yml
@@ -2,17 +2,21 @@ name: Docs Specs
 
 on:
   push:
-    branches: [main]
+    branches: [main, dr/validium]
     paths:
-      - "docs/specs/**"
-      - "crates/precompiles/**"
+      - "docs/specs/src/**"
+      - "docs/specs/test/**"
+      - "docs/specs/foundry.toml"
+      - "crates/tempo-zone/**"
       - ".github/workflows/docs-specs.yml"
   merge_group:
   pull_request:
-    branches: [main]
+    branches: [main, dr/validium]
     paths:
-      - "docs/specs/**"
-      - "crates/precompiles/**"
+      - "docs/specs/src/**"
+      - "docs/specs/test/**"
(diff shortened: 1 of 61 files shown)


dankrad:
Important questions:

- I don't think I want to deploy zones using Solidity contracts, they should be precompiles for upgradability (ignore if that is the plan
- I want to enshrine the verified, I think we should provide this and not leave zone operators the freedom to create their own
- I don't like the term gas token, sounds too technical/specified on one aspect, what about "zone token"?

dankrad:
## REVIEWTODOs

- overview.md: EMPTY_SENTINEL should be 0x0 to clear storage vs growth
- overview.md: IZoneGasToken interface not defined in solidity files
- overview.md: Why both xDomain sender checks and msg.sender == zoneMessenger?
- prover-design.md: finalizeTempo system tx vs advanceTempo sequencing contradiction
- prover-design.md: end-of-batch only match; solve past batch via longer batch
- prover-design.md: `is_some()` meaning in batch proof
- prover-design.md: meaning of binding Tempo header to finalized Tempo state
- prover-design.md: advanceTempo only called from system tx assumption
- DepositQueueLib.sol: hash deposit only; how sequencer gets data?
- WithdrawalQueueLib.sol: dislike special gas accounting; find alternative
- WithdrawalQueueLib.sol: why not set remainingQueue to EMPTY_SENTINEL when empty
- IZone.sol: WithdrawalQueueTransition could just be bytes32
- MockZoneGasToken.sol: interface not defined in ZoneInbox.sol

PR #12, opened 2026-01-27
Proposal for gas rate on zone deposit/withdrawals

dankrad:
## Summary

 - Gas on deposits is charged according to `zoneGasRate` set by sequencer in portal contract
 - Gas on withdrawals is charged according to `tempoGasRate` set by sequencer inside zone

## Withdrawals (Zone → Tempo)

- Rename `withdrawalGasFeeRate` to `tempoGasRate`
- Fee formula: `gasLimit * tempoGasRate`
- Sequencer publishes `tempoGasRate` via `setTempoGasRate()`
- Sequencer takes the risk: if actual Tempo gas is higher, they cover; if lower, they profit

## Deposits (Tempo → Zone)

- Add `DEPOSIT_GAS_ESTIMATE` constant (50,000 gas for minting on zone)
- Add `zoneGasRate` to ZonePortal
- Fee formula: `DEPOSIT_GAS_ESTIMATE * zoneGasRate`
- Fee is deducted from deposit amount and paid to sequencer immediately on Tempo
- Deposit queue stores net amount (`amount - fee`)
- Sequencer publishes `zoneGasRate` via `setZoneGasRate()`
- Sequencer takes the risk on zone gas costs

### DIFF
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -73,17 +73,26 @@ This applies to all zone contracts: `ZonePortal` (Tempo-side), `ZoneInbox`, `Zon
 
 ### Deposit fees
 
-> **TODO**: Deposit fee mechanism is undecided. Options include: minimum deposit amount (spam prevention), fixed fee deducted from deposit, or no fee (sequencer absorbs zone-side gas costs). Zone gas should be cheap due to no data availability costs.
+Deposits incur a processing fee to compensate the sequencer for zone-side gas costs:
+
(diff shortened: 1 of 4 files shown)


dankrad:
A question is how we should handle variable gas on deposit. Even simple transfers can have different gas costs due to state creation etc. But the initiating user doesn't necessarily have any idea about this, because they might not even see what's going inside the zone. So I suggest we leave this risk to the sequencer, and they can always protect themselves by increasing `zoneGasRate`

danrobinson:
This approach seems basically reasonable to me

danrobinson (approved):

PR #13, opened 2026-01-28
L1 state declarations

dankrad:
Specify how L1 state reads can work via L1 state declarations. In this version, all transactions have to declare all L1 state they read (keys and values), and fail if the declaration is incomplete or wrong. This means:
1. Transactions cannot initiate unexpected L1 storage reads. This ensures that the sequencer can request all declared state from an RPC at the start of block building and is not dependent on RPC latency during block building
2. L2 state can be reconstructed from history without having to run L1 state in parallel or needing low latency access to an archive node.

The downsides of this design are:
1. It isn't generally possible to sign transactions without access to a zone RPC
   - can be mitigated by predicting TIP-403 access patterns, which is the most common case, and assuming that the account is not blacklisted
2. Wallets need to implement this new transaction type in order to be able to use tempo zones
   - We could special case this by allowing inference of the TIP-403 generated state access pattern allowing TIP-20 transfers without the state declaration

### DIFF
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -678,7 +678,6 @@ struct TempoAccountState {
 
 /// @notice Complete Tempo state declaration for a transaction
 struct TempoStateDeclaration {
-    uint64 tempoBlockNumber;            // Tempo block this declaration is valid for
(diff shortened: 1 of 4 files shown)


dankrad:
Closing in favour of #19

PR #14, opened 2026-01-29
feat(zone): add encrypted deposits for privacy

dankrad:
## Summary

- Add support for encrypted deposits where recipient and memo are encrypted using the sequencer's public key
- Only the sequencer can decrypt and credit the correct recipient on the zone

## Encryption scheme

ECIES with secp256k1:
1. Sequencer publishes secp256k1 encryption key via `setSequencerEncryptionKey()`
2. User generates ephemeral keypair, derives shared secret via ECDH
3. User encrypts `(to || memo)` with AES-256-GCM
4. Portal stores encrypted deposit; only `sender` and `amount` are public

## New types

- `EncryptedDepositPayload`: ephemeral pubkey + ciphertext + nonce + GCM tag
- `EncryptedDeposit`: sender + amount + encrypted payload

## New functions on IZonePortal

- `sequencerEncryptionKey() / setSequencerEncryptionKey()` - manage encryption key
- `depositEncrypted(amount, encryptedPayload)` - make encrypted deposit

## Privacy properties

| Field | Visibility | Reason |
|-------|------------|--------|
| `sender` | Public | Needed for refunds |
| `amount` | Public | Needed for accounting |
| `to` | **Encrypted** | Privacy |
| `memo` | **Encrypted** | Privacy |

## Security note

Encryption is not post quantum. Data might eventually be decipherable by anyone with quantum computers.

### DIFF
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -1034,6 +1034,59 @@ Notes:
 - Tempo state advancement is combined with deposit processing in `ZoneInbox.advanceTempo()`, which calls `TempoState.finalizeTempo()` internally.
 - The proof validates an exact match to `currentDepositQueueHash` from Tempo state, ensuring it cannot claim to process fake deposits. TODO: implement a recursive ancestor check in the proof or on-chain as a fallback.
 
+### Encrypted deposits
+
+For privacy-sensitive use cases, users can make **encrypted deposits** where the recipient and memo are encrypted using the sequencer's public key. Only the sequencer can decrypt and credit the correct recipient on the zone.
+
+**Encryption scheme**: ECIES with secp256k1
+
+1. Sequencer publishes a secp256k1 encryption public key via `setSequencerEncryptionKey(x, yParity)`
+2. User generates an ephemeral keypair and derives a shared secret via ECDH
+3. User encrypts `(to || memo)` with AES-256-GCM using the derived key
+4. User calls `depositEncrypted(amount, encryptedPayload)` on the portal
+
+```solidity
+/// @notice Encrypted deposit payload
+struct EncryptedDepositPayload {
+    bytes32 ephemeralPubkeyX;     // Ephemeral public key X coordinate (for ECDH)
+    uint8 ephemeralPubkeyYParity; // Y coordinate parity (0x02 or 0x03)
+    bytes ciphertext;             // AES-256-GCM encrypted (to || memo || padding)
+    bytes12 nonce;                // GCM nonce
+    bytes16 tag;                  // GCM authentication tag
+}
+
+/// @notice Encrypted deposit stored in the queue
+struct EncryptedDeposit {
+    address sender;              // Depositor (public, for refunds)
+    uint128 amount;              // Amount (public, for accounting)
+    EncryptedDepositPayload encrypted; // Encrypted (to, memo)
+}
+```
+
+**What's public vs. private:**
+
+| Field | Visibility | Reason |
+|-------|------------|--------|
+| `sender` | Public | Needed for potential refunds if decryption fails |
+| `amount` | Public | Needed for on-chain accounting/escrow |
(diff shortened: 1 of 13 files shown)


dankrad:
## Post-Quantum Encryption Options

The current spec uses ECIES with secp256k1, which is not post-quantum secure. Here are options for PQ-resistant encryption:

### Lattice-based KEMs (ML-KEM / Kyber)

| Parameter | Public Key | Ciphertext | Security |
|-----------|-----------|------------|----------|
| ML-KEM-512 | 800 bytes | 768 bytes | ~AES-128 |
| ML-KEM-768 | 1,184 bytes | 1,088 bytes | ~AES-192 |

✅ Fits size constraints (ML-KEM-512 < 1kB)
✅ Very fast decryption (~50μs)
❌ **ZK-unfriendly** - NTT transforms and modular arithmetic are expensive in circuits (~100k-500k constraints vs ~10k for ECDH)

gakonst:
The docs mention that `advanceTempo()` now processes a unified queue with both regular and encrypted deposits, and the sequencer must provide `DecryptionData` for encrypted deposits. However, the updated function signature for `advanceTempo()` isn't included in this PR.

Consider adding the new signature to `IZoneInbox` in `docs/specs/src/zone/IZone.sol` - something like:

```solidity
function advanceTempo(
    uint64 tempoBlockNumber,
    QueuedDeposit[] calldata deposits,
    DecryptionData[] calldata decryptions,  // 1:1 with encrypted deposits
    bytes calldata proof
) external;
```

This would make the interface changes explicit and complete the spec.

danrobinson:
One pattern that I was trying to do with the way that deposits and withdrawals are processed is that almost all the logic happens in the EVM (mostly on the zone), not in the prover/verifier, so the STF is very close to just the EVM. Should we do that here (by having the EVM verify that the decrypted data encrypts to the encryption data using the ephemeral key)? I guess we would still have to add precompiles for ECIES and AES‑GCM, so maybe it's not worth it. Trust your judgment on it

dankrad:
> Should we do that here (by having the EVM verify that the decrypted data encrypts to the encryption data using the ephemeral key)? I guess we would still have to add precompiles for ECIES and AES‑GCM, so maybe it's not worth it. Trust your judgment on it

~~My big concern with that idea would be that it would mean that the encryption key has to be available inside the sequencer EVM, and I feel like that's dangerous! Ideally we would have it in a secure enclave somewhere to absolutely minimize the chances of it leaking, although to begin with it will probably just be a separate process. Having it in the EVM feels like adding too much attack surface as the node is a pretty complex piece of software.~~

~~(This makes me think, ideally we want a way to easily prove correct decryption without leaking the key, so that even the prover will not have to know the raw encryption key. Will check if there is an easy way to do this and if there is, it may be possible to put that part into the EVM)~~

It turns out, AES-GCM has an encryption tag that makes it easy to verify correctness of encryption. We do need an AES-GCM precompile but no elliptic curves for verification. Added to the spec now.

dankrad:
> Consider adding the new signature to `IZoneInbox` in `docs/specs/src/zone/IZone.sol` - something like:

Added now.

dankrad:
@danrobinson ready for review and merge

dankrad:
Fix the problem that the use can encrypt garbage and there is no way for sequencer to decrypt. Then the sequencer can't progress. We actually need to prove the shared key derivation.

dankrad:
> Fix the problem that the use can encrypt garbage and there is no way for sequencer to decrypt. Then the sequencer can't progress. We actually need to prove the shared key derivation.

This is now fixed. The correctness of the shared secret can be verified using the Chaum-Pedersen protocol. Precompiles needed: (1) Chaum-Pedersen (this is becase secp256k1 isn't available using EVM opcodes, otherwise could be implemented in Solididty); (2) AES-GCM (the HKDF is now implemented in Solidity)

Ready for review @danrobinson

PR #15, opened 2026-01-30
feat: Zone interop and Tempo addresses

dankrad:
## Summary

This PR introduces the complete zone interoperability protocol for Tempo, including address format, cross-zone transfer routing, and the SwapAndDepositRouter contract.

### TIP-XXXX: Zone Interoperability Protocol

**Complete zone-to-zone transfer protocol** enabling seamless token transfers across the Tempo ecosystem:

#### 1. Tempo Address Format (TIP-1006 foundation)
- **2-char ASCII prefix**: `t1` (mainnet), `tz` (zone), `tt` (testnet), `tZ` (testnet zone)
- **Base58-encoded payload**: zone ID (0-9 bytes) + raw address (20 bytes) + checksum (4 bytes)
- **Double-SHA256 checksum** for error detection

#### 2. Cross-Zone Transfer Routing
Wallets automatically determine transfer paths:
- **Same network**: Direct transfer
- **Mainnet → Zone**: Deposit (with optional encryption)
- **Zone → Mainnet**: Withdrawal
- **Zone N → Zone M (same token)**: Single withdrawal + deposit callback
- **Zone N → Zone M (different tokens)**: Single withdrawal + swap + deposit callback

#### 3. SwapAndDepositRouter Contract
**Unified router for all cross-zone transfers**:
- Handles both same-token (no swap) and different-token (swap) transfers
- Supports both plaintext and encrypted deposits
- Swaps via StablecoinDEX when tokens differ
- Single transaction UX for users
- ~100k gas for same-token, ~200k gas with swap

#### 4. Privacy Support
- **Encrypted deposits**: Recipient/memo encrypted to destination sequencer
- **Cross-zone encryption**: End-to-end privacy across zones
- **Minimal overhead**: ~5-7k gas for encryption

### ZoneFactory Update
- Reserve zone ID 0 for potential future use
- Zone IDs now start at 1

### Documentation
- `docs/pages/protocol/tips/tip-xxxx-zone-interop.mdx` - Complete zone interop protocol
- `docs/pages/protocol/tips/tip-1006.mdx` - Tempo address format (TIP-1006)

### Smart Contracts  
- `docs/specs/src/zone/SwapAndDepositRouter.sol` - **NEW**: Unified cross-zone transfer router
- `docs/specs/src/zone/ZonePortal.sol` - Simplified, restored to basic cross-zone deposits
- `docs/specs/src/zone/ZoneFactory.sol` - Start zone IDs at 1

## Architecture Highlights

**Single Entry Point**: All cross-zone transfers (with or without swap, with or without encryption) go through `SwapAndDepositRouter` for consistency and composability.

**Gas Efficient**: 
- Same-token transfers: ~100k gas callback
- Different-token transfers: ~200k gas callback (includes swap)
- Encrypted deposits: +5-7k gas overhead

**User Experience**: One transaction signature for any cross-zone transfer, regardless of complexity.


danrobinson:
What do you think about the tradeoffs of these addresses not being as human-readable as something like https://x.com/Wonderland/status/2018317430754344996? If we're already breaking 20-byte-address compatibility for it.

That said, I do think we could have an offchain address scheme that uses DNS for zone addressing but where it maps to these addresses, so maybe this is like if I was demanding that IP addresses be human-readable

dankrad:
> If we're already breaking 20-byte-address compatibility for it.

Can you clarify what this means? What is 20 byte address compatibility? Where is that relevant?

I think implementing a human readable format should be another goal. I am not completely convinced that this is DNS. That will have some downsides in terms of dependencies, robustness etc., but especially if we do use DNS, it may be even more important to have an internal address format that is very clear.

My reasoning for it is this: The Ethereum L2 roadmap is partially broken because there is no easy way of seeing the system as a whole. At the interop layer, there are no Ethereum-specific network effects or moats; to the user, Ethereum L1 feels the same as Arbitrum feels the same as Gnosis chain or Avalanche.

When I go to a crypto shop in South America, they know how to handle Tron, Bitcoin and Ethereum. I can't receive on Arbitrum. They don't have an option to select it. They don't know what it is. But ideally I think they should never have to know what it is but just enter an Ethereum address as they already know it and it would land in my Arbitrum account

I want to prevent the fragmentation from the start. For that we:
1. Need a way to specify addresses anywhere on Tempo. A bipartite (zone, address) is a possibility but it comes with an extra risk -- interfaces that don't add the zone because they expect everything to be on L1. Interoperability will be broken if that happens (think exchanges implementing Tempo as another EVM-chain with standard addressing. In other words, using existing 20 byte Ethereum addresses might *increase* the risk of breaking interop in practice)
2. An effort to immediately make it possible to transfer to any zone. Part 1 only works if it is actually easily possible to transfer to any such address.

I think not doing it creates a super strong risk of fragmentation.

Alternatives
 * [ERC-7828](https://ethereum-magicians.org/t/erc-7828-chain-specific-addresses-using-ens/21930) chain-specific addresses, relying on ENS for human readability. One disagreement I have about this standard is that it just feels as if an account "should" just be available on all chains.
 * Some new DNS-style scheme. I think this is overall a much bigger effort, and can well still be added at a later date after we already have an addressing scheme. And for users it is always a good fallback to have at the edges of a scheme that likely requires major integrations and may not always be available. It can also be reassuring to users to have an additional "machine" format unique address rather than just peter@company-rollup.com which can feel too scary to just send large amounts of money to.

PR #17, opened 2026-02-03
feat(zone): L1 state subscriptions

dankrad:
Add subscription-based alternative to transaction-level state declarations.

Key changes:
- Add L1StateSubscriptionManager predeploy for managing subscriptions
- Add L1StateAccessLog structure for replay/reconstruction
- Update IZone.sol with new interfaces and data structures
- Rename IZoneGasToken to IZoneToken for consistency
- Create TempoStateSubscription variant with subscription validation
- Add comprehensive documentation of subscription architecture

Benefits over state declarations:
- Lower user friction (subscribe once vs declare per tx)
- Works with standard Ethereum transactions (no new tx types)
- Automatic TIP-403 policy subscription for zone token transfers
- Complete audit trail via L1StateAccessLog

Sequencer responsibilities:
- Maintain real-time L1 state cache for subscribed slots
- Set monthly subscription fees to cover sync costs
- Auto-subscribe TIP-403 policy state as needed
- Include L1StateAccessLog in batch proofs

PR #19, opened 2026-02-04
feat(zone): centralize L1 sequencer management and restrict Tempo state reads

dankrad:
## Summary

Centralizes sequencer management on L1 (Tempo) via a new ZoneConfig system contract and restricts Tempo state access to only zone system contracts. This eliminates duplicate sequencer logic across zone contracts and prevents user transactions from reading arbitrary L1 state.

### ZoneConfig System Contract
- **New predeploy at 0x1c00...0003**: Centralizes zone configuration and L1 state access
- **Single source of truth**: Reads sequencer from L1 ZonePortal via TempoState
- **Provides**: `sequencer()`, `pendingSequencer()`, `sequencerEncryptionKey()`, `isSequencer()`
- **Benefits**: Eliminates ~100 lines of duplicate sequencer management per contract

### Updated Zone System Contracts
- **ZoneInbox & ZoneOutbox**: Removed local sequencer storage, now use `config.sequencer()`
- **TempoState**: Removed sequencer management, added ZoneConfig to system contract allowlist
- **Access control consistency**: Added explicit `onlyZoneInbox` check to `TempoState.finalizeTempo()`

### Tempo State Access Restrictions
- **TempoState.sol**: `onlySystemContract` modifier restricts `readTempoStorageSlot()` and `readTempoStorageSlots()` to:
  - `ZONE_INBOX` (0x1c00...0001)
  - `ZONE_OUTBOX` (0x1c00...0002)
  - `ZONE_CONFIG` (0x1c00...0003)
- **TempoState.finalizeTempo()**: Only callable by `ZONE_INBOX` for consistency with other system contracts

### Documentation
- **IZone.sol**: Added system contract constants, IZoneConfig interface, updated all interfaces
- **overview.md**: Documented ZoneConfig predeploy, L1-based sequencer management, access restrictions

## Rationale

**Centralized L1 management**: L1 ZonePortal is the single source of truth for sequencer. Changes on L1 become visible to zone contracts after Tempo block finalization. This eliminates the risk of zone-side and L1-side sequencer state diverging.

**Access control consistency**: All zone system contracts now use explicit `msg.sender` checks with the same pattern, making the security model clear and uniform.

**Simplified state access**: User transactions cannot directly read arbitrary L1 state, reducing complexity while maintaining necessary system operations.

## Related

- Incorporates sequencer centralization from #17
- Builds on Tempo state verification infrastructure from #13
- Separate from encrypted deposits work

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -701,7 +701,7 @@ interface ITempoState {
 
     /// @notice Finalize a Tempo block header. Only callable by ZoneInbox.
     /// @dev Validates chain continuity (parent hash must match, number must be +1).
-    ///      Called by ZoneInbox.advanceTempo(). Protocol enforces advanceTempo() runs as system transaction.
+    ///      Called by ZoneInbox.advanceTempo(). Executor enforces ZoneInbox-only access.
     /// @param header RLP-encoded Tempo header
     function finalizeTempo(bytes calldata header) external;
 
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -58,12 +58,12 @@ The factory deploys a `ZonePortal` that escrows the zone token on Tempo. The zon
 
 ### Sequencer transfer
 
-The sequencer can transfer control to a new address via a two-step process:
+The sequencer can transfer control to a new address via a two-step process on **Tempo L1 only**:
 
-1. Current sequencer calls `transferSequencer(newSequencer)` to nominate a new sequencer
-2. New sequencer calls `acceptSequencer()` to accept the transfer
+1. Current sequencer calls `ZonePortal.transferSequencer(newSequencer)` to nominate a new sequencer
+2. New sequencer calls `ZonePortal.acceptSequencer()` to accept the transfer
 
-This applies to all zone contracts: `ZonePortal` (Tempo-side), `ZoneInbox`, `ZoneOutbox`, and `TempoState` (zone-side). The two-step pattern prevents accidental transfers to incorrect addresses.
+**Sequencer management is centralized on L1 (Tempo).** Zone-side system contracts (`ZoneInbox`, `ZoneOutbox`) read the sequencer from L1 via `ZoneConfig`, which queries `TempoState` to get the sequencer address from the finalized `ZonePortal` storage. This eliminates duplicate sequencer management logic and ensures L1 is the single source of truth. The two-step pattern prevents accidental transfers to incorrect addresses.
 
 ## Execution and fees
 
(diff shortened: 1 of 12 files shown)


dankrad:
@danrobinson This implements the alternative approach where only system contracts can read from L1 state. Ready for review.

docs/pages/protocol/privacy/overview.md
```diff
@@ -667,12 +694,22 @@ The TempoState stores the Tempo wrapper fields and the inner fields needed by th
 
 **How it works:**
 
-1. The sequencer submits Tempo block headers via `finalizeTempo()`, which decodes the RLP header, validates chain continuity, and stores the wrapper fields and selected inner fields.
+1. ZoneInbox calls `finalizeTempo()` at the start of each block, which decodes the RLP header, validates chain continuity, and stores the wrapper fields and selected inner fields.
```
danrobinson: Why at the start of each block? There isn't necessarily a new Tempo block for every Zone block. And the idea here was that the sequencer could do it as soon as the Tempo block is ready.

docs/pages/protocol/privacy/overview.md
```diff
@@ -667,12 +694,22 @@ The TempoState stores the Tempo wrapper fields and the inner fields needed by th
...
+
+**Access restrictions:**
+
+Tempo state reads are restricted to zone system contracts (ZoneInbox, ZoneOutbox, ZoneConfig) to maintain simplicity and security:
+- **ZoneConfig** reads ZonePortal storage for sequencer address (L1 is single source of truth)
+- **ZoneInbox** reads ZonePortal storage for deposit queue hash
+- **ZoneOutbox** may read ZonePortal storage for withdrawal processing
+- **TIP-403 enforcement** is handled by TIP-20 transfers reading policy state indirectly through system contracts
+
+User transactions **cannot** directly read Tempo state. This eliminates the complexity of state subscriptions and per-transaction state declarations while still allowing the zone to maintain its view of critical Tempo state for system operations.
```
danrobinson: There's a lot of redundancy in this section that I think can be eliminated; like this paragraph is just repeating the other ones.

docs/pages/protocol/privacy/prover-design.md
```diff
@@ -127,10 +127,7 @@ The system transactions must mirror the Solidity reference implementation:
 - `ZoneInbox.advanceTempo(tempo_header_rlp, deposits)` at the start of the block
 - `ZoneOutbox.finalizeWithdrawalBatch(count)` at the end of the block **only if this is the final block of the batch**
 
-User transactions may call the `TempoState` precompile to read Tempo state. User transactions
-**must not** call `ZoneInbox` or `ZoneOutbox`; those are system-only predeploys. The executor
-must reject any non-system call to these addresses and enforce exactly one `advanceTempo`
-at the start of each block and exactly one `finalizeWithdrawalBatch` in the final block of the batch.
+User transactions **must not** call the system contract predeploys (`TempoState`, `ZoneInbox`, `ZoneOutbox`, `ZoneConfig`). The executor must reject any non-system call to these addresses and enforce exactly one `advanceTempo` at the start of each block and exactly one `finalizeWithdrawalBatch` in the final block of the batch. Tempo state reads via `TempoState` are restricted to system contracts only.
```
danrobinson: I'm actually not sure why this logic belonged in the prover design anyway; it's all enforced in the EVM. I think can drop this paragraph

docs/specs/src/zone/IZone.sol
```diff
@@ -280,22 +300,15 @@ struct LastBatch {
 /// @title ITempoState
 /// @notice Interface for zone-side Tempo state verification predeploy
 /// @dev Deployed at 0x1c00000000000000000000000000000000000000
+///      System-only contract. Only ZoneInbox can call finalizeTempo().
```
danrobinson, line 329: IMO should have sequencer do this

docs/specs/src/zone/IZone.sol
```diff
@@ -369,21 +376,9 @@ interface IZoneInbox {
...
     /// @notice The zone's last processed deposit queue hash
     function processedDepositQueueHash() external view returns (bytes32);
 
-    /// @notice Start a sequencer transfer. Only callable by current sequencer.
-    function transferSequencer(address newSequencer) external;
-
-    /// @notice Accept a pending sequencer transfer. Only callable by pending sequencer.
-    function acceptSequencer() external;
-
     /// @notice Advance Tempo state and process deposits in a single system transaction.
```
danrobinson: I think it was outdated that this spec called this a system transaction; in my design it was just a regular transaction from the sequencer. In part because it's not guaranteed that it will be called every block.

docs/specs/src/zone/IZone.sol
dankrad: The sequencer calls `advanceTempo()`, which then calls `finalizeTempo()`. I think this is now correctly implemented.

docs/specs/src/zone/IZone.sol
dankrad: Right, I guess it's not a system transaction in the Tempo sense, but it is a special transaction because it is enforced by the prover that it is only at the start of the block. The witnesses are currently configured that way so I think it's ok to keep this design since it simplifies things.

docs/pages/protocol/privacy/prover-design.md
dankrad: Not all of these are enforced by the EVM, e.g. advanceTempo only at top of block. Simplified now.

docs/pages/protocol/privacy/overview.md
dankrad: Right, not necessarily each block. However it has to be at the start of the block if your second sentence implied it could be anywhere.

PR #58, opened 2026-02-08
feat: add stub verifier system contract (always returns true)

### LINKED ISSUE #54: design: enshrine the verifier instead of abstracting behind IVerifier
## Source

PR #11 comment by @dankrad

## Description

> I want to enshrine the verifier, I think we should provide this and not leave zone operators the freedom to create their own

The verifier is currently abstracted behind a minimal \`IVerifier\` interface, allowing zone operators to provide their own verification implementation:

\`\`\`solidity
interface IVerifier {
    function verifyBatch(
        bytes calldata publicInputs,
        bytes calldata verifierConfig,
        bytes calldata proof
    ) external view returns (bool);
}
\`\`\`

The suggestion is to enshrine a specific verifier implementation (e.g., SP1 ZKVM verifier or TEE attestation verifier) as part of the protocol, rather than leaving it pluggable.

## Considerations

- Enshrining reduces the attack surface — operators can't use a weak/broken verifier
- Standardization simplifies auditing and trust assumptions
- May reduce flexibility for future proving systems
- Could enshrine multiple verifier types (ZK + TEE) while still constraining the set
- The \`verifierConfig\` field could be replaced with protocol-defined parameters

## Location

- \`docs/specs/src/zone/IZone.sol:276-315\` (IVerifier interface)
- \`docs/specs/src/zone/ZonePortal.sol\` (verifier usage in submitBatch)

gakonst:
## Summary
Enshrines the verifier as a system contract (precompile) per #54. Stub always returns true for prototyping.

## Motivation
[Issue #54](https://github.com/tempoxyz/zones/issues/54) — enshrine the verifier instead of abstracting behind IVerifier, reducing attack surface from operator-chosen verifiers.

## Changes
- `crates/contracts/src/precompiles/verifier.rs` — IVerifier Solidity interface definition via `sol!` macro
- `crates/contracts/src/precompiles/mod.rs` — register module + `VERIFIER_ADDRESS` at `0x1E81F1E8...`
- `crates/precompiles/src/verifier/mod.rs` — stub implementation (`verify()` always returns true) + `initialize()`
- `crates/precompiles/src/verifier/dispatch.rs` — call dispatch matching IVerifier interface
- `crates/precompiles/src/lib.rs` — register VerifierPrecompile in `extend_tempo_precompiles`
- `xtask/src/genesis_args.rs` — initialize verifier at genesis

## Testing
```
cargo test -p tempo-precompiles -- verifier  # 4 tests pass
cargo clippy -p tempo-contracts -p tempo-precompiles -p tempo-xtask -- -D warnings  # clean
```

Slack thread: https://tempoxyz.slack.com/archives/C0A87C21805/p1770586772987439

Closes #54

### DIFF
--- a/crates/contracts/src/precompiles/mod.rs
+++ b/crates/contracts/src/precompiles/mod.rs
@@ -7,6 +7,7 @@ pub mod tip20_factory;
 pub mod tip403_registry;
 pub mod tip_fee_manager;
 pub mod validator_config;
+pub mod verifier;
 
 pub use account_keychain::*;
 use alloy_primitives::{Address, address};
@@ -18,6 +19,7 @@ pub use tip20::*;
 pub use tip20_factory::*;
 pub use tip403_registry::*;
 pub use validator_config::*;
+pub use verifier::*;
 
 pub const TIP_FEE_MANAGER_ADDRESS: Address = address!("0xfeec000000000000000000000000000000000000");
 pub const PATH_USD_ADDRESS: Address = address!("0x20C0000000000000000000000000000000000000");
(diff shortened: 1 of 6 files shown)


gakonst:
Closing — will redo as a Solidity spec contract instead of a Rust precompile.

PR #81, opened 2026-02-09
docs(zone): document TempoState header canonicality trust assumption

### LINKED ISSUE #75: LOW: TempoState header canonicality trust assumption undocumented
## Summary

`TempoState.finalizeTempo()` validates only parent-hash continuity and block number increment. It does not verify the header is canonical Tempo consensus output. This is presumably enforced by the zone node/prover/TEE, but the trust assumption is not documented.

A malicious sequencer could feed fake headers to set an arbitrary `tempoStateRoot`, making `readTempoStorageSlot()` return attacker-controlled values, bypassing deposit hash checks, and minting arbitrary zone tokens.

## Location

- `docs/specs/src/zone/TempoState.sol#L88-L104`

## Fix

Add explicit documentation of the trust boundary: *"`finalizeTempo(header)` is only accepted if the header is finalized/canonical; this is enforced by the zone node / proof system, not by on-chain validation."*

## Severity

**LOW** — the trust assumption is likely correct but should be explicit in the spec.

gakonst:
## Summary
Makes explicit the trust assumption that finalizeTempo() relies on the zone node/proof system to enforce header canonicality. On-chain validation only checks chain continuity (parent hash and block number increment).

Closes #75

### DIFF
--- a/docs/specs/src/zone/TempoState.sol
+++ b/docs/specs/src/zone/TempoState.sol
@@ -8,6 +8,9 @@ import { ITempoState } from "./IZone.sol";
 /// @dev Deployed at 0x1c00000000000000000000000000000000000000
 ///      Stores the latest finalized Tempo block info. Sequencer submits Tempo headers
 ///      which are validated for chain continuity and decoded to update state.
+///      TRUST BOUNDARY: On-chain validation only checks chain continuity (parent hash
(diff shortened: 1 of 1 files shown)


dankrad:
Turns out this isn't a real issue: https://github.com/tempoxyz/zones/issues/75#issuecomment-3872121327

PR #82, opened 2026-02-09
fix(zone): enforce monotonic tempoBlockNumber in submitBatch

### LINKED ISSUE #66: HIGH: No monotonic enforcement on tempoBlockNumber in submitBatch
## Summary

`ZonePortal.submitBatch()` sets `lastSyncedTempoBlockNumber = tempoBlockNumber` unconditionally without checking that it does not regress. If the verifier has a bug, a batch could sync to an older Tempo block, causing confusion about which deposits have been processed.

## Location

- `docs/specs/src/zone/ZonePortal.sol#L627`

## Fix

Add defense-in-depth check:

```solidity
require(tempoBlockNumber >= lastSyncedTempoBlockNumber, "tempoBlockNumber regression");
```

## Severity

**HIGH** — relies entirely on verifier correctness with no local guardrail against state regression.

gakonst:
## Summary
Adds defense-in-depth check that tempoBlockNumber cannot regress in submitBatch(). Without this, a verifier bug could allow syncing to an older Tempo block.

Closes #66

### DIFF
--- a/docs/specs/src/zone/ZonePortal.sol
+++ b/docs/specs/src/zone/ZonePortal.sol
@@ -582,6 +582,7 @@ contract ZonePortal is IZonePortal {
 
         // Validate tempoBlockNumber is valid (applies to both direct and ancestry modes)
         if (tempoBlockNumber < genesisTempoBlockNumber) revert InvalidTempoBlockNumber();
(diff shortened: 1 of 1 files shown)


dankrad:
Closing as relying on verifier correctness is fine for this.

PR #84, opened 2026-02-09
fix(zone): allow chunked deposit processing in advanceTempo

### LINKED ISSUE #63: CRITICAL: Deposit processing requires full catch-up — liveness kill via deposit spam
## Summary

`ZoneInbox.advanceTempo()` enforces strict equality between the computed deposit hash and Tempo's `currentDepositQueueHash`. This means every call must process **all** pending deposits in one shot. Any party can spam cheap deposits on L1 to make this unbounded loop exceed zone block gas limits, permanently halting the zone.

## Location

- `docs/specs/src/zone/ZoneInbox.sol#L293`

## Details

```solidity
if (currentHash != tempoCurrentHash) {
    revert InvalidDepositQueueHash();
}
```

The doc has multiple TODOs acknowledging this: *"TODO: implement a recursive ancestor check"*. This is not optional — it is the primary liveness requirement. Without chunked processing + ancestor proofs, the system is trivially DoS-able.

Encrypted deposits make this worse since each one requires Chaum-Pedersen verification + HKDF + AES-GCM decryption, significantly increasing per-deposit cost.

## Fix

Allow `advanceTempo()` to process a bounded subset of deposits and update `processedDepositQueueHash` to an intermediate value. The proof must validate contiguity (ancestor check) rather than exact equality against `currentDepositQueueHash`.

## Severity

**CRITICAL** — any party can permanently halt a zone by spamming L1 deposits.

gakonst:
## Summary
Removes the strict equality requirement between processedDepositQueueHash and Tempo's currentDepositQueueHash in ZoneInbox.advanceTempo(). The proof/TEE validates that the processed hash is a contiguous prefix (ancestor-or-equal) of Tempo's current hash.

This prevents a liveness attack where an adversary spams L1 deposits to make advanceTempo() exceed zone block gas limits.

Closes #63

### DIFF
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -129,7 +129,7 @@ struct BlockTransition {
 
 /// @notice Deposit queue transition inputs/outputs for batch proofs
 /// @dev The proof reads currentDepositQueueHash from Tempo state to validate
-///      that nextProcessedHash matches currentDepositQueueHash for now. TODO: allow ancestor checks.
+///      that nextProcessedHash is an ancestor of (or equal to) currentDepositQueueHash.
 struct DepositQueueTransition {
     bytes32 prevProcessedHash;     // where proof starts (verified against zone state)
     bytes32 nextProcessedHash;     // where zone processed up to (proof output)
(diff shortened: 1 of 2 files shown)


dankrad:
Going for this instead: https://github.com/tempoxyz/zones/pull/91

PR #88, opened 2026-02-09
docs(zone): document ZoneFactory nonce tracking fragility

### LINKED ISSUE #73: LOW: ZoneFactory manual nonce tracking is fragile
## Summary

`ZoneFactory._deploymentNonce` manually tracks the contract nonce for CREATE address prediction. Adding any other `new` statement to the factory (even in a future function) will desync this counter and permanently break `createZone()`. The `require(portal == predictedPortal)` check will catch it at runtime but the factory becomes unusable.

## Location

- `docs/specs/src/zone/ZoneFactory.sol#L23`

## Fix

Use `CREATE2` for at least one of {portal, messenger} to eliminate nonce tracking entirely.

## Severity

**LOW** — no current exploit but fragile for future maintenance.

gakonst:
## Summary
Adds prominent warning comments about the fragile manual nonce tracking in ZoneFactory. The current CREATE-based deployment prediction will break if any additional contracts are deployed without updating the nonce counter.

A future improvement would be to migrate to CREATE2, but this requires resolving the circular dependency between ZoneMessenger and ZonePortal constructor arguments.

Closes #73

### DIFF
--- a/docs/specs/src/zone/ZoneFactory.sol
+++ b/docs/specs/src/zone/ZoneFactory.sol
@@ -19,6 +19,10 @@ contract ZoneFactory is IZoneFactory {
     mapping(address => bool) internal _isZonePortal;
 
     /// @notice Tracks deployment count for CREATE address prediction
+    /// @dev WARNING: This counter must be incremented by exactly the number of CREATE operations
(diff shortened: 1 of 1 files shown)


dankrad:
Not fixing this for now

PR #89, opened 2026-02-09
fix(zone): compact pending withdrawals array to prevent unbounded growth

### LINKED ISSUE #72: MEDIUM: ZoneOutbox _pendingWithdrawals array has unbounded storage growth
## Summary

`ZoneOutbox._pendingWithdrawals` uses a growing array with a logical head pointer (`_pendingWithdrawalsHead`). Between full resets, the array and head can grow unboundedly. If `finalizeWithdrawalBatch` is called with `count < pending`, the head advances but old slots are only `delete`d individually (zeroed but array length stays), consuming storage slots.

## Location

- `docs/specs/src/zone/ZoneOutbox.sol#L152` — `push()`
- `docs/specs/src/zone/ZoneOutbox.sol#L208` — individual `delete`
- `docs/specs/src/zone/ZoneOutbox.sol#L216-L219` — full reset only when head catches up

## Fix

Consider a ring buffer or periodic compaction. This is a zone-side gas concern (not L1), but could become expensive over time.

## Severity

**MEDIUM** — unbounded zone-side storage growth.

gakonst:
## Summary
Removes `_pendingWithdrawalsHead` and always compacts the `_pendingWithdrawals` array after `finalizeWithdrawalBatch()`. This prevents unbounded storage growth from partial batch processing.

## Changes
- Removed `_pendingWithdrawalsHead` state variable
- After processing withdrawals, remaining entries are shifted to the front and the array is shrunk via `pop()`
- Simplified `pendingWithdrawalsCount()` to return `_pendingWithdrawals.length`

## Testing
`forge build` passes with no new errors.

Closes #72

### DIFF
--- a/docs/specs/src/zone/ZoneOutbox.sol
+++ b/docs/specs/src/zone/ZoneOutbox.sol
@@ -55,7 +55,6 @@ contract ZoneOutbox is IZoneOutbox {
 
     /// @notice Pending withdrawals waiting to be batched
     Withdrawal[] internal _pendingWithdrawals;
-    uint256 internal _pendingWithdrawalsHead;
 
     /*//////////////////////////////////////////////////////////////
                                 ERRORS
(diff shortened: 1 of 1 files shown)


dankrad:
Don't want to fix this now

PR #98, opened 2026-02-10
feat: implement deposit flow

0xKitsune:


crates/tempo-zone/src/node.rs
```diff
@@ -15,49 +25,85 @@ use reth_node_api::{
...
     validator::TempoTransactionValidator,
 };
 
+sol! {
+    function mint(address to, uint256 amount);
+}
+
+/// TEMPO TIP-20 token address.
+const TEMPO_TOKEN_ADDRESS: alloy_primitives::Address =
+    alloy_primitives::address!("0x20C0000000000000000000000000000000000000");
```
0xKitsune: This should be read from the L1 portal contract, not hardcoded.

crates/tempo-zone/src/node.rs
```diff
@@ -387,6 +435,301 @@ where
...
+{
+    type Attributes = TempoPayloadBuilderAttributes;
+    type BuiltPayload = EthBuiltPayload<TempoPrimitives>;
+
+    fn try_build(
+        &self,
+        args: BuildArguments<Self::Attributes, Self::BuiltPayload>,
+    ) -> Result<BuildOutcome<Self::BuiltPayload>, PayloadBuilderError> {
+        let BuildArguments {
+            mut cached_reads,
```
0xKitsune: Lets move this to a separate mod for the builder logic.

mattsse (approved):
cool, this is pretty good already, and has the missing piece we need to demo https://github.com/tempoxyz/zones/pull/101 (the actual sequencer impl)

I'd like to merge this pr here first, then reanchor #101 and then make more incremental changes while we start testing

PR #103, opened 2026-02-11
feat: subscribe to new blocks and include block header with deposits

0xKitsune:


crates/tempo-zone/src/l1.rs
```diff
@@ -84,68 +85,91 @@ impl L1Subscriber {
...
-                warn!(error = %e, "Failed to process L1 log");
+        while let Some(header) = stream.next().await {
+            let block_number = header.number;
+
+            // Fetch deposit logs for this specific block
+            let filter = Filter::new()
+                .address(self.config.portal_address)
+                .event_signature(DepositMade::SIGNATURE_HASH)
+                .from_block(BlockNumberOrTag::Number(block_number))
+                .to_block(BlockNumberOrTag::Number(block_number));
```
0xKitsune: We dont need to add the block number to the filter. Also we only need to init the filter once since it will be the same each time.

crates/tempo-zone/src/l1.rs
```diff
@@ -84,68 +85,91 @@ impl L1Subscriber {
...
+                Ok(logs) => logs,
+                Err(e) => {
+                    warn!(
+                        error = %e,
+                        block = block_number,
+                        "Failed to fetch logs for block, skipping"
+                    );
+                    continue;
+                }
+            };
```
0xKitsune: We cant fail and continue, we need to make sure to process every block.

crates/tempo-zone/src/l1.rs
```diff
@@ -84,68 +85,91 @@ impl L1Subscriber {
...
                 debug!(
                     l1_block = block_number,
                     sender = %deposit.sender,
                     to = %deposit.to,
                     amount = %deposit.amount,
                     memo = %deposit.memo,
-                    "Received deposit from L1"
+                    "Deposit from L1"
                 );
+            }
```
0xKitsune: Log the deposit after parsing rather than here.

crates/tempo-zone/src/builder.rs
```diff
@@ -134,27 +134,35 @@ where
...
             .deposit_queue
             .lock()
             .expect("deposit queue poisoned")
             .drain();
 
-        if !pending_deposits.is_empty() {
+        let all_deposits: Vec<Deposit> = pending_l1_blocks
+            .iter()
+            .flat_map(|block| block.deposits.clone())
+            .collect();
```
0xKitsune: Dont change the name from `pending_deposits`, this is deposit data with block header, rather than pending blocks.

mattsse (approved):
that should do it for now, we can add things like filling gap later once we actually have the l2 contracts avail

PR #104, opened 2026-02-11
feat(xtask): add create-zone and generate-zone-genesis commands

mattsse:
Add two new xtask subcommands for zone lifecycle management:

- **`create-zone`**: Calls the `ZoneFactory` contract on L1 to register a new zone, computing genesis block hashes and submitting the `CreateZoneParams` transaction.

- **`generate-zone-genesis`**: Generates a zone genesis file by deploying zone system contracts (`TempoState`, `ZoneInbox`, `ZoneOutbox`, `ZoneConfig`) into a local EVM and producing the resulting genesis JSON.

### Changes
- `xtask/src/create_zone.rs` — new command
- `xtask/src/generate_zone_genesis.rs` — new command
- `xtask/src/main.rs` — wire up both subcommands
- `xtask/Cargo.toml` — add `tempo-primitives`, `alloy-consensus`, `alloy-rlp`, `alloy/contract` deps

### DIFF
--- a/xtask/src/create_zone.rs
+++ b/xtask/src/create_zone.rs
@@ -1,16 +1,15 @@
 use alloy::{
-    network::EthereumWallet,
-    primitives::{Address, Bloom, Bytes, B256, B64, U256},
+    network::{EthereumWallet, primitives::{HeaderResponse, ReceiptResponse}},
+    primitives::{Address, B256, keccak256},
     providers::{Provider, ProviderBuilder},
     signers::local::PrivateKeySigner,
     sol,
 };
-use alloy_consensus::Header;
 use alloy_rlp::Encodable;
-use eyre::{WrapErr as _, eyre};
+use eyre::eyre;
 use std::path::PathBuf;
+use tempo_alloy::TempoNetwork;
 use tempo_chainspec::spec::TEMPO_BASE_FEE;
-use tempo_primitives::TempoHeader;
 
 sol! {
     struct ZoneParams {
@@ -99,7 +98,7 @@ impl CreateZone {
         let key_str = self.private_key.strip_prefix("0x").unwrap_or(&self.private_key);
         let signer: PrivateKeySigner = key_str.parse()?;
         let wallet = EthereumWallet::from(signer);
-        let provider = ProviderBuilder::new()
+        let provider = ProviderBuilder::new_with_network::<TempoNetwork>()
             .wallet(wallet)
             .connect_http(self.l1_rpc_url.parse()?);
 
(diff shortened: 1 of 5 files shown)


xtask/src/generate_zone_genesis.rs
```diff
@@ -0,0 +1,324 @@
...
+        .join(format!("{name}.json"));
+    let content = std::fs::read_to_string(&path)
+        .wrap_err_with(|| format!("failed to read artifact at `{}`", path.display()))?;
+    let artifact: FoundryArtifact = serde_json::from_str(&content)
+        .wrap_err_with(|| format!("failed to parse artifact at `{}`", path.display()))?;
+    decode_hex(&artifact.bytecode.object)
+}
+
+fn decode_hex(input: &str) -> eyre::Result<Vec<u8>> {
+    let stripped = input.strip_prefix("0x").unwrap_or(input);
```
SuperFluffy: Pretty sure `const_hex::decode` does that for you.

EDIT: https://docs.rs/const-hex/latest/const_hex/fn.decode.html

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+        let wallet = EthereumWallet::from(signer);
+        let provider = ProviderBuilder::new()
+            .wallet(wallet)
+            .connect_http(self.l1_rpc_url.parse()?);
+
+        let block_number = match self.tempo_block_number {
+            Some(n) => alloy::eips::BlockNumberOrTag::Number(n),
+            None => alloy::eips::BlockNumberOrTag::Latest,
+        };
+
```
0xKitsune: nit: is there a reason we wouldnt want to use latest? I think we might be able to remove this and always use latest.

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+        let header = parse_tempo_header(&raw_block)?;
+        let mut header_rlp = Vec::new();
+        header.encode(&mut header_rlp);
+        let computed_hash = alloy_primitives::keccak256(&header_rlp);
+        let expected_hash: B256 = raw_block
+            .get("hash")
+            .and_then(|v| v.as_str())
+            .ok_or_else(|| eyre!("no hash in block"))?
+            .parse()?;
+
```
0xKitsune: Cant we use an alloy provider and `get_block()` (or similar)? Seems like this would simplify this.

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+            provider
+                .raw_request("eth_getBlockByNumber".into(), (block_number, false))
+                .await
+        })
+        .await?;
+
+        let header = parse_tempo_header(&raw_block)?;
+        let mut header_rlp = Vec::new();
+        header.encode(&mut header_rlp);
+        let computed_hash = alloy_primitives::keccak256(&header_rlp);
```
0xKitsune: Why do we need to compute the hash here if we get it from the block?

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+            .parse()?;
+
+        if computed_hash != expected_hash {
+            return Err(eyre!(
+                "reconstructed header hash {computed_hash} does not match block hash {expected_hash}"
+            ));
+        }
+        println!("Tempo block {} header validated (hash: {computed_hash})", header.inner.number);
+
+        tokio::time::sleep(std::time::Duration::from_millis(500)).await;
```
0xKitsune: nit: why do we need to sleep here?

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        let factory = ZoneFactory::new(self.zone_factory, &provider);
+        println!("Fetching verifier address from ZoneFactory...");
+        let verifier = Address::from(retry(|| async { factory.verifier().call().await }).await?.0);
+        println!("Verifier: {verifier}");
+
+        tokio::time::sleep(std::time::Duration::from_millis(500)).await;
```
0xKitsune: Same here why do we need to sleep?

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+            verifier,
+            zoneParams: ZoneParams {
+                genesisBlockHash: B256::ZERO,
+                genesisTempoBlockHash: tempo_block_hash,
+                genesisTempoBlockNumber: tempo_block_number,
+            },
+        };
+
+        println!("Creating zone on L1 via ZoneFactory at {}...", self.zone_factory);
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
0xKitsune: Same here, is there a reason we are sleeping? Lmk if I am overlooking something.

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+        println!(
+            "  Genesis written to: {}",
+            self.output.join("genesis.json").display()
+        );
+
+        Ok(())
+    }
+}
+
+fn parse_tempo_header(block: &serde_json::Value) -> eyre::Result<TempoHeader> {
```
0xKitsune: Is there not a way we can get this from the provider?

xtask/src/create_zone.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    Ok(TempoHeader {
+        general_gas_limit,
+        shared_gas_limit,
+        timestamp_millis_part,
+        inner,
+    })
+}
+
+fn parse_u64_hex(val: Option<&serde_json::Value>) -> eyre::Result<u64> {
```
0xKitsune: If we use the provider, can we get rid of all of these functions?

0xKitsune (approved):

PR #106, opened 2026-02-11
feat: update sync logic to fill missing blocks

0xKitsune:
This PR adds logic to sync any missing blocks. Additionally, if this is the first time the node is syncing it will fetch the genesis block from the ZonePortal contract on L1.

### DIFF
--- a/crates/tempo-zone/src/l1.rs
+++ b/crates/tempo-zone/src/l1.rs
@@ -164,42 +164,29 @@ impl L1Subscriber {
             let logs = provider.get_logs(&filter.clone().select(block_number)).await?;
 
             // Parse deposit events from the logs
-            let mut deposits = Vec::new();
-            for log in logs {
-                let deposit = self.parse_deposit(log, block_number)?;
-
-                debug!(
-                    l1_block = block_number,
-                    sender = %deposit.sender,
-                    to = %deposit.to,
-                    amount = %deposit.amount,
-                    memo = %deposit.memo,
-                    "Deposit from L1"
-                );
-
-                deposits.push(deposit);
-            }
+            let deposits: Vec<Deposit> = logs
+                .into_iter()
+                .map(|log| self.parse_deposit(log, block_number))
+                .collect::<eyre::Result<_>>()?;
 
             if deposits.is_empty() {
-                debug!(block = block_number, "No deposits in L1 block");
                 continue;
             }
 
-            info!(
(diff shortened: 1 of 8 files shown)


crates/tempo-zone/src/l1.rs
```diff
@@ -68,10 +87,108 @@ impl L1Subscriber {
...
+        provider_rw.commit()?;
+        Ok(())
+    }
+
+    /// Sync to the current L1 tip.
+    ///
+    /// Reads the last synced L1 block from the database and fetches any missed
+    /// blocks up to the current tip. On first boot (no checkpoint), discovers the
+    /// portal's deploy block via `eth_getCode` binary search and backfills from
+    /// there so no deposits are missed.
```
0xKitsune: This is inefficient, we should just have Zone creation blockheight on the corresponding portal contract.

crates/tempo-zone/src/l1.rs
```diff
@@ -149,6 +271,46 @@ impl L1Subscriber {
         Ok(())
     }
 
+    /// Find the block at which the ZonePortal contract was deployed on L1 by
+    /// binary-searching `eth_getCode`. Returns the first block where the contract
+    /// has code.
+    async fn find_portal_deploy_block(
```
0xKitsune: Lets update the contracts to store the l1 blockheight rather than try to find the blockheight.

crates/tempo-zone/src/l1.rs
```diff
@@ -85,13 +173,15 @@ impl L1Subscriber {
...
 
-        // Init filter once — same for every block
         let filter = Filter::new()
             .address(self.config.portal_address)
             .event_signature(DepositMade::SIGNATURE_HASH);
 
+        self.sync_to_l1_tip(&provider, &filter).await?;
+
+        // ── Live subscription: process new blocks as they arrive ──
         let sub = provider.subscribe_blocks().await?;
```
mattsse: this can also drift I think?

so by the time we synced to l1 tip, and then create a new subscription we might have another gap perhaps?

mattsse (approved):
makes sense, lets try and fix if buggy

crates/tempo-zone/src/l1.rs
0xKitsune: Yes this is possible, we could spawn a task to buffer the blockstream in this case?

PR #110, opened 2026-02-12
docs: privacy zone RPC and execution environment specs

dankrad:
## Summary

- Adds **Zone RPC Specification** (`docs/pages/protocol/privacy/rpc.md`) — locks down the Ethereum JSON-RPC for privacy zones with access key authentication (supports secp256k1, P256, WebAuthn, and Keychain signatures), per-method access control, scoped event filtering, and timing side-channel mitigations.
- Adds **Execution Environment Specification** (`docs/pages/protocol/privacy/execution.md`) — specifies TIP-20 precompile modifications (`balanceOf`/`allowance` privacy, fixed 100k gas transfers, system mint/burn permissions for ZoneInbox/ZoneOutbox) and EVM restrictions (CREATE/CREATE2 disabled).
- Adds **PrivateZoneToken.sol** (`docs/specs/src/zone/PrivateZoneToken.sol`) — reference Solidity contract documenting all TIP-20 precompile modifications in one readable file.

## Key design decisions

- **Access keys**: 30-minute EIP-712-style signed messages. All Tempo signature types supported so P256/WebAuthn-only accounts aren't locked out. Keychain access keys (from AccountKeychain precompile) can authenticate on behalf of their root account.
- **Two-layer privacy**: Execution-level changes (balanceOf reverts for non-owners) protect against in-EVM leaks. RPC-level changes (method restrictions, event filtering) protect against raw state inspection. Neither layer alone is sufficient.
- **Fixed 100k gas**: Removes the account-existence side channel where gas differences reveal whether a recipient has previously received tokens.
- **100ms speed bump**: Methods that must fetch data before checking authorization (e.g., `eth_getTransactionByHash`) enforce a minimum 100ms response time to prevent timing attacks that probe whether a transaction hash is valid.
- **Sequencer universal access**: The registered sequencer address bypasses all restrictions, as required for block production.

## Test plan

- [ ] Review RPC method classification (allowed/scoped/restricted/disabled) for completeness
- [ ] Verify PrivateZoneToken.sol compiles against existing IZone.sol imports
- [ ] Review access key signature format for compatibility with existing Tempo wallet implementations
- [ ] Cross-check system mint/burn call sites against ZoneInbox.sol and ZoneOutbox.sol


gakonst (commented):
## Review

Overall this is a well-structured spec — the two-layer defense (execution + RPC), method classification, and timing mitigations are solid. A few things worth addressing:

### Concrete bug: `_requireMintBurnAuth` is too permissive

In `PrivateZoneToken.sol`, the shared `_requireMintBurnAuth()` helper allows both ZoneInbox and ZoneOutbox to mint AND burn. But per the execution spec, only ZoneInbox should mint (deposits) and only ZoneOutbox should burn (withdrawals). Split into `_requireMintAuth()` (checks `ZONE_INBOX`) and `_requireBurnAuth()` (checks `ZONE_OUTBOX`).

### Missing: zone chain ID derivation

The spec references `chainId` in the access key message and `zone_getZoneInfo` but never defines how it's derived. Neither `ZoneFactory` nor `ZonePortal` assign one. Suggest adding:

```
chainId = tempoChainId * 100_000 + zoneId
```

This gives ~3.9B zones within uint32 (important for Ledger and other hardware wallets that use uint32 chain IDs), is deterministic from `zoneId` alone, and keeps a human-readable `4217` prefix for the first 100k zones.

### Missing: threat model section

The spec jumps into mechanisms without stating the adversary model. A short section clarifying: is the adversary an unauthorized RPC user? A passive L1 observer? A malicious node operator? This scopes whether the mitigations are sufficient.

### Ambiguous: `eth_getBlockByNumber`/`eth_getBlockByHash` dual classification

These appear in both the "Allowed" table (returns headers without txs) and the "Restricted" table (with `true` for full transactions). Suggest clarifying that the method is allowed but `include_transactions=true` triggers sequencer-only behavior, rather than listing it in two places.

### Ambiguous: `eth_getFilterLogs` dual classification

Appears in both "Scoped" (filtered to authenticated account) and "Disabled" (marked as "unscoped"). Suggest removing the "Disabled" entry and noting in the scoped section that unscoped log access is not supported.

### Naming: "access key" overloading

The term "access key" is used for both the RPC credential (signed message + fields in `X-Access-Key` header) and AccountKeychain keys. Consider calling the RPC credential something distinct (e.g., "RPC token" or "auth token") to reduce confusion.

### Minor: WebSocket + Keychain revocation

The spec covers WebSocket termination on access key expiry, and Keychain revocation within 1 second for new requests. Worth explicitly stating that active WebSocket connections authenticated via a Keychain key MUST also be terminated within the 1-second revocation window.

### Minor: `gasUsed` leaks transfer count

Block headers expose `gasUsed` (intentionally public per the spec). Combined with fixed 100k gas per transfer and no contract creation, this reveals the exact number of TIP-20 operations per block. Worth a sentence explicitly acknowledging this trade-off.

### Minor: timing jitter

The 100ms fixed floor is a good start, but a fixed delay is distinguishable with repeated sampling from low-latency networks. Consider adding random jitter (e.g., `100ms + uniform(0, 50ms)`) so attackers can't average it out.

PR #126, opened 2026-02-16
feat(zone): EIP-2935 anchor mode, sequencer robustness, and e2e test harness

mattsse:
## Summary

Fixes several critical issues in the Tempo Zone e2e testing infrastructure and batch submission pipeline.

> **Supersedes #122, #123, and #124** — all three PRs are included in this branch and can be closed once this merges.

### Bug Fixes

- **EIP-2935 timing in BatchSubmitter** (`batch.rs`): Changed `>` to `>=` in `resolve_anchor_mode` — EIP-2935 only stores a block's hash when the *next* block processes it, so anchoring to `tempo_block_number == current_l1_block` would revert on-chain.

- **Duplicate block enqueuing in L1Subscriber** (`l1.rs`): The WS subscription starts before backfill to avoid missing blocks, but buffered WS blocks that were already processed during backfill were being enqueued again, breaking chain continuity. Now skips blocks `<= last_enqueued`.

- **EIP-2935 system contract in test genesis** (`test-genesis.json`): Added the EIP-2935 contract with `nonce: 1` so reth recognizes it as deployed, fixing `InvalidTempoBlockNumber` errors in L1 e2e tests.

- **Nonce conflicts in cross-zone tests** (`utils.rs`, `l1_e2e.rs`): Separated user (mnemonic index 1) and sequencer (index 0) L1 accounts. Cross-zone tests now use dedicated sequencer keys (indices 2, 3) per zone. Added `fund_user`, `signer_at`, `create_zone_with_sequencer`, `deploy_two_zones_with_sequencers`.

- **Flaky balance assertions** (`e2e.rs`, `utils.rs`): Changed `wait_for_balance` predicate from `>` to `>=` so tests can wait for the exact expected amount. Relaxed strict equality assertions in L1 e2e tests to account for withdrawal fees.

### From #122 — EIP-2935 direct vs ancestry mode
- Auto-select direct or ancestry mode in `BatchSubmitter::resolve_anchor_mode` based on how old `tempoBlockNumber` is relative to the EIP-2935 history window.

### From #123 — Portal withdrawal queue tail init
- Initialize `portal_withdrawal_queue_tail` from the portal's on-chain state at startup and during resync, so withdrawal data is stored under the correct queue slot.

### From #124 — Batch submission and withdrawal robustness
- Retry logic and error handling improvements for batch submission and withdrawal processing in the zone sequencer.

### Cleanup

- Removed unused imports (`engine.rs`, `executor.rs`)
- Tightened visibility of `build_finalize_withdrawal_batch_tx` and `build_advance_tempo_tx` to `pub(crate)` (`builder.rs`)
- Removed unused `signer` field from `ZoneAccount` test helper


crates/tempo-zone/src/l1_state/provider.rs
```diff
@@ -80,11 +80,29 @@ impl L1StateProvider {
...
+                        config
+                            .l1_rpc_url
+                            .replacen("wss://", "https://", 1)
+                            .replacen("ws://", "http://", 1)
+                            .parse()
+                            .unwrap_or_else(|_| "http://127.0.0.1:1".parse().unwrap()),
+                    )
+                    .erased()
+            }
+        };
```
0xKitsune: I think we should consider returning `Result<Self>` here instead of handling failure like this.

crates/tempo-zone/src/batch.rs
```diff
@@ -45,26 +57,50 @@ pub struct BatchData {
 /// Holds a contract instance pointing at the portal, backed by a shared
 /// [`DynProvider`] with the sequencer's signing wallet.
 pub struct BatchSubmitter {
+    /// ZonePortal contract address on Tempo L1 (used in tracing spans).
     portal_address: Address,
+    /// Shared L1 provider for querying the current block number (EIP-2935
+    /// window check). The same provider backs the `portal` contract instance.
+    l1_provider: DynProvider<TempoNetwork>,
```
0xKitsune, line 66: minor nit: should this be Http provider?

crates/tempo-zone/src/batch.rs
```diff
@@ -14,13 +14,25 @@
...
+/// ancestry mode.
+const EIP2935_HISTORY_WINDOW: u64 = 8192;
+
+/// Safety margin (~3 min at 500ms block time) to avoid race conditions where
+/// the block falls out of the window between our check and on-chain execution.
+const EIP2935_SAFETY_MARGIN: u64 = 360;
+
+/// Finality margin for L1 anchor blocks. Using 64 blocks (~32s at 500ms
+/// block time) to avoid anchoring to blocks that may be reorged.
+const FINALITY_MARGIN: u64 = 64;
```
0xKitsune: Flagging that I dont have high context on if these values are correct. Are these specced? If not we should have @danrobinson or @dankrad weigh in as well.

crates/tempo-zone/src/batch.rs
```diff
@@ -114,6 +166,70 @@ impl BatchSubmitter {
...
+        // in the same or next block where getBlockHash would still return zero.
+        // Wait until L1 advances past our anchor block.
+        if tempo_block_number >= current_l1_block {
+            return Err(eyre::eyre!(
+                "tempo_block_number ({tempo_block_number}) is not yet confirmed on L1 (tip={current_l1_block}), \
+                 will retry after L1 advances"
+            ));
+        }
+
+        let effective_window = EIP2935_HISTORY_WINDOW.saturating_sub(EIP2935_SAFETY_MARGIN);
```
0xKitsune: nit: this should be a const

crates/tempo-zone/src/batch.rs
```diff
@@ -114,6 +166,70 @@ impl BatchSubmitter {
...
+
+        if current_l1_block.saturating_sub(tempo_block_number) < effective_window {
+            return Ok(AnchorMode::Direct);
+        }
+
+        // tempo_block_number is outside the EIP-2935 window — use ancestry mode.
+        // Pick a recent block well within the window as anchor. Use the larger
+        // of the two margins to ensure the anchor is both within EIP-2935 and
+        // past the finality horizon.
+        let anchor_offset = EIP2935_SAFETY_MARGIN.max(FINALITY_MARGIN);
```
0xKitsune: nit: same here, this should be a const

crates/tempo-zone/src/batch.rs
```diff
@@ -114,6 +166,70 @@ impl BatchSubmitter {
...
+            "tempo_block_number outside EIP-2935 window, using ancestry mode"
+        );
+
+        Ok(AnchorMode::Ancestry { anchor_block })
+    }
+
+    /// Read the portal's `genesisTempoBlockNumber` from L1.
+    pub async fn read_genesis_tempo_block_number(&self) -> Result<u64> {
+        let n = self.portal.genesisTempoBlockNumber().call().await?;
+        Ok(n)
```
0xKitsune: nit: 
```suggestion
    self.portal.genesisTempoBlockNumber().call().await
```

crates/tempo-zone/src/batch.rs
```diff
@@ -122,4 +238,77 @@ impl BatchSubmitter {
...
+    /// Check if the withdrawal queue has capacity for another batch.
+    ///
+    /// The portal uses a ring buffer with 100 slots. Returns an error if the
+    /// queue is full (`tail - head >= 100`).
+    pub async fn check_withdrawal_queue_capacity(&self) -> Result<()> {
+        let (head, tail) = tokio::try_join!(
+            self.read_portal_withdrawal_queue_head(),
+            self.read_portal_withdrawal_queue_tail(),
+        )?;
+        const WITHDRAWAL_QUEUE_CAPACITY: u64 = 100;
```
0xKitsune: nit: consider moving this to the top of the file.

crates/tempo-zone/src/withdrawals.rs
```diff
@@ -274,21 +274,44 @@ impl WithdrawalProcessor {
...
-                        index = i,
-                        tx_hash = %pending.tx_hash(),
-                        to = %withdrawal.to,
-                        amount = %withdrawal.amount,
-                        "processWithdrawal tx submitted to L1"
-                    );
+                    let tx_hash = *pending.tx_hash();
+                    match pending
+                        .with_required_confirmations(1)
+                        .with_timeout(Some(std::time::Duration::from_secs(30)))
```
0xKitsune, line 288: 30 seconds seems long here.

crates/tempo-zone/src/zonemonitor.rs
```diff
@@ -218,6 +246,14 @@ impl ZoneMonitor {
             B256::ZERO
         };
 
+        // --- 3. Validate withdrawal data consistency ---
+        if !withdrawal_queue_hash.is_zero() && all_withdrawals.is_empty() {
+            return Err(eyre::eyre!(
+                "withdrawal_queue_hash is non-zero but no withdrawal events found — \
+                 RPC may have returned incomplete data"
+            ));
+        }
```
0xKitsune, line 255: Should this check be here or should we be checking this when processing withdrawals instead?

0xKitsune (approved):
Approving to unblock e2e testing for the other open PRs. In general the changes look straightforward but I think there is some cleanup/iteration we can do. Should be fine for the PoC but noting that we should plan to go back through after the initial PoC.

crates/tempo-zone/src/batch.rs
mattsse: for now these are some random values I picked, 
with the eip2935 window and 500ms blocktime we have ~1hr before the hash disappears from the eip2935 buffer.

we can definitely get lower here, but for now this should do, also need to implement the 

but going to remove the FINALITY_MARGIN because this is redundant since we let the zone follow finalized state

PR #128, opened 2026-02-17
feat(zone): implement public read method handlers and fix method classification

mattsse:
Implements the remaining public read method handlers for the private zone RPC proxy. This proxy acts as an enforced whitelist of JSON-RPC methods with built-in privacy redactions — only explicitly routed methods are reachable, and each applies scoping/redaction before serialization.

**New handlers:** `eth_blockNumber`, `eth_chainId`, `eth_gasPrice`, `eth_maxPriorityFeePerGas`, `eth_feeHistory`, `net_version`, `net_listening`, `web3_clientVersion`, `eth_getBalance`, `eth_getTransactionCount`

**Fixes:**
- `eth_getCode`, `eth_getStorageAt`, `eth_getBlockReceipts` moved to sequencer-only (was incorrectly Public)
- `net_version` returns decimal string instead of hex chain ID
- State override rejection enforced in both handler and API layer (defense-in-depth)
- Deduplicated `BoxFut` type alias, added `raw_zero()` helper

**Follow-up:** `eth_getLogs` / `eth_getFilterLogs` / `eth_getFilterChanges` with TIP-20 event filtering + 100ms timing floor for side-channel protection.


bin/tempo-zone/src/main.rs
```diff
@@ -116,6 +125,20 @@ fn main() {
...
+                listen_addr: ([0, 0, 0, 0], args.private_rpc_port).into(),
+                zone_id: 0, // TODO: hardcoded for now
+                chain_id: handle.node.chain_spec().chain().id(),
+                zone_portal: args.portal_address,
+                sequencer: sequencer_addr.unwrap_or_default(),
+            };
+            let api: Arc<dyn zone::rpc::ZoneRpcApi> =
+                Arc::new(zone::rpc::TempoZoneRpc::new(eth_handlers));
+            let local_addr = zone::rpc::start_private_rpc(private_rpc_config, api).await?;
+            info!(target: "reth::cli", %local_addr, "Private zone RPC server started");
```
0xKitsune, line 140: nit: I think we should be able to move this into `ZoneAddOns`, just noting for now, we can do this in a subsequent PR.

crates/tempo-zone/tests/it/private_rpc.rs
```diff
@@ -0,0 +1,310 @@
...
+        "eth_blockNumber",
+        "eth_gasPrice",
+        "eth_maxPriorityFeePerGas",
+        "eth_feeHistory",
+        "eth_getBalance",
+        "eth_getTransactionCount",
+        "eth_call",
+        "eth_estimateGas",
+        "eth_getCode",
+        "eth_getStorageAt",
```
0xKitsune: Test asserts `eth_getCode`, `eth_getStorageAt`, `eth_getBlockReceipts` are Public, but `types.rs` classifies them as Restricted. This test fails atm.

crates/tempo-zone/src/rpc/types.rs
```diff
@@ -0,0 +1,235 @@
...
+        | "debug_traceBlockByHash"
+        | "txpool_content"
+        | "txpool_status"
+        | "txpool_inspect" => Some(MethodTier::Restricted),
+
+        // Disabled (mining, subscriptions not supported via HTTP proxy)
+        "eth_mining" | "eth_hashrate" | "eth_submitWork" | "eth_submitHashrate"
+        | "eth_subscribe" | "eth_unsubscribe" => Some(MethodTier::Disabled),
+
+        _ => None,
```
0xKitsune, line 225: The spec lists the following as `Restricted` but they're absent and rejected as unknown: 
- eth_createAccessList
- eth_getBlockTransactionCountByNumber
- eth_getBlockTransactionCountByHash
- eth_getTransactionByBlockNumberAndIndex
- eth_getTransactionByBlockHashAndIndex
- eth_getUncleCountByBlockNumber
- eth_getUncleCountByBlockHash
- admin_*

Currently they return -32601 (unknown) instead of -32604 (sequencer-only), either the spec or impl should be updated.

crates/tempo-zone/src/rpc/types.rs
```diff
@@ -0,0 +1,235 @@
...
+        | "eth_getTransactionCount"
+        | "eth_call"
+        | "eth_estimateGas"
+        | "eth_feeHistory"
+        | "eth_maxPriorityFeePerGas"
+        | "eth_getBlockByNumber"
+        | "eth_getBlockByHash"
+        | "net_version"
+        | "net_listening"
+        | "web3_clientVersion" => Some(MethodTier::Public),
```
0xKitsune: `eth_syncing`, `eth_coinbase`, `web3_sha3` are listed as Allowed in the spec but absent from the implementation and rejected as unknown instead of routed.

crates/tempo-zone/src/rpc/types.rs
```diff
@@ -0,0 +1,235 @@
...
+        }
+    }
+
+    /// Sequencer-only method (-32604).
+    pub fn sequencer_only() -> Self {
+        Self {
+            code: -32604,
+            message: "Method restricted to sequencer".to_string(),
+            data: None,
+        }
```
0xKitsune, line 110: Spec says `-32005` for sequencer-only but the implementation uses `-32604`.

crates/tempo-zone/src/rpc/api.rs
```diff
@@ -0,0 +1,374 @@
...
+    ) -> BoxFut<'_> {
+        Box::pin(async move {
+            if !auth.is_sequencer {
+                enforce_from(&mut request, &auth)?;
+            }
+
+            let result =
+                EthCall::estimate_gas_at(&self.eth.api, request, block.unwrap_or_default(), None)
+                    .await
+                    .map_err(internal)?;
```
0xKitsune: The spec mentions "Requests that include a state override set or block override object MUST be rejected with -32602". `eth_call` enforces this but `eth_estimateGas` does not parse or reject state overrides.

crates/tempo-zone/src/rpc/types.rs
```diff
@@ -0,0 +1,235 @@
...
+        | "net_version"
+        | "net_listening"
+        | "web3_clientVersion" => Some(MethodTier::Public),
+
+        // Fetch-then-check: public but redacted based on caller identity
+        "eth_getTransactionByHash"
+        | "eth_getTransactionReceipt"
+        | "eth_getLogs"
+        | "eth_getFilterLogs"
+        | "eth_getFilterChanges" => Some(MethodTier::Public),
```
0xKitsune: `eth_newFilter`, `eth_newBlockFilter`, `eth_uninstallFilter` not routed. The spec lists these as allowed/scoped but these are not explictly handled and will be rejected as unknown.

crates/tempo-zone/src/rpc/auth/token.rs
```diff
@@ -0,0 +1,228 @@
...
+        // Build the signing digest
+        let mut msg = Vec::with_capacity(32 + TOKEN_FIELDS_LEN);
+        msg.extend_from_slice(&TEMPO_ZONE_RPC_MAGIC);
+        msg.push(version);
+        msg.extend_from_slice(&zone_id.to_be_bytes());
+        msg.extend_from_slice(&chain_id.to_be_bytes());
+        msg.extend_from_slice(zone_portal.as_slice());
+        msg.extend_from_slice(&issued_at.to_be_bytes());
+        msg.extend_from_slice(&expires_at.to_be_bytes());
+        let digest = keccak256(&msg);
```
0xKitsune, line 90: Should we use `build_token_fields()` directly to avoid any footguns if this is updated in one place and not the other?

0xKitsune (approved):
Similar to #126, approving to unblock the PoC, but some iteration/cleanup should be done in a follow-up.

PR #156, opened 2026-02-20
fix: detect L1 reorgs in deposit subscriber and purge stale entries

mattsse:
The L1 deposit subscriber tracked `last_enqueued` as a bare block number, so after an L1 reorg the subscription would skip the new canonical block at the same height (`block_number <= last_enqueued`). This left stale old-fork entries in the deposit queue, causing the builder to permanently fail with `L1 parent hash mismatch — chain continuity broken`.

**Fix:** track `last_enqueued` as `NumHash` (number + hash) and detect reorgs via two signals:

1. Block at or below our tip with a different hash — the subscription re-delivered after a reorg
2. Next sequential block whose parent hash doesn't match our tip — shallow reorg the subscription didn't surface as a lower-height block

On detection, stale pending entries are purged from the deposit queue and the existing gap-detection + backfill logic naturally re-fetches canonical blocks.


mattsse:
clippy nightly unrelated

PR #224, opened 2026-03-06
feat: use Tempo 2D nonce keys for sequencer L1 operations

### LINKED ISSUE #196: Sequencer: use Tempo transactions with per-key nonces for L1 operations
The sequencer sends multiple transaction types from the same key on L1 (ZonePortal), creating nonce contention when sent concurrently.

**Frequent:**
- `submitBatch` — batch proof submission (every batch)
- `processWithdrawal` — one per withdrawal (potentially many per batch)

**Infrequent:**
- `enableToken` — enable new TIP-20 for bridging
- `setZoneGasRate` — update deposit fee rate
- `setSequencerEncryptionKey` — rotate encryption key
- `pauseDeposits` / `resumeDeposits` — toggle deposits per token
- `transferSequencer` — sequencer handoff

Using regular Ethereum transactions, these must be serialized to avoid nonce conflicts. Tempo transactions with custom nonce keys would allow each operation type to use its own nonce key, enabling concurrent submission.

**Proposed nonce key partitioning:**
- `submitBatch` — dedicated key (highest throughput)
- `processWithdrawal` — dedicated key (high throughput, multiple per batch)
- Admin operations (enableToken, setZoneGasRate, etc.) — shared key (low frequency)

mattsse:
Assigns dedicated nonce lanes to each sequencer operation type so that `submitBatch`, `processWithdrawal`, and admin transactions can be submitted concurrently without nonce contention.

Uses the upstream `NonceKeyFiller` from tempo-alloy v1.4.1 which automatically reads initial nonces from the L1 `NonceManager` precompile on first use and caches them per `(address, nonce_key)` pair. Callers only need to set `.nonce_key(KEY)` on each contract call.

**Nonce key partitioning:**
- Key 1: `submitBatch` (highest throughput, one per batch cycle)
- Key 2: `processWithdrawal` (high throughput, N per batch)
- Key 3: admin ops (shared, low frequency)

Closes #196

### DIFF
--- a/crates/tempo-zone/src/nonce_keys.rs
+++ b/crates/tempo-zone/src/nonce_keys.rs
@@ -1,0 +1,23 @@
+//! Nonce key constants for zone sequencer L1 operations.
+//!
+//! Tempo's 2D nonce system allows each account to maintain independent nonce
+//! counters ("lanes") keyed by a `U256` nonce key. Each sequencer operation
+//! type uses a dedicated lane so that `submitBatch`, `processWithdrawal`, and
+//! admin transactions can be submitted concurrently without nonce contention.
+//!
+//! Nonce management is handled by [`NonceKeyFiller`](tempo_alloy::fillers::NonceKeyFiller)
+//! in the provider pipeline — callers only need to set `.nonce_key(KEY)` on
+//! each contract call.
+
+use alloy_primitives::{U256, uint};
+
+/// Nonce key for `submitBatch` calls (highest throughput, one per batch cycle).
+pub const SUBMIT_BATCH_NONCE_KEY: U256 = uint!(1_U256);
+
+/// Nonce key for `processWithdrawal` calls (high throughput, N per batch).
+pub const PROCESS_WITHDRAWAL_NONCE_KEY: U256 = uint!(2_U256);
+
+/// Nonce key for admin operations (`enableToken`, `setZoneGasRate`,
+/// `setSequencerEncryptionKey`, `pauseDeposits`, `resumeDeposits`,
+/// `transferSequencer`). Low frequency, shared key.
+pub const ADMIN_OPS_NONCE_KEY: U256 = uint!(3_U256);
--- a/crates/tempo-zone/src/batch.rs
+++ b/crates/tempo-zone/src/batch.rs
@@ -34,8 +34,10 @@ use alloy_provider::{DynProvider, Provider};
 use alloy_rlp::Encodable;
 use eyre::Result;
 use futures::{StreamExt, TryStreamExt};
-use tempo_alloy::TempoNetwork;
+use tempo_alloy::{TempoNetwork, rpc::TempoCallBuilderExt};
 use tracing::{info, instrument, warn};
+
+use crate::nonce_keys::SUBMIT_BATCH_NONCE_KEY;
 
 /// EIP-2935 stores the last 8192 block hashes (~68 min at 500ms block time).
 const EIP2935_HISTORY_WINDOW: u64 = 8192;
@@ -173,6 +175,7 @@ impl BatchSubmitter {
                 Bytes::new(),
                 Bytes::new(),
             )
+            .nonce_key(SUBMIT_BATCH_NONCE_KEY)
             .send()
             .await?
             .with_required_confirmations(1)
--- a/crates/tempo-zone/src/lib.rs
+++ b/crates/tempo-zone/src/lib.rs
@@ -16,6 +16,7 @@ mod executor;
 pub mod l1;
 pub mod l1_state;
 mod node;
+pub mod nonce_keys;
 pub mod payload;
 pub mod precompiles;
 pub mod rpc;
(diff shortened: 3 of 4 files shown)


crates/tempo-zone/src/nonce_keys.rs
```diff
@@ -0,0 +1,242 @@
...
+/// Nonce key for `submitBatch` calls (highest throughput, one per batch cycle).
+pub const SUBMIT_BATCH_NONCE_KEY: U256 = U256::from_limbs([1, 0, 0, 0]);
+
+/// Nonce key for `processWithdrawal` calls (high throughput, N per batch).
+pub const PROCESS_WITHDRAWAL_NONCE_KEY: U256 = U256::from_limbs([2, 0, 0, 0]);
+
+/// Nonce key for admin operations (`enableToken`, `setZoneGasRate`,
+/// `setSequencerEncryptionKey`, `pauseDeposits`, `resumeDeposits`,
+/// `transferSequencer`). Low frequency, shared key.
+pub const ADMIN_OPS_NONCE_KEY: U256 = U256::from_limbs([3, 0, 0, 0]);
```
0xKitsune: We can use `uint!(1)`, `uint!(2)`, `uint!(3)` here.

crates/tempo-zone/src/nonce_keys.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Shared handle to the [`NonceKeyManager`].
+///
+/// When `enabled` is `false` (NonceManager precompile unavailable), all
+/// `nonce_key_*` methods return `None` and nonce allocation is a no-op,
+/// falling back to the provider's default nonce management.
+#[derive(Clone, Debug)]
+pub struct SharedNonceKeyManager {
+    inner: Arc<Mutex<NonceKeyManager>>,
+    enabled: bool,
```
0xKitsune: Do we need this `enabled` bool? Would rather this be explicit The doc comments mention `When enabled is false (NonceManager precompile unavailable), all nonce_key_* methods return None and nonce allocation is a no-op`. Would rather this be explicit and fail if the nonce manager precompile is not available.

crates/tempo-zone/src/nonce_keys.rs
```diff
@@ -0,0 +1,242 @@
...
+        use alloy_sol_types::SolCall;
+        use tempo_alloy::rpc::TempoTransactionRequest;
+
+        // NonceManager predeploy address on Tempo L1
+        const NONCE_MANAGER: alloy_primitives::Address =
+            address!("0xE550000000000000000000000000000000000000");
+
+        alloy_sol_types::sol! {
+            function getNonce(address account, uint256 nonceKey) external view returns (uint64);
+        }
```
0xKitsune: Can we move these to the top of the file?

crates/tempo-zone/src/nonce_keys.rs
```diff
@@ -0,0 +1,242 @@
...
+        sequencer: alloy_primitives::Address,
+    ) {
+        if !self.enabled {
+            return;
+        }
+
+        let (submit_batch, process_withdrawal, admin_ops) = tokio::join!(
+            Self::read_nonce(provider, sequencer, SUBMIT_BATCH_NONCE_KEY),
+            Self::read_nonce(provider, sequencer, PROCESS_WITHDRAWAL_NONCE_KEY),
+            Self::read_nonce(provider, sequencer, ADMIN_OPS_NONCE_KEY),
```
0xKitsune: It looks like we are always reading all three nonce keys, can we move this all into a separate function that returns the nonces?

crates/tempo-zone/src/nonce_keys.rs
```diff
@@ -0,0 +1,242 @@
...
+        );
+
+        let (submit_batch, process_withdrawal, admin_ops, enabled) =
+            match (results.0, results.1, results.2) {
+                (Ok(sb), Ok(pw), Ok(ao)) => (sb, pw, ao, true),
+                _ => {
+                    info!(
+                        "NonceManager precompile not available, \
+                         2D nonce keys disabled — using provider default nonce management"
+                    );
```
0xKitsune: Is there ever a case where the noncemanager precompile will not be available? Seems like we should fail here instead if we are using the `SharedNonceKeyManager`.

crates/tempo-zone/src/zonemonitor.rs
```diff
@@ -583,6 +588,15 @@ impl ZoneMonitor {
                     .await
                     .unwrap_or(self.prev_processed_deposit_hash);
 
+                // Resync 2D nonce counters from the NonceManager precompile.
+                self.batch_submitter
+                    .nonce_manager()
+                    .resync(
+                        self.batch_submitter.l1_provider(),
+                        self.batch_submitter.sequencer_address(),
+                    )
```
0xKitsune: Should this simply have the provider and sequencer address as a part of the nonce manager rather than passing this in?

0xKitsune (commented):
Nice, this seems useful. Left a few comments where I think we can simplify.

PR #238, opened 2026-03-11
docs(zones): specify hard fork activation mechanism

dankrad:
## Summary

- Adds a **Hard fork activation** section to `docs/pages/protocol/privacy/overview.md` specifying how zones upgrade in lockstep with Tempo L1.
- Key design decisions:
  - **Same-block activation**: the zone block that imports the fork L1 block uses new rules for its entire scope (not the next block).
  - **Caller-specified verifier**: `submitBatch` takes a `targetVerifier` parameter instead of the portal deducing it from `tempoBlockNumber`. The portal only checks it's one of two recognized verifiers.
  - **Two-verifier invariant**: at most two verifiers active at any time, with rolling deprecation at each successive fork.
  - **Fork signaling via `protocolVersion`**: `ZoneFactory` maintains a counter incremented at each fork. Outdated zone nodes halt cleanly instead of producing invalid blocks.
- Updates Non-goals to reference the new upgradeability section.

## Test plan

- [ ] Review spec for correctness and completeness
- [ ] Confirm verifier routing pseudocode matches intended `submitBatch` interface
- [ ] Validate two-verifier rotation logic covers first-fork and multi-fork scenarios
- [ ] Check open questions section for any missing concerns


dankrad:
Idea: We can stop user errors (zone operators not upgrading their node software) by adding the verifier contract to the list of watched state and stopping zone if the block is imported with a verifier upgrade that is not supported

Issue #255, opened 2026-03-13
Improve observability for hosted zone operations


mattsse:
## Summary

We should treat observability as in-scope for the April 5 release so the hosted Moderato zone is not just functional, but operable.

This issue tracks a focused observability pass for the zone stack centered on service-level metrics for the critical deposit -> execute -> batch -> withdraw flow.

## What we already have

- Generic node metrics ports are already exposed in local tooling via `--metrics` and `--consensus.metrics-address`.
- `tempo.nu` advertises an observability stack (`nu tempo.nu infra up`) with Prometheus + Grafana.
- The codebase already emits structured logs across the main zone services:
  - `L1Subscriber`
  - `ZoneMonitor`
  - `BatchSubmitter`
  - `WithdrawalProcessor`
  - private RPC server / handlers
- There is already one meaningful custom metrics surface for TIP-403:
  - authorization checks / cache hits / cache misses
  - RPC resolution latency / errors
  - prefetch request counts / in-flight work
  - cached policy / token-policy gauges
  - listener-applied event counts

## Gaps / problems

The current observability is not yet sufficient for operating a hosted zone release:

- We have very little service-level instrumentation for the core zone flows:
  - deposit ingestion / backfill / live L1 sync
  - zone block production lag
  - batch submission success / retry / latency / anchor mode
  - withdrawal queue progress and stuck-slot detection
  - private RPC auth / request / latency / failure visibility
- Most sequencing visibility today comes from logs rather than queryable counters/gauges/histograms.

## Proposed scope

Add a minimal-but-real observability layer for hosted zones with the following coverage.

### 1. Core service metrics

Add metrics for:

- L1 subscriber
  - current connected state
  - backfill start/end/duration
  - latest L1 block seen
  - current lag from L1 tip
  - blocks enqueued to deposit queue
  - deposit count / encrypted deposit count / token-enabled event count
  - reorgs detected
  - receipt fetch failures / reconnects
- Deposit queue / block production
  - queue depth
  - oldest pending L1 block age
  - latest L1 block confirmed into zone
  - zone block production rate
  - block-building failures / retries
- Zone monitor / batch submission
  - latest zone block observed
  - latest zone block submitted to L1
  - zone->L1 submission lag (blocks)
  - batch size in blocks
  - batch submit latency
  - batch submit success / failure / retry counters
  - direct vs ancestry mode counts
  - resync-from-portal count
  - withdrawals collected per batch
- Withdrawal processor
[...]

PR #260, opened 2026-03-13
feat(zone): add L1 subscriber observability metrics

mattsse:
## Summary
- add dedicated `zone_l1_subscriber` metrics for connection state, backfill lifecycle, latest seen block, lag, reconnects, reorgs, and receipt fetch failures
- record regular deposit, encrypted deposit, token-enabled, and accepted-into-queue counts from both live ingestion and backfill
- keep the existing TIP-403 metrics in place while separating subscriber-level observability into its own internal metrics module

## Testing
- `cargo check -p zone`
- `cargo test -p zone --lib l1::tests`

Part of #255


crates/tempo-zone/src/l1.rs
```diff
@@ -378,14 +436,27 @@ impl L1Subscriber {
...
+                    enqueued += 1;
+                    self.subscriber_metrics.blocks_enqueued_total.increment(1);
+                }
+                EnqueueOutcome::Duplicate => {}
+                EnqueueOutcome::NeedBackfill { from, to } => {
+                    panic!(
+                        "backfill enqueue expected Accepted or Duplicate, got NeedBackfill({from}..={to})"
+                    );
+                }
+            }
```
mattsse: we can keep this as it was

crates/tempo-zone/src/l1.rs
```diff
@@ -337,18 +383,30 @@ impl L1Subscriber {
...
+                            {
+                                Ok(Some(receipts)) => Ok(receipts),
+                                Ok(None) => {
+                                    subscriber_metrics.receipt_fetch_failures_total.increment(1);
+                                    Err(eyre::eyre!("no receipts for block {block_number}"))
+                                }
+                                Err(err) => {
+                                    subscriber_metrics.receipt_fetch_failures_total.increment(1);
+                                    Err(err.into())
+                                }
```
mattsse: I'd like to use inspect and inspect_err for this instead

crates/tempo-zone/src/l1.rs
```diff
@@ -208,16 +225,28 @@ impl L1Subscriber {
...
+                        Ok(Some(receipts)) => receipts,
+                        Ok(None) => {
+                            subscriber_metrics.receipt_fetch_failures_total.increment(1);
+                            return Err(eyre::eyre!("no receipts for block {block_number}"));
+                        }
+                        Err(err) => {
+                            subscriber_metrics.receipt_fetch_failures_total.increment(1);
+                            return Err(err.into());
+                        }
+                    };
```
mattsse: same here inspect+inspect_err

crates/tempo-zone/src/l1.rs
```diff
@@ -90,21 +92,35 @@ impl L1Subscriber {
...
-                    if let Err(e) = subscriber.clone().run().await {
-                        let retry_interval = subscriber.config.retry_connection_interval;
-                        error!(
-                            error = %e,
-                            retry_secs = retry_interval.as_secs_f32(),
-                            "L1 subscriber failed, reconnecting after retry interval"
-                        );
-                        tokio::time::sleep(retry_interval).await;
+                    let retry_interval = subscriber.config.retry_connection_interval;
+                    match subscriber.clone().run().await {
```
mattsse: we can keep the if let err because we assume this is running forever

crates/tempo-zone/src/metrics.rs
```diff
@@ -0,0 +1,53 @@
...
+    Metrics,
+    metrics::{Counter, Gauge, Histogram},
+};
+
+/// Metrics emitted by the L1 subscriber / deposit ingestion pipeline.
+#[derive(Metrics, Clone)]
+#[metrics(scope = "zone_l1_subscriber")]
+pub(crate) struct L1SubscriberMetrics {
+    /// Whether the subscriber is currently connected to L1 (1) or disconnected (0).
+    pub connected: Gauge,
```
mattsse: this metric isnt very useful, lets remove

crates/tempo-zone/src/metrics.rs
mattsse, line 1: lets also add a histogram that tracks how long it takes to await stream.try_next() (preparing the block)

PR #272, opened 2026-03-13
Add withdrawal and private RPC metrics

mattsse:
## Summary
- add withdrawal processor metrics for queue state, store batch count, stuck-head age, per-withdrawal outcomes, and slot processing latency
- add private RPC metrics for auth failures, per-method HTTP/WS dispatch outcomes, WS session lifecycle, and provider token refreshes
- add targeted metrics tests for withdrawal helpers, provider refresh behavior, and private RPC HTTP/WS paths

## Testing
- cargo test -p zone --lib withdrawals::tests
- cargo test -p zone-rpc provider::tests --lib
- cargo test -p zone-rpc --test it metric

Part of #255.


0xKitsune (commented):
Overall lgtm, noticed a few small issues:

- N+1 query pattern in `zone_getDepositStatus`
    - `rpc.rs` L751–811 — each deposit fires a separate `get_logs` RPC call via `terminal_event_for_deposit`. If a block has many deposits (e.g. CEX batch), this serializes into high latency. Consider batching into a single `get_logs` with multiple `topic1` hashes.

- Possible double-recording of `slot_processing_duration_seconds`
    - `withdrawals.rs` L467–500 — `slot_metrics.finish()` is called in error branches *and* again at the end of the loop. If the error branches don't `return Err(...)` (i.e. they just break), the histogram gets two entries for the same batch. Worth verifying the control flow guarantees a single recording.

- Missing test: pending encrypted deposit viewed by recipient
    - `private_rpc_e2e.rs` — tests cover processed regular/encrypted deposits, but there's no test asserting that a **recipient** sees an empty list when querying a **pending** (not yet processed) encrypted deposit. This is the key privacy invariant for encrypted deposits.

- `classify_public_methods` classification
    - `private_rpc.rs` L331 — new zone methods added to the `Public` method list, but they require `AuthContext`. Worth confirming `Public` here means "any authenticated user" not "anonymous/unauthenticated".

mattsse:
Superseded by #282, which rebuilds the #255 metrics work cleanly on top of main without the unrelated blended changes from this PR.

PR #274, opened 2026-03-13
docs(zones): specify authenticated withdrawals

dankrad:
## Summary

- Replaces plaintext `sender` in the `Withdrawal` struct with `senderTag` — a commitment `keccak256(sender || txHash)` that hides the sender on L1 while enabling selective disclosure via the zone transaction hash as a blinding factor.
- Adds `encryptedSender` field and `revealTo` parameter to `requestWithdrawal`, enabling the sequencer to ECDH-encrypt the sender reveal to a target public key (e.g., another zone's sequencer). This supports zone-to-zone transfers where the target sequencer can automatically verify the sender.
- Updates `IWithdrawalReceiver.onWithdrawalReceived` and `IZoneMessenger.relayMessage` to pass `senderTag` instead of `sender`.
- Documents the trust model (sequencer-attested, with upgrade path to ZK-proven), encrypted sender format, and the zone-to-zone flow.

## Test plan

- [ ] Review `senderTag` commitment scheme for correctness (blinding, uniqueness per withdrawal)
- [ ] Verify `encryptedSender` ECDH format is consistent with existing encrypted deposit scheme
- [ ] Check zone-to-zone flow for completeness (sender -> Zone A sequencer -> L1 -> Zone B sequencer)
- [ ] Confirm callback interface changes are reflected in all relevant code blocks
- [ ] Validate that zone-side events retain plaintext sender (private, no leak)

### DIFF
--- a/docs/pages/protocol/privacy/overview.md
+++ b/docs/pages/protocol/privacy/overview.md
@@ -361,7 +361,7 @@ struct Deposit {
 
 struct Withdrawal {
     address token;              // TIP-20 token being withdrawn
-    address sender;             // who initiated the withdrawal on the zone
+    bytes32 senderTag;          // keccak256(abi.encodePacked(sender, txHash)) — see Authenticated withdrawals
     address to;                 // Tempo recipient
     uint128 amount;             // amount to send to recipient (excludes fee)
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense to me

PR #309, opened 2026-03-25
feat(l1-subscriber): simplify subscriber flow, deposit processing

0xKitsune:
This PR simplifies the L1 subscriber flow, reduces latency in deposit processing from L1, and consolidates reorg detection and cache locking.

### Changes

* **Removed 1-block confirmation buffer**
  Blocks are now applied immediately as they arrive. Given L1’s fast finality, the buffer did not provide meaningful reorg protection but introduced ~500ms of latency before deposits were reflected on the zone.

* **Improved reorg detection flow**
  `sync_to_l1_tip` now returns the tip block hash, which seeds a `parent_hash` tracker in the new `handle_l1_block_stream` method. This enables simpler and more direct reorg detection.

* **Unified cache updates**
  Merged `update_l1_state_anchor` and `apply_portal_state_events` into `update_l1_state_cache`, reducing cache updates to a single write lock per block.

* **Removed duplicate reorg handling**
  Eliminated redundant reorg detection from `update_l1_state_anchor`. Reorgs are now handled once at the stream level.

* **Removed non-actionable metric**
  Dropped the `listener_events_applied` metric. Existing cache gauges provide sufficient visibility.

### Follow Ups

Reorg handling currently only clears the L1 state cache. A follow-up PR is needed to extend this to fully unwind any already-applied state.

### DIFF
--- a/.github/workflows/pr-audit.yml
+++ b/.github/workflows/pr-audit.yml
@@ -1,0 +1,15 @@
+name: Pull request audit
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  audit:
+    uses: tempoxyz/gh-actions/.github/workflows/pr-audit.yml@main
+    with:
+      required-label: cyclops
+    secrets:
+      EVENTS_KEY: ${{ secrets.EVENTS_KEY }}
+      EVENTS_CERT: ${{ secrets.EVENTS_CERT }}
+      EVENTS_ARGS: ${{ secrets.EVENTS_ARGS }}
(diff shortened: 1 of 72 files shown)


mattsse (commented):
lots of changes in here, unsure I fully understand this or if ci failure is related

0xKitsune:
I suspect the failures are due to a code change, will take a closer look.

0xKitsune:
Closing for now, will reopen with a more scoped set of changes.

PR #321, opened 2026-03-26
fix: pin policy data and existence reads to engine block height

0xKitsune:
This PR adds a `created_at` block number to cached policy records so that `policyData`, `compoundPolicyData`, and `policyExists` queries only return cache hits for policies created at or before the engine's committed L1 height, preventing nodes with different L1 subscriber progress from producing divergent EVM results.


0xKitsune:
Closing for now — single-sequencer architecture means no consensus divergence is possible today. Will revisit if/when multi-node validation is added.

PR #348, opened 2026-03-31
feat(fees): add isFeeToken flag to TIP-20 for fee token eligibility

decofe:
Adds a per-token `isFeeToken` boolean to TIP-20 that controls whether a token can be used for fee payment.

## Motivation

Zones don't have a FeeAMM, so the L1 approach of calling `_requireUSDTIP20()` (which lives in FeeAMM) doesn't apply. Instead, a per-token flag gives fine-grained control: USD tokens get `isFeeToken = true` at creation, and in the future individual USD tokens can be disabled or non-USD tokens enabled without code changes.

## Changes

- **`TIP20.sol`**: add `bool public isFeeToken`, set to `true` in constructor when `currency == "USD"`
- **`ITIP20.sol`**: add `isFeeToken()` to the interface
- **`IFeeManager.sol`**: add `NotFeeToken` error
- **`FeeManager.sol`**: replace `_requireUSDTIP20()` calls with `_requireFeeToken()` that checks the flag; add validation to `collectFeePreTx` and `collectFeePostTx`
- **`FeeManager.t.sol`**: add tests for non-fee-token rejection in `setValidatorToken`, `collectFeePreTx`, `collectFeePostTx`, and flag correctness

## Testing

```
forge test --match-contract FeeManagerTest -vvv
# 23 passed, 0 failed
```

Prompted by: dankrad

### DIFF
--- a/docs/specs/src/FeeManager.sol
+++ b/docs/specs/src/FeeManager.sol
@@ -19,21 +19,26 @@ contract FeeManager is IFeeManager, FeeAMM {
         _;
     }
 
+    function _requireFeeToken(address token) internal view {
+        if (!ITIP20(token).isFeeToken()) revert NotFeeToken();
+    }
+
     function setValidatorToken(address token) external onlyDirectCall {
         require(msg.sender != block.coinbase, "CANNOT_CHANGE_WITHIN_BLOCK");
-        _requireUSDTIP20(token);
+        _requireFeeToken(token);
         validatorTokens[msg.sender] = token;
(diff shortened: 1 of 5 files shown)


mattsse (approved):
unsure if we even need this if the tokens need to be enabled manually on l1

0xKitsune:
> unsure if we even need this if the tokens need to be enabled manually on l1

Yes closing this, we can explore toggling fee tokens from the ZonePortal contract on L1

PR #356, opened 2026-04-01
feat(tempo-zone): update executor to override validator fee token

0xKitsune:
Currently, zones use the Tempo EVM which routes fee payments through `FeeAMM` when the user's fee token differs from the sequencer's `validatorToken`. This reverts if no pool exists, which means users can only pay fees in the sequencer's preferred token.

Zones have a different design where the sequencer enables TIP-20 tokens and users should be able to pay gas in any enabled token without AMM pools.

This PR adds an override in `ZoneBlockExecutor` that writes `validatorTokens[beneficiary] = resolved_fee_token` before each transaction, so the handler sees matching tokens and skips FeeAMM entirely. Uses the same `get_fee_token` resolution as the handler, resulting in zero upstream changes.

Tests verify multi-token fee collection across 3 TIP-20s with no FeeAMM liquidity and confirm pool reserves stay zero.

### DIFF
--- a/crates/tempo-zone/src/executor.rs
+++ b/crates/tempo-zone/src/executor.rs
@@ -65,11 +65,9 @@ where
         let slot = TipFeeManager::new().validator_tokens[beneficiary].slot();
 
         let _ = ctx.journal_mut().load_account(TIP_FEE_MANAGER_ADDRESS);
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think this should do it

PR #364, opened 2026-04-07
fix: remove redundant to/memo from DecryptionData

### LINKED ISSUE #357: DecryptionData contains too much info
We don't need memo and to in decryption data as it is derived on chain

dankrad:
## Summary

- Remove `to` and `memo` fields from the `DecryptionData` struct. These were redundant because `ZoneInbox` already decrypts the ciphertext via AES-256-GCM on-chain and can derive `(to, memo)` directly from the plaintext.
- Update `ZoneInbox.sol` to use the decrypted values directly instead of comparing them against sequencer-supplied duplicates.
- Saves 52 bytes of calldata per encrypted deposit (20-byte address + 32-byte memo).

Closes #357

## Changes

| Layer | Files | What changed |
|-------|-------|-------------|
| Solidity interface | `IZone.sol` | Removed `to` and `memo` from `DecryptionData` |
| Solidity logic | `ZoneInbox.sol` | Uses `decryptedTo`/`decryptedMemo` from `decodePlaintext()` directly |
| Solidity tests | `ZoneInbox.t.sol`, `ZoneBridge.t.sol` | Removed fields from all struct constructors |
| Rust ABI | `abi.rs`, `advance_tempo.rs` | Removed fields from sol! struct |
| Rust builder | `l1.rs`, `builder.rs` | Removed fields from construction sites |
| Docs | `overview.md`, `crypto-review.md`, `prover-design.md` | Updated struct definitions and code snippets |

## Note

The off-chain TIP-403 policy check in `l1.rs` that previously redirected unauthorized recipients by setting `to: d.sender` in `DecryptionData` no longer has an on-chain effect — the contract now always uses the decrypted recipient. The policy check is retained for logging but the redirect variable is prefixed with `_` to suppress the unused warning. This behavioral change should be evaluated separately if sequencer-side recipient override is still desired.

## Test plan

- [x] `forge build` — Solidity compiles
- [x] `forge test --match-contract "ZoneInbox|ZoneBridge"` — 41/41 tests pass
- [x] `cargo build --workspace` — Rust compiles
- [x] `cargo test -p zone` — advance_tempo and builder tests pass
- [ ] CI


mattsse (approved):
Approved after updating the Rust bindings/builders and spec text to match the trimmed `DecryptionData` layout. I also added a Rust ABI regression test and reran the focused Rust checks:

- `cargo test -p zone-primitives`
- `cargo test -p zone test_build_advance_tempo_tx_with_encrypted_deposit`

Both pass locally.

PR #365, opened 2026-04-07
fix: chunk zone log queries

mattsse:
## Summary
- bound large zone log scans during catch-up to avoid oversized RPC responses from the zone node
- use Alloy `chunked()` queries for `fetch_slot_withdrawals()` on `WithdrawalRequested` and `BatchFinalized`
- keep `has_pending_withdrawals()` on a small manual chunk loop so it can short-circuit on the first non-zero withdrawal batch

## Why
When the zone monitor falls far behind, it can end up querying a very large `[from, to]` block range. That can make the zone node's `eth_getLogs` response exceed the RPC server size limit and leave the monitor stuck retrying the same backlog.

## Impact
This keeps the catch-up path from depending on a single huge log response while preserving the existing batching behavior.

## Validation
- `cargo fmt --all -- crates/tempo-zone/src/batch.rs crates/tempo-zone/src/zonemonitor.rs`
- `cargo test --manifest-path crates/tempo-zone/Cargo.toml log_query_ranges_chunk_large_ranges -- --nocapture`
- `cargo nextest run -p zone` in progress
- `cargo clippy -p zone --all-targets -- -D warnings` in progress

### DIFF
--- a/crates/tempo-zone/src/batch.rs
+++ b/crates/tempo-zone/src/batch.rs
@@ -57,6 +57,12 @@ const EIP2935_EFFECTIVE_WINDOW: u64 = EIP2935_HISTORY_WINDOW - EIP2935_SAFETY_MA
 /// Maximum number of pending withdrawal queue slots in the portal ring buffer.
 const WITHDRAWAL_QUEUE_CAPACITY: u64 = 100;
 
+/// Maximum zone-block span for a single `eth_getLogs` request during catch-up.
+///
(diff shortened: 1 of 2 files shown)


mattsse:
Validation finished.

- `cargo nextest run --manifest-path crates/tempo-zone/Cargo.toml`: 232 passed, 6 skipped
- `cargo clippy --manifest-path crates/tempo-zone/Cargo.toml --all-targets --all-features -- -D warnings`: passed

PR #370, opened 2026-04-09
fix(zone): harden withdrawal repair resync

mattsse:
## Summary
- add a dedicated monitor repair path for missing withdrawal head slots and document why it performs a full portal resync
- throttle repeated repair requests to the withdrawal processor fallback interval and take one consistent store snapshot for diagnostics
- fail closed when post-resync withdrawal restore fails by clearing the stale local store and incrementing a restore-failure metric
- extend tests to cover repair request throttling, the repair path, and the restore-failure case

## Why
Withdrawals could stop processing when the portal queue still had pending slots but the in-memory withdrawal store no longer contained the current head slot. A restart fixed the issue because startup rebuilt pending withdrawals from chain state. This change teaches the live monitor to do the same repair path without needing a restart, while avoiding back-to-back resync churn if reconstruction still cannot see the head slot yet.

## Impact
The sequencer can now detect this desync at runtime, resync from portal-confirmed state, and rebuild pending withdrawals in place. If the rebuild fails after the portal anchor has already advanced, the node no longer continues with stale withdrawal data silently; it clears the local store and emits a metric so the failure is visible.

## Validation
- `cargo fmt --package zone`
- `cargo test -p zone --lib`

### DIFF
--- a/crates/tempo-zone/src/lib.rs
+++ b/crates/tempo-zone/src/lib.rs
@@ -122,6 +122,7 @@ pub async fn spawn_zone_sequencer(
 
     let withdrawal_store: SharedWithdrawalStore = Default::default();
     let withdrawal_notify = Arc::new(Notify::new());
+    let withdrawal_repair_notify = Arc::new(Notify::new());
 
     let withdrawal_config = WithdrawalProcessorConfig {
         portal_address: config.portal_address,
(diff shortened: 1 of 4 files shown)


0xKitsune (approved):
Approving to unblock, as mentioned another thread, we should see if there are ways to simplify the withdrawal processor.

PR #377, opened 2026-04-10
docs(specs): restructure zones spec

0xKitsune:
This PR reconstructs the zone specification to provide a clearer, more consistent structure and narrative flow.

The updated spec presents concepts in a structured sequence, beginning with a high-level system overview and sequence diagram, followed by zone deployment, sequencer operations, deposits (including encrypted deposits and on-chain decryption verification), execution, and state access. It then covers policy definitions (TIP-403), private RPC, the proving system, batch submission, and withdrawal mechanisms (including authenticated withdrawals and zone-to-zone transfers), and concludes with precompiles, contract interfaces, network upgrades, and security considerations.

Overall, this improvement is aimed at making the specification easier to follow and reason about.


dankrad:
Quick notes:

- I think we should not use "escrow" as it suggests discretion on whether funds are paid out. I suggest replacing with "tokens locked in portal" or similar
- I would add more introduction at the top on actors/trust assumptions/goals etc. Can take from my PR overview.md
- wouldn't do "open questions" in the spec
- "This enforces the non-custodial withdrawal guarantee: the sequencer can halt deposits but can never prevent users from withdrawing an enabled token." this is not correct as they can freeze tokens

Generally feels like a good structure and seems easy to read

dankrad:
Does it make sense to say "system contract" instead of predeploy?

dankrad:
Could make sense to add this table as well: https://github.com/tempoxyz/zones/pull/379/changes/9082f7de91c48b72b4aa593c7749c4c7689ab5f2

docs/specs/zone_spec.md
```diff
@@ -0,0 +1,1198 @@
...
+
+Zones rely on the following trust assumptions: the verifier must be sound for state transition integrity, the sequencer is trusted for liveness and data availability, and there is no forced inclusion or permissionless exit mechanism. Privacy protects against public observers on Tempo but not against the sequencer, who has full visibility into zone activity.
+
+When a user wants to exit, they request a withdrawal on the zone. Their tokens are burned, and the withdrawal is added to a pending list. At the end of a batch, the sequencer finalizes all pending withdrawals into a hash chain and generates a proof covering the full batch of zone blocks. The sequencer submits this batch and proof to the portal on Tempo, which verifies the proof and queues the withdrawals. The sequencer then processes each withdrawal, releasing tokens from the portal to the recipient.
```
dankrad:
```suggestion
When a user wants to exit, they request a withdrawal on the zone. Their tokens are burned on the zone side, and the withdrawal is added to a pending list. At the end of a batch, the sequencer finalizes all pending withdrawals into a hash chain and generates a proof covering the full batch of zone blocks. The sequencer submits this batch and proof to the portal on Tempo, which verifies the proof and queues the withdrawals. The sequencer then processes each withdrawal, releasing tokens from the portal to the recipient.
```

docs/specs/zone_spec.md
```diff
@@ -0,0 +1,1198 @@
...
+| `zoneParams` | Genesis configuration: genesis block hash, genesis Tempo block hash, and genesis Tempo block number. |
+
+The factory assigns a unique `zoneId`, deploys a [`ZonePortal`](#izoneportal) and a [`ZoneMessenger`](#izonemessenger), and enables the initial token. The [`ZoneCreated`](#izonefactory) event emits all deployment parameters.
+
+### Chain ID
+
+Each zone has a unique chain ID derived from its zone ID:
+
+```
+chain_id = 4217000000 + zone_id
```
dankrad:
```suggestion
chain_id = 421700000 + zone_id
```

dankrad:
image
Should the arrow for finalizeWithdrawalBatch() come from the sequencer?

dankrad:
1. Medium: docs/specs/zone_spec.md’s Private RPC section is no longer spec-complete relative to docs/specs/privacy/rpc.md. The old doc explicitly defined the auth-token preimage and validation rules, HTTP-layer behavior (401/403/500 cases), exhaustive allowed/scoped/restricted/disabled method tables, and the full zone_getDepositStatus request/response and visibility contract. The new unified section summarizes that material with prose like “and others” and “mining-related methods,” which leaves several methods and behaviors unspecified, including examples like eth_getBlockByHash, eth_newFilter, eth_newBlockFilter, eth_uninstallFilter, eth_createAccessList, and the uncle/block-count helpers.
2. Medium: docs/specs/zone_spec.md’s Network Upgrades and Hard Fork Activation section drops concrete upgrade mechanics that existed in docs/specs/privacy/upgrades.md and the old Hard fork activation section in docs/specs/privacy/overview.md. Missing details include the forkActivationBlock cutoff semantics, N-2 verifier deprecation/catch-up risk, the exact setForkVerifier plus protocolVersion system-transaction sequence, predeploy injection ordering, and the explicit stale-node / stale-prover / behind-L1 failure modes. 
3. Medium-low: A couple of normative wire/implementation details disappeared instead of being relocated. docs/specs/privacy/overview.md used to define the encryptedSender wire format as ephemeralPubKey || ciphertext || mac, but docs/specs/zone_spec.md now only says it is ECDH-encrypted bytes. docs/specs/privacy/execution.md also had the explicit MUST NOT about sharing inbox/outbox authorization for mint vs burn, and that constraint is no longer stated in the unified execution section. 
4. Low: docs/specs/zone_spec.md has a stale internal link. ZoneConfig says “See Tempo L1 State Reads” and links to #tempo-l1-state-reads, but the section heading is Tempo State Reads, so that anchor no longer resolves.

0xKitsune:
> Does it make sense to say "system contract" instead of predeploy?

I think predeploy is accurate in the current iteration of the protocol.


> Should the arrow for finalizeWithdrawalBatch() come from the sequencer?

This sequence diagram has been updated, lmk what you think.

dankrad (approved):

PR #416, opened 2026-04-20
spec: add EIP-712 signing scheme for zone transactions

dankrad:
## Summary

Adds an alternate signing scheme for zone transactions, based on EIP-712 typed messages, so that wallets without native Tempo or zone envelope support (hardware wallets, WalletConnect-only dapps, legacy SDKs) can author zone transactions by signing structured data with `eth_signTypedData_v4`.

- Introduces signature type `0x05` layered onto the existing Tempo transaction envelope. The wire-level envelope and fields are unchanged; only the digest the signature commits to changes (EIP-712 digest instead of the native tx hash).
- Domain separator binds `chainId` (which already encodes `zoneId`) and the zone's `ZonePortal` address, giving each zone a unique domain and a stable address for wallets to display.
- Two typed structs:
  - `ZoneTransaction` — used when a root account signs directly. Mapping to envelope fields is bijective.
  - `AccessKeyZoneTransaction` — used when a Keychain access key signs on behalf of a root. Includes `userAddress` as the first field so the wallet prompt shows which root account the access key is acting for.
- Access-key composition reuses the existing Keychain V2 wrapper (`0x04 || innerSig || user_address`) around a `0x05` inner signature. Secp256k1 access keys work under EIP-712 without re-registration (the curve is registered, not the digest scheme).
- Keychain V1 (`0x03`) is rejected as an EIP-712 wrapper because it does not bind `user_address` into the signing hash — an EIP-712 wallet cannot see `user_address`, so V1 wrapping would allow a malicious client to rebind the signature to any root.
- Fee-payer delegation and EIP-7702 authorization lists are intentionally not supported under this type; applications needing either use the native Tempo signing scheme.
- `Private RPC > Signature Types` table is extended with the new type, and the TOC is updated.

## Test plan

- [ ] Spec-only change; no runtime behavior affected until the format is implemented.
- [ ] Confirm the rendered TOC entry and internal anchors (`#eip-712-typed-transactions`, `#access-key-signing`, `#signature-types`) resolve correctly on GitHub.
- [ ] Reviewers sanity-check the `ZoneTransaction` and `AccessKeyZoneTransaction` field sets against the final Tempo transaction envelope used on zones.


0xKitsune:
Closing as this is will be implemented into `tempo` via [TIP-1051](https://github.com/tempoxyz/tempo/pull/3682).

PR #427, opened 2026-04-23
feat(node): separate private RPC from sequencer addons

0xKitsune:
This PR extracts private RPC configuration into its own `ZonePrivateRpcConfig` so it can be launched independently of sequencer background tasks. Nodes running in sequencer mode now pass a `--sequencer` flag.

This enables infra setups where additional nodes serve private RPC traffic for state queries and load balancing without sequencing the chain.

This PR also refactors the monolithic `launch_add_ons` into focused helper methods for readability.

```rust
impl<N> NodeAddOns<N> for ZoneAddOns<N>
{
    async fn launch_add_ons(mut self, ctx: AddOnsContext<'_, N>) -> eyre::Result<Self::Handle> {
        // --snip--

        self.resolve_and_seed_tokens(&l1_provider).await?;
        self.spawn_l1_subscriber(&ctx);
        self.spawn_policy_tasks(&l1_provider, &ctx);
        self.spawn_zone_engine(l1_provider, &ctx)?;

       // --snip--
}
```

### DIFF
--- a/crates/tempo-zone/src/cli.rs
+++ b/crates/tempo-zone/src/cli.rs
@@ -94,7 +94,9 @@ impl ZoneCli {
                     zone_id: args.zone_id,
                     zone_poll_interval: Duration::from_secs(args.zone_poll_interval_secs),
                     batch_interval: Duration::from_secs(args.zone_batch_interval_secs),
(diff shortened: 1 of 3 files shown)


mattsse (approved):
ah nice, we def want this

PR #437, opened 2026-04-28
docs: fix demo-swap-and-deposit args

umeeSthein:
## Summary

- Update the optional `just demo-swap-and-deposit` example to use positional arguments.
- Match the `Justfile` recipe so `amount` and `tick` are passed as numeric values.

## Test plan

- `just --dry-run demo-swap-and-deposit my-zone 100000000 0`


mattsse (approved):
Reviewed locally: the documented positional arguments match the Justfile recipe and dry-run output.

PR #440, opened 2026-05-05
fix(tip403): ignore stale cache fallback writes

mattsse:
## Summary

TIP-403 policy cache writes now treat the advanced baseline as finalized state. Delayed RPC fallback results at or below a versioned entry's baseline height are discarded instead of mutating the baseline, so a prefetch or synchronous fallback that started before a later L1 policy event cannot overwrite the event-derived state after the engine advances.

This applies to both token-to-policy `HeightVersioned` entries and simple policy `MembershipSet` entries. Membership stale writes are rejected before touching `observed`, which keeps future cache misses able to fall back to L1 instead of permanently trusting stale negative membership observations.

The regression coverage models stale blacklist membership and stale token-policy fallback writes returning after `advance()`, plus the lower-level versioned primitive behavior.

### DIFF
--- a/crates/tempo-zone/src/l1_state/tip403/policy_set.rs
+++ b/crates/tempo-zone/src/l1_state/tip403/policy_set.rs
@@ -54,23 +54,23 @@ impl PolicySet {
     }
 
     /// Record a set update at the given block height.
+    ///
+    /// Updates at or below the baseline height are ignored. The baseline represents finalized
+    /// engine-consumed state and is only updated by [`advance`](Self::advance), which prevents
+    /// delayed RPC fallback results from overwriting newer event-derived membership.
     pub fn record_status(&mut self, user: Address, block_number: u64, in_set: bool) {
-        self.observed.insert(user);
         if block_number <= self.baseline_height {
-            if in_set {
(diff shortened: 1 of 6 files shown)


crates/tempo-zone/src/l1_state/tip403/policy_set.rs
```diff
@@ -54,23 +54,23 @@ impl PolicySet {
...
-                .entry(block_number)
-                .or_default()
-                .push(PolicySetUpdate {
-                    account: user,
-                    in_set,
-                });
+            return;
         }
+
+        self.observed.insert(user);
```
0xKitsune, line 66: IIUC the early return skips `observed.insert` means that any user who was added to a whitelist/blacklist before the node started up will not be cached. Lmk if I am overlooking something here.

0xKitsune (approved):

crates/tempo-zone/src/l1_state/tip403/policy_set.rs
mattsse: yeah that makes sense, if we need this data we'd need to fetch it on demand but this prevents updating the cache

PR #429, opened 2026-04-23
feat(zones): symmetric deposit bounce-back implementation

dankrad:
## Summary

Implementation half of the split from #360. Depends on the spec change in **#428 (dankrad/deposit-bounceback-spec-only)**; review/merge that first.

Adds the symmetric deposit bounce-back mechanism: deposits now carry a \`bouncebackRecipient\`, and if zone-side processing fails the escrowed funds are refunded to that address on Tempo via a zero-fee, zero-callback withdrawal.

## What changes

**Reference Solidity contracts (\`specs/ref-impls/src/zone/\`)**

- \`IZone.sol\`: adds \`bouncebackRecipient\` to \`Deposit\` and \`EncryptedDeposit\`; adds \`bouncebackRecipient\` to \`deposit\` / \`depositEncrypted\` / \`DepositMade\`; adds \`DepositFailed\` (inbox) and \`DepositBounceBack\` (portal) events; renames \`BounceBack\` → \`WithdrawalBounceBack\`.
- \`ZonePortal.sol\`:
    - \`deposit\` / \`depositEncrypted\` accept \`bouncebackRecipient\`; the portal validates it against the token's TIP-403 policy at deposit time so the eventual refund transfer is guaranteed to be accepted by that policy.
    - Renames the internal helper \`_enqueueBounceBack\` → \`_enqueueWithdrawalBounceBack\` and sets \`bouncebackRecipient = address(0)\` on bounce-back deposits so they cannot themselves bounce.
    - Emits \`WithdrawalBounceBack\` (renamed from \`BounceBack\`).
- \`ZoneInbox.sol\`:
    - New immutable \`IZoneOutbox outbox\` constructor argument used to enqueue deposit bounce-backs.
    - Wraps the final mint (regular, encrypted-with-valid-decryption, and encrypted-with-failed-decryption fallback) in \`try/catch\`. On revert, calls \`outbox.enqueueDepositBounceBack\` and emits \`DepositFailed\` / \`EncryptedDepositFailed\`.
    - Deposits with \`bouncebackRecipient = address(0)\` (i.e. portal-generated withdrawal bounce-backs) keep the old unconditional-mint behavior.
- \`ZoneOutbox.sol\`: new \`enqueueDepositBounceBack(token, amount, bouncebackRecipient)\`, restricted to the sequencer (matching other outbox bookkeeping calls). Enqueues a zero-fee, zero-callback, zero-\`fallbackRecipient\` \`PendingWithdrawal\`.
[...]

PR #428, opened 2026-04-23
spec: symmetric deposit bounce-back mechanism

dankrad:
## Summary

Splits out the spec portion of #360. This PR only touches `specs/spec.md` and adds a full prose description of the **symmetric deposit bounce-back mechanism**, plus the corresponding updates to the `IZonePortal` interface stub.

Companion implementation PR: **dankrad/deposit-bounceback-impl** (Solidity reference contracts, Rust bindings, tests, and genesis regeneration).

## Motivation

Today a withdrawal that reverts on Tempo (blocked callback, TIP-403 change, paused token, etc.) bounces back into the zone as a fresh deposit credited to `fallbackRecipient`. The deposit side had no such mechanism: if the zone-side mint reverts, the deposit queue gets stuck and the escrowed funds are effectively stranded in the portal.

TIP-403 makes this an operational concern rather than a corner case. The policy for a token can change between the time the user initiates the deposit on Tempo and the time the zone processes it, and the new policy may forbid minting to `deposit.to`. Without a recovery path, legitimate deposits can be bricked by a policy update between signing and processing.

This spec describes the deposit-side counterpart: every deposit carries a `bouncebackRecipient`, and if zone-side processing fails the funds are refunded to that address on Tempo via a zero-fee, zero-callback withdrawal.

## What changes in the spec

1. A new section **Deposit Failures and Bounce-Back** under Deposits, documenting:
   - Why the mechanism is needed (mostly driven by TIP-403 policy drift).
   - The new `bouncebackRecipient` argument to `deposit` / `depositEncrypted` and the TIP-403 validation the portal performs at deposit time to guarantee the refund transfer will succeed.
   - Triggering conditions: any mint revert during `ZoneInbox.advanceTempo` (including the fallback-to-sender mint after a failed decryption).
   - Zone-side handling: `ZoneInbox` catches the revert and calls `ZoneOutbox.enqueueDepositBounceBack(token, amount, bouncebackRecipient)`, which enqueues a zero-fee, zero-callback, zero-`fallbackRecipient` withdrawal.
   - Tempo-side handling: `ZonePortal.processWithdrawal` refunds the escrowed tokens and emits `DepositBounceBack`.
   - The no-recursive-bounce rule: bounce-back deposits have `bouncebackRecipient = address(0)` and bounce-back withdrawals have `fallbackRecipient = address(0)`, so neither can spawn another bounce.
[...]

PR #441, opened 2026-05-05
spec: portal deposit TIP-403 policy + escrow on block

dankrad:
## Summary

Spec-only change. Adds a sequencer-controlled, portal-level TIP-403 policy that is enforced on the **depositor** (`msg.sender`) of every `ZonePortal.deposit` / `depositEncrypted` call, with a TIP-1028-style escrow path on block instead of a revert.

Stacked on #428 (deposit-bounceback spec) and shares its refund registry. The Solidity / Rust implementation lives in #429, which is the common implementation PR for both this policy and the bounceback spec from #428.

## Design

- **`depositPolicyId`** — sequencer-controlled `uint64` on `ZonePortal`, set via `setDepositPolicyId(uint64)` (emits `DepositPolicyUpdated`).
- **Default = policy 1** — the TIP-403 registry's built-in empty BLACKLIST (allow-all), so deposits work out of the box without the sequencer doing anything.
- **Policy 0 halts deposits** — built-in empty WHITELIST (reject-all). Independent of the per-token `pauseDeposits` flag, which remains for token-specific control.
- **Always enforced** — both `deposit` and `depositEncrypted` call `TIP403Registry.isAuthorizedSender(depositPolicyId, msg.sender)` after pulling the funds.
- **Escrow on block (no revert).** When the sender fails the policy:
  - Funds stay in the portal and are credited to `_refunds[token][bouncebackRecipient]` (the same registry from #428).
  - `DepositBlocked(token, sender, amount, depositPolicyId)` is emitted.
  - No fee is charged, the deposit is not enqueued, and `bouncebackRecipient` recovers via `claimRefund(token)`.
- This keeps `deposit(...)` callable from contracts that rely on it returning success (analogous to TIP-1028's escrow path for receiver-side blocks). The trade-off is documented: the portal policy provides liveness, not asset confiscation — a blocked depositor can name any unblocked address as `bouncebackRecipient` and pull the funds back.

## Spec changes (`specs/spec.md`)

- **Regular Deposits** flow: new step 5 describing the sender check, escrow on block, and `DepositBlocked` emission. Remaining steps renumbered.
- **Deposit Failures and Bounce-Back**: extends the failure catalogue from three to four modes; new bullet for the deposit-time portal policy block (the only failure path that doesn't round-trip through the zone).
- **Policy Enforcement on Zones**: new *Portal deposit policy* subsection covering the setter, the policy-1 / policy-0 sentinels, and the liveness-not-confiscation trade-off vs. token-level TIP-403.
- **`IZonePortal` stub**: adds `event DepositPolicyUpdated`, `event DepositBlocked`, `setDepositPolicyId`, `depositPolicyId()`. Reuses `refunds` / `claimRefund` / `RefundClaimed` from #428.

No Solidity, Rust, or test changes here — those go in #429.

## Test plan

- [ ] CI on `dankrad/deposit-policy-check` (markdown lint / link check / typos).
- [ ] Implementation + tests land on #429.

### DIFF
--- a/specs/spec.md
+++ b/specs/spec.md
@@ -279,9 +279,10 @@ A user deposits by calling `deposit(token, to, amount, memo, bouncebackRecipient
 2. Requires `bouncebackRecipient != address(0)` (reverts otherwise).
 3. Snapshots the deposit fee at the current `zoneGasRate` and the bounce-back fee at the current portal-side `tempoGasRate` (see [Deposit Fees](#deposit-fees)), and requires `amount >= depositFee + bouncebackFee` (reverts `DepositTooSmall` otherwise). The bounce-back fee covers the worst-case Tempo gas of paying out a refund (including new-account creation for `bouncebackRecipient`), so it is priced in Tempo gas, not zone gas.
 4. Transfers `amount` from the user into the portal.
-5. Pays the `depositFee` to the sequencer immediately. The `bouncebackFee` is reserved on the queued entry; it is only consumed if the deposit later bounces back.
-6. Appends the deposit to the deposit queue hash chain with the net amount (`amount - depositFee`), `bouncebackRecipient`, and the snapshotted `bouncebackFee`.
-7. Emits `DepositMade`.
+5. Enforces the portal's deposit-side TIP-403 policy on `msg.sender` via `isAuthorizedSender(depositPolicyId, msg.sender)` (see [Policy Enforcement on Zones](#policy-enforcement-on-zones)). If the depositor is blocked the portal does **not** revert: it credits `amount` to `_refunds[token][bouncebackRecipient]`, emits `DepositBlocked`, and returns. No fee is charged, the deposit is not enqueued, and `bouncebackRecipient` can later pull the funds via `claimRefund(token)`. This keeps `deposit(...)` callable from contracts that rely on it returning success (analogous to TIP-1028's escrow path for receiver-side blocks). See [Deposit Failures and Bounce-Back](#deposit-failures-and-bounce-back) for the full refund-registry semantics.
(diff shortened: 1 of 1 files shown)


specs/spec.md
```diff
@@ -279,9 +279,10 @@ A user deposits by calling `deposit(token, to, amount, memo, bouncebackRecipient
...
 4. Transfers `amount` from the user into the portal.
-5. Pays the `depositFee` to the sequencer immediately. The `bouncebackFee` is reserved on the queued entry; it is only consumed if the deposit later bounces back.
-6. Appends the deposit to the deposit queue hash chain with the net amount (`amount - depositFee`), `bouncebackRecipient`, and the snapshotted `bouncebackFee`.
-7. Emits `DepositMade`.
+5. Enforces the portal's deposit-side TIP-403 policy on `msg.sender` via `isAuthorizedSender(depositPolicyId, msg.sender)` (see [Policy Enforcement on Zones](#policy-enforcement-on-zones)). If the depositor is blocked the portal does **not** revert: it credits `amount` to `_refunds[token][bouncebackRecipient]`, emits `DepositBlocked`, and returns. No fee is charged, the deposit is not enqueued, and `bouncebackRecipient` can later pull the funds via `claimRefund(token)`. This keeps `deposit(...)` callable from contracts that rely on it returning success (analogous to TIP-1028's escrow path for receiver-side blocks). See [Deposit Failures and Bounce-Back](#deposit-failures-and-bounce-back) for the full refund-registry semantics.
```
mattsse, line 282: this makes sense to me

mattsse (approved):
this should be a L1 portal contract only change I believe (`ITIP20.isAuthorizedSender`)

0xKitsune (approved):

PR #449, opened 2026-05-08
refactor(policy-cache): update policy cache logic

0xKitsune:
Refactors TIP-403 policy cache internals into modules for event decoding and policy set tracking.

Also cleans up naming around policy state and fixes stale comments about cache seeding.


mattsse (approved):
lgtm

PR #450, opened 2026-05-11
[codex] add zone RPC URL metadata

danrobinson:
## Summary
- add ZonePortal `zoneRpcUrl` storage plus `setZoneRpcUrl(string)` with `ZoneRpcUrlUpdated(address indexed updater, string newZoneRpcUrl)`
- allow empty strings, cap URLs at 256 bytes, require an ASCII-case-insensitive `https` scheme, and intentionally do not validate after the first `:` to mirror the TIP-1026 scheme-validation style
- thread the URL through `ZoneFactory.createZone`, `ZoneCreated`, `ZoneInfo`, Rust ABI bindings, `create-zone`, `zone-info`, and a new `set-zone-rpc-url` xtask/Just recipe

## Validation
- `cargo check -p zone-primitives -p tempo-xtask`
- `forge build --skip test`
- `forge test --match-path 'test/zone/Zone{Factory,Portal}.t.sol'` currently compiles, then fails in local test setup because this machine’s `forge` does not recognize the repo’s Tempo hardfork/precompile config: `MissingPrecompile("AccountKeychain", 0xaAAAaaAA00000000000000000000000000000000)`.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -112,7 +112,7 @@ zone-info identifier:
     cargo run -p tempo-xtask -- zone-info {{identifier}}
 
 [group('zone')]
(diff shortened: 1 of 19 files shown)


0xKitsune:
Closing as this is already implemented.

PR #474, opened 2026-06-17
refactor(tempo-zone): clean up node launch and consolidate L1 state modules

0xKitsune:
This PR cleans up various components of the `tempo-zone` including launching add-ons, l1 state logic and TIP403 prefetch tasks.


crates/tempo-zone/src/node.rs
```diff
@@ -254,8 +244,6 @@ pub struct ZoneAddOns<N: FullNodeComponents<Types = ZoneNode, Evm = ZoneEvmConfi
     policy_cache: PolicyCache,
     /// ZonePortal address on L1.
     portal_address: Address,
-    /// Pre-configured list of initial tokens.
-    initial_tokens: Option<Vec<Address>>,
```
mattsse, line 258: I believe this is now breaking the tests

mattsse (commented):
not sure I fully understand all of these, could we perhaps break this down because unsure how to review this or what the motivation here is

0xKitsune:
> not sure I fully understand all of these, could we perhaps break this down because unsure how to review this or what the motivation here is

The motivation was to make a flows clearer, reduce verbosity and consolidate some of the parts around node launch, L1 state modules, and related prefetch/task logic.

I agree this ended up touching a few separable areas and can break this in to smaller PRs.

PR #477, opened 2026-06-17
Clarify zone Tempo proof lookback requirements

decofe:
## Summary
- Clarifies that Tempo state reads require trie proofs for the finalized Tempo state root, not arbitrary historical state proofs
- Documents the EIP-2935 direct anchor lookback limit: 8192 blocks, with the current 360-block safety margin yielding a 7832-block effective window

## Testing
- Not run; docs-only change


decofe:
Closing because the thread was informational only; no spec change requested.

PR #479, opened 2026-06-17
fix: reuse tempo eth api for zones

mattsse:
## Summary

- Move Tempo dependencies to commit `19631ec55ccbb9e8996b3d1481332d06d1ef5803`, which contains the generic `TempoEthApi`/`TempoEthApiBuilder` support needed by zones while staying on the current Reth/Alloy line.
- Replace the local zone Eth API builder with `tempo_node::rpc::TempoEthApiBuilder<N>` so zones reuse Tempo's balance, nonce-key, receipt-conversion, and fee-token gas estimation behavior.
- Add regression coverage to the deposit-then-transfer path for Tempo's placeholder native balance and successful TIP-20 transfer gas estimation before submitting without a manual gas limit.

## Why

Zone deposits credit the TIP-20 fee token balance, but the previous public Eth API path still used native-balance gas allowance semantics during estimation. That left deposited users with a zero native balance and caused ordinary in-zone transfers to fail before execution.

Tempo's Eth API already encodes the intended semantics: `eth_getBalance` returns the Tempo placeholder value and gas allowance is calculated from the configured fee token balance. Reusing the upstream builder keeps zones aligned with Tempo instead of carrying a local API implementation.

## Impact

Deposited pathUSD users can submit standard in-zone TIP-20 transfers through the public RPC without needing a separate native balance. The dependency is pinned to the untagged Tempo commit that introduced the generic builder because the current later Tempo `main` revs also require a broader Reth/Alloy bump.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10971,7 +10971,7 @@ dependencies = [
 [[package]]
 name = "tempo-alloy"
 version = "1.8.0"
-source = "git+https://github.com/tempoxyz/tempo?tag=v1.9.0#948196c8a551c520a2e88e36e91f271cff0cd34e"
+source = "git+https://github.com/tempoxyz/tempo?rev=19631ec55ccbb9e8996b3d1481332d06d1ef5803#19631ec55ccbb9e8996b3d1481332d06d1ef5803"
 dependencies = [
  "alloy-consensus",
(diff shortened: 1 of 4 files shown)


0xKitsune (approved):
Nice, lgtm

PR #481, opened 2026-06-17
refactor(l1): separate `l1.rs` into submodules 

0xKitsune:
This PR breaks  `l1.rs` into specific submodules for subscriber, deposit, event, block and queue logic  under a new `l1` module.


mattsse (approved):
lgtm

PR #484, opened 2026-06-19
feat: add bounceback recipient

0xKitsune:
This PR adds a new `bouncebackRecipeint` field to `Deposit` and `EncryptedDeposit` events.


specs/ref-impls/src/zone/ZonePortal.sol
```diff
@@ -498,7 +484,8 @@ contract ZonePortal is IZonePortal {
         address _token,
         address to,
         uint128 amount,
-        bytes32 memo
+        bytes32 memo,
+        address bouncebackRecipient
```
mattsse, line 488: do we need something like

```
if (bouncebackRecipient == address(0)) revert MissingBouncebackRecipient();
```
As defined in the spec:

Primary rule: [specs/spec.md#L464](https://github.com/tempoxyz/zones/blob/47ffd91be3cc30ab3433f69f0047fe9bf46d75a6/specs/spec.md#L464)
Regular deposit flow: [specs/spec.md#L328-L332](https://github.com/tempoxyz/zones/blob/47ffd91be3cc30ab3433f69f0047fe9bf46d75a6/specs/spec.md#L328-L332)
Encrypted deposit flow: [specs/spec.md#L394](https://github.com/tempoxyz/zones/blob/47ffd91be3cc30ab3433f69f0047fe9bf46d75a6/specs/spec.md#L394)

mattsse (dismissed):
lgtm

specs/ref-impls/src/zone/ZonePortal.sol
0xKitsune: Yes will add this in a fast follow. This PR is only focused on adding the new param.

mattsse (approved):

PR #486, opened 2026-06-20
feat: implement deposit bouncebacks

0xKitsune:
This PR implements deposit bouncebacks. Deposits now  handle a `bouncebackFee` and require a nonzero `bouncebackRecipient`. Deposits bounce back through the zone outbox when minting, encrypted verification, or sequencer acceptance fails. B

Zone-side deposit processing catches failure/rejection and enqueues a bounceback:

  ```solidity
  if (qd.rejected) {
      _enqueueDepositBounceBack(d.token, d.amount, d.bouncebackRecipient, d.bouncebackFee);
      emit DepositRejected(...);
  } else {
      try IZoneToken(d.token).mint(d.to, d.amount) {
          emit DepositProcessed(...);
      } catch {
          _enqueueDepositBounceBack(d.token, d.amount, d.bouncebackRecipient, d.bouncebackFee);
          emit DepositFailed(...);
      }
  }

The outbox records this as a zero fee withdrawal with the reserved bounceback fee carried separately:

```solidity
  PendingWithdrawal({
      sender: address(0),
      txHash: bytes32(0),
      to: bouncebackRecipient,
      amount: amount,
      fee: 0,
      bouncebackFee: bouncebackFee,
      gasLimit: 0,
      fallbackRecipient: address(0),
      callbackData: "",
      revealTo: ""
  });
```

  The portal recognizes the sentinel withdrawal, pays the sequencer, and refunds net funds or parks them:

```solidity
  uint128 bouncebackFee = withdrawal.bouncebackFee;
  uint128 refundAmount = withdrawal.amount - bouncebackFee;

  if (bouncebackFee > 0) {
      ITIP20(_token).transfer(sequencer, bouncebackFee);
  }

  try ITIP20(_token).transfer(withdrawal.to, refundAmount) returns (bool ok) {
      success = ok;
  } catch {
      success = false;
  }

  if (!success) {
      refunds[_token][withdrawal.to] += refundAmount;
      emit DepositBounceBackPending(withdrawal.to, _token, refundAmount, bouncebackFee);
  }
```

Most of the core logic is in `ZoneInbox` and `ZoneOutbox`.


crates/primitives/src/abi.rs
```diff
@@ -178,7 +183,15 @@ macro_rules! define_abi {
...
+                    uint128 bouncebackFee
+                );
+
+                #[derive(Debug)]
+                event DepositBounceBackPending(
+                    address indexed bouncebackRecipient,
+                    address token,
+                    uint128 amount,
+                    uint128 bouncebackFee
                 );
```
0xKitsune: We should be able to collapse these into a single `DepositBounceBack` event with a `success` bool to represent if the bounceback on l1 failed.

crates/primitives/src/abi.rs
```diff
@@ -206,6 +219,8 @@ macro_rules! define_abi {
                 error InvalidTempoBlockNumber();
                 #[derive(Debug)]
                 error DepositPolicyForbids();
+                #[derive(Debug)]
+                error MissingBouncebackRecipient();
```
0xKitsune: nit: update to `InvalidBouncebackRecipient`.

crates/primitives/src/abi.rs
```diff
@@ -517,6 +546,35 @@ macro_rules! define_abi {
...
+
+                #[derive(Debug)]
+                event DepositRejected(
+                    bytes32 indexed depositHash,
+                    address indexed sender,
+                    DepositType depositType,
+                    address token,
+                    uint128 amount,
+                    address bouncebackRecipient
+                );
```
0xKitsune: Same thing here, we can probably collapse into a single event and add additional fields if necessary.

crates/primitives/src/abi.rs
```diff
@@ -517,6 +546,35 @@ macro_rules! define_abi {
...
+
+                #[derive(Debug)]
+                event WithdrawalBounceBackProcessed(address indexed fallbackRecipient, address token, uint128 amount);
+
+                #[derive(Debug)]
+                event WithdrawalBounceBackPending(address indexed fallbackRecipient, address token, uint128 amount);
```
0xKitsune: Same thing here, we could use a `success` bool and collapse these into one event.

specs/ref-impls/src/zone/ZonePortal.sol
```diff
@@ -727,6 +772,48 @@ contract ZonePortal is IZonePortal {
         }
     }
 
+    function _processDepositBounceBack(Withdrawal calldata withdrawal) internal {
+        address _token = withdrawal.token;
+        uint128 bouncebackFee = withdrawal.bouncebackFee;
+        uint128 refundAmount = withdrawal.amount - bouncebackFee;
+
+        if (bouncebackFee > 0) {
+            ITIP20(_token).transfer(sequencer, bouncebackFee);
```
0xKitsune, line 773: Could this fail if the sequencer is blacklisted by an enabled token?

crates/contracts/src/precompiles/zone_inbox.rs
```diff
@@ -86,6 +88,30 @@ crate::sol! {
...
+        );
+
+        event DepositRejected(
+            bytes32 indexed depositHash,
+            address indexed sender,
+            DepositType depositType,
+            address token,
+            uint128 amount,
+            address bouncebackRecipient
+        );
```
0xKitsune, line 107: We should be able to collapse these into a single event with a `success` bool to represent if the deposit failed.

crates/contracts/src/precompiles/zone_portal.rs
```diff
@@ -114,7 +118,14 @@ crate::sol! {
...
             uint128 amount,
-            bool success
+            uint128 bouncebackFee
+        );
+
+        event DepositBounceBackPending(
+            address indexed bouncebackRecipient,
+            address token,
+            uint128 amount,
+            uint128 bouncebackFee
```
0xKitsune, line 128: Same thing here, we can probably collapse into a single `DepositBounceBack` event with a `success` bool.

crates/contracts/src/precompiles/zone_inbox.rs
```diff
@@ -86,6 +88,30 @@ crate::sol! {
...
+
+        event WithdrawalBounceBackProcessed(address indexed fallbackRecipient, address token, uint128 amount);
+
+        event WithdrawalBounceBackPending(address indexed fallbackRecipient, address token, uint128 amount);
```
0xKitsune, line 111: Same thing here, we could use a `success` bool and collapse these into one event.

specs/ref-impls/src/zone/ZoneOutbox.sol
```diff
@@ -449,7 +490,7 @@ contract ZoneOutbox is IZoneOutbox {
     /// @param gasLimit L1 callback gas limit included in the fee
     /// @return fee The total fee in zone token units
     function _calculateWithdrawalFee(uint64 gasLimit) internal view returns (uint128) {
-        return uint128(WITHDRAWAL_BASE_GAS + gasLimit) * tempoGasRate;
+        return uint128(WITHDRAWAL_BASE_GAS + gasLimit) * _tempoGasRate();
```
mattsse: this doesnt look right if we set this to the basefee which is e18

so I think we need smth like

```
* _tempoGasRate() / 1e12
```

here to account for the different decimals

specs/spec.md
```diff
@@ -301,7 +301,7 @@ The sequencer configures two gas rates that determine fees for deposits, withdra
...
 
-`tempoGasRate` defaults to `TEMPO_T0_BASE_FEE = 10,000,000,000` token units per gas unit at zone genesis. All rates are denominated in token units per gas unit and fees are paid in the same token being deposited or withdrawn. The sequencer takes the risk on gas-price fluctuations: if actual costs exceed the fee collected, the sequencer covers the difference; if they are lower, the sequencer keeps the surplus.
+`tempoGasRate` defaults to `TEMPO_T1_BASE_FEE = 20,000,000,000` token units per gas unit at zone genesis. All rates are denominated in token units per gas unit and fees are paid in the same token being deposited or withdrawn. The sequencer takes the risk on gas-price fluctuations: if actual costs exceed the fee collected, the sequencer covers the difference; if they are lower, the sequencer keeps the surplus.
```
mattsse: imo this is very ambigous.

this defines 20,000,000,000 token units per gas unit.

but I assume we need smmth like

> tempoGasRate defaults to TEMPO_T1_BASE_FEE = 20,000,000,000, an 18-decimal fixed-point TIP20 amount per gas unit. Because all supported TIP20s use 6 decimals, fees are computed as ceil(gasAmount * gasRate / 1e12) and paid in the same TIP20 being deposited or withdrawn.

crates/contracts/src/precompiles/zone_inbox.rs
mattsse: it would need to be an indexed bool then I guess, tbh idk what the best practice is for this

maybe @legion2002 can give a recommendation here

mattsse (changes requested):
q re fees

introducing the tempoGasRate is actually an entirely different change?

because the units are wrong, this now requires patching the rate in every test.

should we tackle this separately?

specs/ref-impls/src/zone/ZonePortal.sol
mattsse: yes, same as the with the portal itself.

specs/spec.md
0xKitsune: Updated in https://github.com/tempoxyz/zones/pull/486/changes/ecc492723174e869a4bc8df228cba3d966c42bd9

specs/ref-impls/src/zone/ZoneOutbox.sol
0xKitsune: updated in https://github.com/tempoxyz/zones/pull/486/changes/3734ced7c4893f29d6e53fab9126de44734e48aa

0xKitsune:
> should we tackle this separately?

Sounds good, I removed the `tempoGasRate` from this PR and can introduce this in a subsequent PR.

mattsse (approved):
LGTM

mattsse (approved):
LGTM

PR #487, opened 2026-06-22
refactor(precompiles): move zone registration helper

mattsse:
## Summary

Move zone-specific precompile registration into `zone-precompiles` behind a documented `extend_zone_precompiles` helper. The helper owns the static address registrations and the dynamic TIP-20/Tempo lookup override, while `ZoneEvmFactory` now only forwards the precompile map, hardfork config, L1 provider, and optional policy provider.

This also moves `TempoStateReader` and `ZoneTxContext` into `zone-precompiles`. `TempoStateReader` now depends on a small provider trait so the node keeps its concrete cache/RPC implementation without creating a crate cycle.

## Impact

Precompile wiring now lives with the precompile implementations, which makes the zone registration path mirror Tempo's map-extension pattern and keeps the node EVM factory focused on construction rather than address routing.


mattsse:
Closing per request; this cleanup is not worth the churn.

PR #488, opened 2026-06-23
Poll finalized L1 blocks for zone imports

decofe:
## Summary
- replace L1 `newHeads`/latest-tip ingestion with finalized-block polling
- backfill only through the current finalized L1 block
- start live polling from the next unprocessed finalized height

## Tests
- cargo check -p tempo-zone
- cargo test -p zone l1::tests

Prompted by: @0xalpharush


0xalpharush:
https://github.com/tempoxyz/zones/pull/597

PR #489, opened 2026-06-23
chore: zone solidity cleanups

0xalpharush:


specs/ref-impls/src/zone/ZoneMessenger.sol
```diff
@@ -48,16 +48,14 @@ contract ZoneMessenger is IZoneMessenger {
 
     /// @notice Relay a withdrawal message. Only callable by the portal.
     /// @dev Transfers tokens from portal to target via transferFrom, then executes callback.
-    ///      If callback reverts, returns false (does not revert). The transfer is atomic
```
0xalpharush, line 51: this is not the case. caller uses try-catch to handle revert, so think it's fine to not return bool

specs/ref-impls/src/zone/ZoneMessenger.sol
```diff
@@ -67,33 +65,14 @@ contract ZoneMessenger is IZoneMessenger {
         external
         onlyPortal
     {
-        // Atomic transfer + callback via self-call
```
0xalpharush, line 70: it is atomic without the self-call as it's in the same CALL frame

specs/ref-impls/src/zone/ZoneInbox.sol
```diff
@@ -400,12 +404,8 @@ contract ZoneInbox is IZoneInbox {
         amount = refunds[token][msg.sender];
         refunds[token][msg.sender] = 0;
 
-        try IZoneToken(token).mint(msg.sender, amount) {
-            emit RefundClaimed(msg.sender, token, amount);
-        } catch {
-            refunds[token][msg.sender] = amount;
```
0xalpharush, line 394: useless to restore prior to revert

0xalpharush:
cyclops audit fast

grandizzy:
makes sense

mattsse (dismissed):
resolved conflitcs, lgtm

grandizzy (dismissed):

mattsse (dismissed):

mattsse (approved):

grandizzy (approved):

PR #490, opened 2026-06-23
feat: use `block.baseFee` for bouncebackFee

0xKitsune:
This PR removes the stored deposit `bouncebackFee` and instead computes the fee from Tempo's `block.basefee` when the bounceback is processed.

Deposit bounceback refunds happen on Tempo and in the normative case they are processed shortly after the failed deposit. If the sequencer is down or the bounceback is delayed, the fee is still computed from Tempo `block.basefee` when the refund is eventually processed. If gas has moved up, the user pays the higher refund fee out of the bounced amount. If the fee would exceed the bounced amount, it is capped at the amount so the queue cannot get stuck.

```solidity
function calculateBouncebackFee() public view returns (uint128 fee) {
    uint256 gasFee = uint256(FIXED_BOUNCEBACK_GAS) * block.basefee;
    fee = uint128((gasFee + TEMPO_BASE_FEE_SCALE - 1) / TEMPO_BASE_FEE_SCALE);
}
```

To ensure that a given deposit has enough funds to cover the bounceback fee, deposits still check that the amount is large enough for an immediate bounce-back:

```solidity
uint128 fee = calculateDepositFee();
uint128 bouncebackFee = calculateBouncebackFee();

if (amount < fee + bouncebackFee) revert DepositTooSmall();
```


0xKitsune:
cyclops audit fast

mattsse (approved):
LGTM

mattsse (dismissed):

mattsse (dismissed):

mattsse (approved):
lgtm

PR #491, opened 2026-06-23
fix(rpc): fail closed in eth_getFilterChanges

yilirong1992:
### Summary
`eth_getFilterChanges` on the private RPC could pass an unfiltered upstream
response through to the caller. This change makes it fail closed: it filters log
results per caller, passes block/transaction-hash results through, and errors on
any other shape instead of leaking a potentially private payload.

### Problem
`eth_getFilterChanges` applied the per-caller log privacy filter only when the
upstream result parsed as `Vec<Log>`. Any other payload fell through to an
unfiltered passthrough of the raw upstream response.

For a log filter the result *is* the privacy-sensitive `Vec<Log>`. So if the
payload failed to deserialize as `Vec<Log>` — an unexpected/extra field, or a
future change to the log schema — the method returned the raw upstream payload,
which can contain other accounts' logs. This is a fail-open guard on private
data, and it is inconsistent with `eth_getLogs` and `eth_getFilterLogs`, both of
which hard-error on a parse failure.

### Fix
Disambiguate the two valid result shapes explicitly:
- A **log filter** returns `Vec<Log>` → filtered per caller.
- A **block / pending-transaction filter** returns `Vec<B256>` hashes → no
  private state, passed through unchanged.

Any other shape now returns an error instead of leaking the raw payload.

### Testing
- `cargo +nightly test -p zone-rpc --lib proxy::` — pass (3 new tests:
  log filtering, block-hash passthrough, fail-closed on unexpected shape)
- `cargo +nightly clippy -p zone-rpc --all-targets` — clean
- `cargo +nightly fmt -p zone-rpc -- --check` — clean

### DIFF
--- a/crates/rpc/src/proxy.rs
+++ b/crates/rpc/src/proxy.rs
@@ -7,7 +7,7 @@
 use std::{collections::HashMap, sync::Arc};
 
 use alloy_network::ReceiptResponse;
-use alloy_primitives::{Address, Bytes, hex};
+use alloy_primitives::{Address, B256, Bytes, hex};
 use alloy_rpc_types_eth::{BlockId, BlockNumberOrTag, Filter, FilterId, Log, state::StateOverride};
 use alloy_sol_types::SolCall;
 use eyre::WrapErr;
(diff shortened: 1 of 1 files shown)


mattsse:
Closing as superseded by #497.

PR #492, opened 2026-06-23
fix(l1): capture same-block policy updates for newly-enabled tokens

yilimin999:
# fix(l1): capture same-block policy updates for newly-enabled tokens

## 1. Motivation

The L1 subscriber mirrors TIP-403 policy state into the zone by decoding events
from each L1 block's receipts. Correctness of this mirror depends on every
relevant policy event in a block being captured. This change addresses a case
in which a token's first transfer-policy update is not captured, leaving the
cache to disagree with L1 until an RPC fallback happens to repair it.

## 2. Root cause

Within `extract_events`, a TIP-20 `TransferPolicyUpdate` log is decoded only if
the emitting token already appears in `tracked_tokens` at the instant the log is
visited. `tracked_tokens` is, in turn, populated during the very same single
pass over the block's logs, as each portal `TokenEnabled` log is reached.

A token may be enabled and have its transfer policy set within one L1 block by
two distinct transactions. Because the relative order of logs from different
transactions is not constrained, the `TransferPolicyUpdate` log may precede the
`TokenEnabled` log. Under that ordering the policy update is discarded: at the
time it is visited the token is not yet tracked, and so the cache never records
the initial `TokenPolicyChanged`. Subsequent policy resolution is then forced
onto an L1 RPC round-trip, and cross-node behavior becomes a function of
intra-block log ordering.

## 3. Proposed change

Event extraction is restructured into two passes:

1. Portal events (deposits, sequencer transfers, token enables) and TIP-403
   registry policy events are decoded. As a side effect, `tracked_tokens` is
   grown to include every token enabled in the block.
2. TIP-20 transfer-policy updates are collected, by which point every token
   enabled in the block is already tracked.

Registry policy events and TIP-20 policy events are written to disjoint cache
keys; consequently, reordering the two classes of events relative to one another
preserves the resulting cache state.

## 4. Validation

| Command | Result |
| --- | --- |
| `cargo +nightly test -p zone --lib l1::tests::extract_events` | pass (new regression test) |
| `cargo +nightly test -p zone --lib l1::` | pass — 46 tests |
| `cargo +nightly clippy -p zone --lib` | no warnings |
| `cargo +nightly fmt -p zone -- --check` | no diff |

The regression test constructs a single receipt in which the
`TransferPolicyUpdate` log is ordered ahead of `TokenEnabled` and asserts that
the corresponding `TokenPolicyChanged` event is produced.

### DIFF
--- a/crates/tempo-zone/src/l1/subscriber.rs
+++ b/crates/tempo-zone/src/l1/subscriber.rs
@@ -534,7 +534,7 @@ impl L1Subscriber {
     }
 
     /// Extract portal and policy events from pre-fetched receipts (no RPC).
-    fn extract_events(
+    pub(crate) fn extract_events(
         &mut self,
         block_number: u64,
         receipts: &[tempo_alloy::rpc::TempoTransactionReceipt],
@@ -545,6 +545,9 @@ impl L1Subscriber {
         let mut portal_events = L1PortalEvents::default();
         let mut policy_events = Vec::new();
 
+        // First pass: decode portal events (deposits, sequencer transfers, token
+        // enables) and TIP-403 registry policy events. This also grows
+        // `tracked_tokens` with every token enabled in this block.
         for receipt in receipts {
             for log in receipt.logs() {
                 let addr = log.address();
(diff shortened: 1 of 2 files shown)


mattsse (commented):
we expect enabled tokens to be rare so this is something we can ignore

PR #493, opened 2026-06-23
chore: simplify deposit and bounceback events

0xKitsune:
This PR updates `ZoneInbox` and `ZonePortal` to report each Deposit/Withrawal result with a single event, rather than separate events indicating success/failure.

Prev:
```solidity
 event EncryptedDepositProcessed(
            bytes32 indexed depositHash,
            address indexed sender,
            address indexed to,
            address token,
            uint128 amount,
            bytes32 memo
        );

        event EncryptedDepositFailed(
            bytes32 indexed depositHash,
            address indexed sender,
            address token,
            uint128 amount
        );
````

New:
```solidity
  event EncryptedDepositProcessed(
            bytes32 indexed depositHash,
            address indexed sender,
            address indexed to,
            address token,
            uint128 amount,
            bytes32 memo,
            address bouncebackRecipient,
            bool success
        );
```


mattsse (dismissed):
makes sense,

since this is abi changing, I think we need to update bytecode in zone-test-genesis.json

this should be improved so that this happens automatically

mattsse (dismissed):

mattsse (dismissed):

mattsse (approved):

PR #494, opened 2026-06-23
feat: tip403 checks for deposit paths

0xKitsune:
This PR adds TIP-403 checks for deposit paths.

For encrypted deposits, the portal now checks the `bouncebackRecipient` before depositing into the zone.

```solidity
uint64 policyId = ITIP20(_token).transferPolicyId();

if (!TIP403_REGISTRY.isAuthorizedRecipient(policyId, bouncebackRecipient)) {
    revert ITIP20.PolicyForbids();
}
```

For regular deposits, the portal also checks the recipient and mint recipient before accepting the deposit.
```solidity
if (!TIP403_REGISTRY.isAuthorizedRecipient(policyId, to)) {
    revert ITIP20.PolicyForbids();
}

if (!TIP403_REGISTRY.isAuthorizedMintRecipient(policyId, to)) {
    revert ITIP20.PolicyForbids();
}
```

If a bounce-back later fails because the policy changed, funds are parked in the portal refund registry instead of blocking the queue.

This also reuses Tempo’s existing TIP-403 errors instead of using the current zone specific error types.

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -79,6 +79,7 @@ jobs:
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           repository: tempoxyz/tempo
+          ref: b78b440950a0fa31b5e55c8a2c81976131e44ba3
           path: tempo
           persist-credentials: false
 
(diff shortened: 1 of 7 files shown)


mattsse (approved):
LGTM

PR #497, opened 2026-06-23
fix(rpc): reject pending filter changes

mattsse:
## Summary
- Track private RPC filter ownership with the filter type that created each ID.
- Make `eth_getFilterChanges` accept only logs for log filters and block hashes for block filters.
- Explicitly disable `eth_newPendingTransactionFilter` and reject transaction-shaped filter changes.

## Context
`eth_getFilterChanges` previously filtered log-shaped results, then returned any other upstream result unchanged. Pending transaction filters are disabled by policy, but this hardens the proxy so transaction-shaped filter changes cannot pass through if an unexpected upstream shape reaches the private RPC path.


0xKitsune (approved):
Approving, we can probably simplify a bit and remove the `proxy.rs` logic but can discuss and do this in a follow on.

PR #502, opened 2026-06-23
add bounceback address to send-deposit

adityapk00:
This PR adds support for the bounceback address to the just `send-deposit` recipe (introduced in #486 ). Also print the response code and response if something goes wrong, which makes debugging slightly easier. 

- update README to set PRIVATE_KEY which is needed for the subsequent commands


adityapk00:
cc @0xKitsune and @mattsse because i couldn't figure out how to add reviewers

mattsse (approved):

PR #505, opened 2026-06-23
Align ZonePortal admin permissions

decofe:
## Summary
- add a ZonePortal admin role and wire it through ZoneFactory creation metadata/events
- restrict token enablement and deposit pause/resume to admin instead of sequencer
- update storage slot constants, bindings, CLI/test helpers, and permission regression tests

## Tests
- forge test --match-contract 'ZonePortalTest|ZoneFactoryTest'
- cargo check -q -p tempo-zone-contracts
- cargo check -q -p tempo-xtask
- cargo check -q -p tempo-zone --tests

Prompted by: @grandizzy


mattsse (changes requested):
I think these should be fixed before merge.

1. **Default ZoneFactory address is stale after the `createZone` ABI change.**

   [`CreateZoneParams` now includes `admin`](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/xtask/src/create_zone.rs#L139-L142), which changes the tuple type and therefore the `createZone` selector/calldata shape. The built-in Moderato factory constant still points at the existing deployment: [`MODERATO_ZONE_FACTORY`](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/xtask/src/zone_utils.rs#L19-L25). Default `create-zone`, `zone-info`, and any tool that decodes `ZoneInfo` through this new ABI will be incompatible with that old contract.

   **Suggested fix:** deploy a fresh `ZoneFactory` compiled from this PR, verify it, update `MODERATO_ZONE_FACTORY`, and update the corresponding docs/address table. If deployment is intentionally handled separately, this PR should at least not switch the default CLI ABI against the old default address.

2. **Rust portal slot constants still disagree with the Solidity layout.**

   Solidity now defines `admin` at slot `1` and `pendingSequencer` at slot `2`: [`IZone.sol`](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/specs/ref-impls/src/zone/IZone.sol#L396-L399). The shared Rust constant still exports `PORTAL_PENDING_SEQUENCER_SLOT` as slot `1`: [`constants.rs`](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/crates/primitives/src/constants.rs#L58-L66). Code that applies sequencer-transfer events to the L1 state cache will write pending sequencer data into the admin slot and leave the actual pending sequencer slot uncached.

   **Suggested fix:** update `zone_primitives::constants::PORTAL_PENDING_SEQUENCER_SLOT` to slot `2`, add/export `PORTAL_ADMIN_SLOT` if Rust needs it, and add a cross-language regression test so the Rust constants stay aligned with the Solidity storage constants.

3. **Separate-admin zones cannot use the existing token-enable tooling.**

   `create-zone` now accepts a separate [`--admin`](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/xtask/src/create_zone.rs#L84-L87), but the generated [`zone.json` does not record it](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/xtask/src/create_zone.rs#L221-L231). The `enable-token` recipe still requires and signs with `SEQUENCER_KEY`: [`Justfile`](https://github.com/tempoxyz/zones/blob/71e48a3729cb68942676b39a807adcdcda8306af/Justfile#L318-L339), while `ZonePortal.enableToken` is now `onlyAdmin`.

   **Suggested fix:** persist the admin address in `zone.json`, add an `ADMIN_KEY`/`adminKey` path for admin-only portal operations, and have `enable-token` plus the demo xtasks sign `enableToken`, `pauseDeposits`, and `resumeDeposits` with the admin key. Keep falling back to the sequencer key only for legacy/default zones where `admin == sequencer`.

grandizzy:
@mattsse pls recheck

mattsse (approved):
lgtm

mattsse (approved):

PR #520, opened 2026-06-24
fix(rpc): enforce minimum response time on scoped read methods

Dollyerls:
# fix(rpc): enforce minimum response time on scoped read methods

Was reading through the spec's "Timing Side Channels" section and noticed it
calls for something the server doesn't actually do yet:

> The RPC server enforces a minimum response time of 100ms on
> `eth_getTransactionByHash`, `eth_getTransactionReceipt`, `eth_getLogs`,
> `eth_getFilterLogs`, and `eth_getFilterChanges`.

These five fetch data *before* they check whether the caller is allowed to see
it, so how long they take to answer leaks information: "this hash exists but
isn't yours" comes back at a different speed than "this hash doesn't exist." On
a private chain that's a real existence oracle. The other scoped reads
(`eth_getBalance`, `eth_call`, `eth_sendRawTransaction`) authorize first, so the
spec explicitly exempts them — and so does this change.

Turns out the server already measured per-call latency, but only to feed a
metric — it never enforced a floor. So I added the floor where the metric is
taken: `dispatch_request`, which is the one spot both the HTTP and the WebSocket
paths funnel through. After the handler runs, if the method is one of the five,
sleep off whatever's left under 100ms. It pads error responses too, since the
error path can return faster than the data path and would otherwise leak the
same way.

### DIFF
--- a/crates/rpc/src/server.rs
+++ b/crates/rpc/src/server.rs
@@ -15,11 +15,9 @@ use axum::{
 };
 use std::{
     sync::Arc,
-    time::{Instant, SystemTime, UNIX_EPOCH},
+    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
-use tempo_contracts::precompiles::account_keychain::IAccountKeychain::{
-    KeyInfo, SignatureType as KeyInfoSignatureType,
-};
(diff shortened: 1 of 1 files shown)


0xKitsune:
Thanks for the PR, closing this as this was discussed in a separate PR https://github.com/tempoxyz/zones/pull/580#issuecomment-4896021249.

PR #521, opened 2026-06-24
fix(zone): validate L1 receipt roots

mattsse:
## Summary

Validate L1 receipts fetched by the zone subscriber against the corresponding L1 header `receiptsRoot` and logs bloom before decoding portal or policy events.

The subscriber now requests receipts by the observed block hash, converts the returned Tempo RPC receipt logs into consensus-log form, computes the receipt root from the returned `ReceiptWithBloom` values, and fails the fetch if the computed root or aggregate logs bloom differs from the header. This prevents the zone from enqueueing events from a receipt set that does not match the L1 block being processed.

## Impact

Live subscription and backfill paths both perform the same validation. Backfill still pipelines multiple blocks concurrently, but fetches each header before the receipt request so receipts can be requested by block hash and checked against the header root and bloom.

The tests cover matching roots, mismatched roots, tampered receipt bloom values, and header logs-bloom mismatches.

## Linear

Resolves ZONE-26


grandizzy (dismissed):
lgtm!

0xKitsune (approved):

PR #523, opened 2026-06-24
Fix: Add checks specified in the spec for withdrawals

adityapk00:
Fix spec.md <-> implementation drift. 

1. remove old method from rust zone_portal.rs ABI
2. add explicit `config.isEnabledToken(token)` check for withdrawal
3. `WithdrawalQueueLib.enqueue` should work for 0 withdrawal blocks when the
   queue  is full


mattsse (approved):
LGTM

PR #525, opened 2026-06-24
chore: update Moderato ZoneFactory default

mattsse:
## Summary

- Redeploy the shared Moderato ZoneFactory and update the built-in default to `0x3F07435187fB4B4d4A562138A93C0397D0734F2b`.
- Restore the xtask/Justfile default behavior for `create-zone`, `zone-info`, and `deploy-router`, with `--zone-factory` / `ZONE_FACTORY` still available as overrides.
- Update the ZoneFactory deployment record with the new transaction, block, and deployed timestamp.

## Impact

Zone creation and lookup commands can use the admin-aware shared factory without requiring callers to pass `ZONE_FACTORY` manually.

### DIFF
--- a/xtask/src/zone_utils.rs
+++ b/xtask/src/zone_utils.rs
@@ -16,6 +16,14 @@ use tempo_contracts::precompiles::ITIP20 as TIP20Token;
 use zone::abi::{ZoneInbox, ZonePortal};
 
 pub(crate) const L1_EXPLORER: &str = "https://explore.moderato.tempo.xyz/tx";
+/// Shared Moderato ZoneFactory.
+///
+/// `create-zone`, `deploy-router`, and `zone-info` use this as their default
+/// factory unless the caller overrides `--zone-factory` or `ZONE_FACTORY`, or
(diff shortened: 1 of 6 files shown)


xtask/src/zone_utils.rs
```diff
@@ -16,6 +16,14 @@ use tempo_contracts::precompiles::ITIP20 as TIP20Token;
 use zone::abi::{ZoneInbox, ZonePortal};
 
 pub(crate) const L1_EXPLORER: &str = "https://explore.moderato.tempo.xyz/tx";
+/// Shared Moderato ZoneFactory.
+///
+/// `create-zone`, `deploy-router`, and `zone-info` use this as their default
+/// factory unless the caller overrides `--zone-factory` or `ZONE_FACTORY`, or
+/// `zone.json` already provides a zone-specific value.
+/// Explorer: https://explore.moderato.tempo.xyz/address/0x3F07435187fB4B4d4A562138A93C0397D0734F2b
```
0xKitsune, line 24: nit: can probably remove this.

0xKitsune (approved):

PR #527, opened 2026-06-24
refactor(payload): move builder logic into separate crate

0xKitsune:
This PR moves Zone payload/builder related logic into a dedicated crate.


mattsse (approved):
lgtm

PR #530, opened 2026-06-24
refactor(sequencer): move sequencer logic to dedicated crate

0xKitsune:
This PR moves sequencer related logic to a dedicated crate.

We can clean this up even further in a followon, this PR is focused on simply moving related logic to a crate.


mattsse (approved):
LGTM

PR #531, opened 2026-06-24
refactor(rpc): remove HTTP proxy backend

mattsse:
## Summary

- Remove the standalone HTTP `ProxyZoneRpc` backend and its proxy module/export.
- Make the EthHandlers-backed private RPC implementation the canonical `ZoneRpc` type.
- Update zone launch and integration test helpers to construct `ZoneRpc` from `handle.eth_handlers()`.

## Why

The private RPC is launched with the zone node, so the backend should use the node's in-process reth handlers instead of keeping a separate forwarding proxy implementation around.

## Impact

Downstream code should use `zone::rpc::ZoneRpc` for the in-process zone backend. The `zone-rpc::ProxyZoneRpc` export is removed. `ZoneProvider` HTTP client behavior is unchanged.


0xKitsune (approved):
Nice, lgtm

PR #532, opened 2026-06-25
refactor(node): move node assembly in to zone-node crate

0xKitsune:
This PR moves node related logic to a dedicated crate.


mattsse (approved):
LGTM

PR #535, opened 2026-06-25
fix: restrict TempoStateReader callers

mattsse:
## Summary
Restricts the standalone `TempoStateReader` precompile to the `TempoState` wrapper only. Contracts that need Tempo L1 state must read through `TempoState`, which validates its caller and forwards reads with the current `tempoBlockNumber`.

## Root Cause
The reader precompile rejected delegatecall but accepted direct calls from unauthorized EVM callers, so user transactions could bypass the wrapper-only caller check and trigger L1 storage reads with caller-selected inputs.

## Impact
Unauthorized direct callers are rejected at the precompile boundary. Legitimate wrapper-forwarded reads from `TempoState` continue to work.


grandizzy (approved):
lgtm!

PR #537, opened 2026-06-25
fix: meter AES-GCM AAD gas

mattsse:
## Summary
Charges AES-GCM precompile gas for both ciphertext bytes and authenticated associated data bytes.

## Root Cause
The precompile decoded caller-controlled `aad` and passed it into AES-GCM authentication, but gas accounting only used `ciphertext.len()`. A caller could therefore make AAD arbitrarily larger without paying the existing per-byte precompile charge for that authenticated input.

## Impact
Calls with large AAD now scale gas with the amount of authenticated data processed, while normal no-AAD and ciphertext-only paths keep their existing gas behavior.


grandizzy (approved):
lgtm!

PR #539, opened 2026-06-25
fix: scope ZoneInbox refund calls

mattsse:
## Summary
Rejects private RPC `eth_call` requests that try to read `ZoneInbox.refunds(token, owner)` for an owner other than the authenticated caller.

## Root Cause
Private RPC simulation enforced the request `from` address but did not inspect sensitive ZoneInbox view calldata, so callers could target the inbox predeploy and read arbitrary refund owners.

## Impact
Non-owner refund reads now fail before execution with the existing account-mismatch error, while own-owner reads and sequencer reads retain normal behavior.


crates/node/src/rpc.rs
```diff
@@ -598,6 +616,8 @@ where
 
             zone_rpc::policy::enforce_from(&mut request, &auth)?;
             zone_rpc::policy::enforce_no_contract_creation(&request)?;
+            self.enforce_zone_inbox_refund_call_privacy(&request, &auth)
+                .await?;
```
0xKitsune: Should this in the same `zone_rpc::policy` crate as the other `enforce_` checks?

Maybe we can combine these functions into a single `enforce_authorized` function?

0xKitsune (approved):

PR #540, opened 2026-06-25
fix: scope private RPC log scans

mattsse:
## Summary
Requires private RPC log filters to include the authenticated caller in an eligible indexed topic before the query reaches the backend.

## Root Cause
Private log APIs previously scoped token addresses and whitelisted event topic0 before backend retrieval, but caller-specific topic eligibility was only enforced after logs were fetched. That allowed broad backend scans even though unrelated logs were filtered out of the response.

## Impact
`eth_getLogs`, installed log filters, and websocket log subscriptions now reject underscoped private log filters with invalid params. Callers can still query sender/owner-style visibility with the caller in topic1, or recipient/spender-style visibility for two-party events with the caller in topic2, while the existing final visibility filter remains in place.

### DIFF
--- a/crates/rpc/src/filter.rs
+++ b/crates/rpc/src/filter.rs
@@ -41,6 +41,10 @@ pub const WHITELISTED_TOPICS: [B256; 5] = [
     BURN_TOPIC,
 ];
 
+const TWO_PARTY_TOPICS: [B256; 3] = [TRANSFER_TOPIC, APPROVAL_TOPIC, TRANSFER_WITH_MEMO_TOPIC];
+const CALLER_SCOPED_FILTER_ERROR: &str =
+    "private log filter must include authenticated caller in topic1 or topic2";
+
 /// Returns `true` if `caller` appears in an eligible indexed-topic position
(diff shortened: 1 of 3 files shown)


crates/rpc/src/filter.rs
```diff
@@ -148,6 +152,37 @@ pub fn scope_filter(filter: &mut Filter) {
...
 }
 
+/// Scopes a user-supplied filter to whitelisted event topics and requires the
+/// authenticated caller to appear in an eligible indexed topic before backend
+/// log retrieval.
+pub fn scope_filter_for_caller(filter: &mut Filter, caller: &Address) -> Result<(), JsonRpcError> {
+    scope_filter(filter);
+    if filter.topics[0].len() == 1 && filter.topics[0].contains(&B256::ZERO) {
+        return Ok(());
+    }
```
0xKitsune, line 162: nit: I think this can just be:

```rust

if filter.topics[0] == FilterSet::from(B256::ZERO) {
    return Ok(());
}
```

0xKitsune (approved):

PR #542, opened 2026-06-25
fix: reject uninitialized zone TIP20 prechecks

mattsse:
## Summary

Adds the vanilla TIP-20 initialized-token guard to the zone TIP-20 wrapper before any zone-specific policy precheck runs.

## Root Cause

`ZoneTip20Token` decoded transfer and mint selectors and resolved TIP-403 policy state before delegating to the underlying `TIP20Token::call`. The underlying implementation rejects uninitialized token storage first, but the wrapper ordering let uninitialized TIP-20-prefix addresses reach policy resolution.

## Impact

Uninitialized zone TIP-20 addresses now return `TIP20Error::uninitialized()` before any policy lookup. Initialized tokens continue through the existing privacy, bridge-auth, fixed-gas, and policy checks before delegation.

### DIFF
--- a/crates/precompiles/src/ztip20.rs
+++ b/crates/precompiles/src/ztip20.rs
@@ -18,9 +18,10 @@ use alloy_sol_types::{SolCall, SolError, SolInterface};
 use revm::precompile::{
     PrecompileError, PrecompileHalt, PrecompileId, PrecompileOutput, PrecompileResult,
 };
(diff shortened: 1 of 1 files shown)


0xrusowsky (approved):

0xrusowsky (approved):

grandizzy (approved):
makes sense!

PR #543, opened 2026-06-25
fix: Update withdrawals spec to reflect implementation

adityapk00:
Update spec to match current implementation for withdrawals
  - describe the use of EMPTY_SENTINEL during the hashing
  - Add revert types
  - Update IZone event and solidity ABI in the spec to match current implementation


mattsse (approved):
LGTM

PR #546, opened 2026-06-26
test(zone): expand unit, fuzz & symex coverage with CI gate

grandizzy:
## What

Strengthens the Solidity test suite for the zone implementation and wires it into CI with a coverage gate.

### Tests
- Expands unit + fuzz tests across the zone contracts: `ZonePortal`, `ZoneInbox`, `ZoneOutbox`, `ZoneMessenger`, `ZoneConfig`, `TempoState`, `BlockHashHistory`, `DepositQueueLib`, `WithdrawalQueueLib`, `Secp256k1Lib`, `EncryptedDeposit`, `PrivateZoneToken`, `Verifier`.
- Adds native symbolic (symex) property tests in `ZoneSymbolic.t.sol` (11 `check_*` properties across 4 contracts).
- Adds mutation-guided tests (driven by `forge --mutate`) that close real gaps line/branch coverage cannot see — boundaries, storage-slot math, and accounting — rather than equivalent or cross-contract mutants:
  - `ZoneConfig`: 3-entry encryption-key read + above-sequencer membership check (73.7% -> 81.2% mutation score).
  - `ZoneInbox`: HMAC-SHA256 known-answer test (RFC 4231 + golden vectors across every key-length branch) and direct `_readEncryptionKey` tests via a harness exposing the internal helpers the encrypted-deposit suite only reaches through the mocked AES-GCM path (49.5% -> 78.2%).
  - `ZoneOutbox`: amount + fee burn accounting, callback-data size boundary, and the `length - head` pending clamp (80.1% -> 84.1%).
  - `ZonePortal`: gas-rate exact-maximum and encryption-key index-past-end boundaries.
  - `TempoState`: already scores 100%, no changes.
- Splits `TempoState` header encoding in the test so `forge coverage --ir-minimum` no longer overflows the stack.

### CI (`specs.yml`)
- Builds the Tempo-capable `forge` once and reuses it for unit/fuzz, symex, and coverage in a single job.
- Installs `z3` and runs the `--symbolic check_*` property suite after the unit/fuzz tests.
- Runs `forge coverage --exclude-tests` (so the summary lists only production `src/zone` contracts) and fails the build if `src/zone` drops below 88% lines / 60% branches.

## Coverage
`src/zone`: 89.86% lines (771/858), 64.53% branches (151/234).

### DIFF
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -129,7 +129,7 @@ jobs:
         working-directory: specs/ref-impls
         run: |
           "$GITHUB_WORKSPACE/foundry/target/release/forge" coverage \
-            --ir-minimum --match-path "test/zone/**" --report summary | tee coverage.txt
+            --ir-minimum --match-path "test/zone/**" --exclude-tests --report summary | tee coverage.txt
           {
             echo '### Zone spec coverage'
             echo '```'
--- a/.github/workflows/specs.yml
+++ b/.github/workflows/specs.yml
@@ -24,7 +24,6 @@ permissions:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTC_WRAPPER: "sccache"
 
 jobs:
   build:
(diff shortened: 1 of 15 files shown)


mattsse (approved):
LGTM

PR #552, opened 2026-06-26
feat: align submitBatch between zone and tempo L1

adityapk00:
Fix withdrawal batching so L1 submitBatch uses the same withdrawal batch hash and index that are produced on-chain by
  ZoneOutbox.finalizeWithdrawalBatch.

Previously, finalizeWithdrawalBatch ran every zone block, but the monitor submitted multiple zone blocks as one L1
  batch. That meant L2 advanced withdrawalBatchIndex once per block while L1 advanced once per submitBatch, and the
  monitor recomputed a combined withdrawal hash off-chain that did not  match ZoneOutbox.lastBatch(). This
  worked with the mock verifier but is unprovable with a real verifier.

  - Move withdrawal batch boundary ownership to the payload builder.
  - Finalize a withdrawal batch when:
      - the open batch already has withdrawals, or
      - the current block creates withdrawals, or
      - every 60s

  - Use finalizeWithdrawalBatch(U256::MAX, ...) so finalizing blocks sweep the full pending batch of withdrawals
 
  - Treat BatchFinalized as the authoritative batch boundary and withdrawal hash source.
  - Update monitor/settlement logic to submit exactly one L1 submitBatch per BatchFinalized boundary, in order.
  - Preserve empty batch submissions so L2 and L1 withdrawalBatchIndex stay aligned.

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -10,6 +10,7 @@ use reth_ethereum::cli::Cli;
 use reth_tracing::tracing::info;
 use tempo_chainspec::spec::{TempoChainSpec, TempoChainSpecParser};
 use zone_evm::ZoneEvmConfig;
+use zone_payload::DEFAULT_WITHDRAWAL_BATCH_INTERVAL;
 
 use crate::{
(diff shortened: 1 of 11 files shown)


mattsse (approved):
makes sense

PR #553, opened 2026-06-26
fix: Handle fee transfer failure in bouceback

adityapk00:
If the deposit fee transfer to the sequencer fails for whatever reason (eg policy blacklist) then the whole withdrawal queue can become stalled. We previously fixed this in #529 for the normal withdrawal case, this does the same fix for the deposit bounceback case.


0xKitsune:
cyclops audit fast

mattsse (approved):

PR #558, opened 2026-06-29
fix: Update spec.md to accurately reflect the submitBatch() and withdrawals

adityapk00:
1. Update the spec to accurately reflect the submitBatch() and withdrawals flow
2. Add a solidity check in ZonePortal.sol to assert some invariants for safety (while we don't have a verifier)


mattsse (approved):
LGTM

PR #559, opened 2026-06-29
fix: Simplify the builder.rs rules for finalizing withdrawal batches

adityapk00:
- Remove the `has_current_withdrawals` check, which was redundant.
- Refactor common test utilities so they are reused
- add tests to test sequencer crash/restart during `submitBatch()` for pending withdrawal edge cases


mattsse (approved):
LGTM

PR #563, opened 2026-06-30
fix: Remove MockEIP2935 

adityapk00:
Cherry-pick the Mock 2935 removal code from #557 from @grandizzy to fix tests


grandizzy (approved):
lgtm

PR #566, opened 2026-06-30
fix: Separate admin and sequencer in tests and scripts

adityapk00:
- in the startup scripts (xtask and just scripts) make sure to accept admin, sequencer separately
- save the 2 keys in the generated configs
- Separate admin and sequencer role in tests using separate keys
- add tests to ensure admin key can't do sequencer actions and sequencer key cant do admin actions.


mattsse (approved):
LGTM

PR #575, opened 2026-07-01
feat: Allow Admin keys to be rotated

adityapk00:
- Add a two-step rotation for admin keys (mirroring sequencer key rotation)
- Tests
- Add check to make sure admin/sequencer keys can't be renounced


mattsse (approved):
LGTM

PR #576, opened 2026-07-02
feat(precompiles): add tempo state precompile

0xKitsune:
This PR  updates `TempoState` to be  Zone precompile while preserving the existing ABI and storage layout. The precompile now handles header finalization, getters, and L1 storage reads directly, and genesis initializes its storage/code instead of deploying the Solidity contract.

After this PR lands, I will open follow-ons to combine the `TempoStateReader` into `TempoState`, slim `TempoState` storage variables down to only the fields we actually need, and update the spec/ref impl once that storage/API cleanup lands.


adityapk00:
LGTM !

adityapk00:
Looks good!

mattsse (approved):

PR #578, opened 2026-07-02
fix(xtask): include Tempo fork times in zone genesis

### LINKED ISSUE #459: tempo-xtask create-zone generates genesis.json without Tempo fork activation fields
### Describe the bug

`tempo-xtask create-zone` generates a zone `genesis.json` that omits the Tempo fork activation fields (`t0Time` through `t4Time`).

Because of that, a freshly created zone can boot with older protocol behavior unless the generated genesis is patched manually before the zone node is started.

This is easy to misdiagnose because zone creation itself succeeds, and the problem only shows up later as runtime/protocol behavior mismatches. In our case, the zone behaved as pre-T1C / pre-T3, which caused access-
key/keychain flows to fail in ways that initially looked like wallet or frontend issues.

### Steps to reproduce

Steps to reproduce

1. Run `tempo-xtask create-zone` to generate a new zone.
2. Inspect the generated `genesis.json`.
3. Check `.config` and verify that Tempo fork activation fields such as `t1cTime` and `t3Time` are absent.
4. Start the zone from that generated genesis.
5. Exercise behavior that depends on post-fork Tempo keychain/access-key semantics.
6. Observe that the zone is running older protocol behavior unless the genesis is manually patched.

A minimal example of the workaround we had to apply was:

```
jq '.config += {
  t0Time: 0,
  t1Time: 0,
  t1aTime: 0,
  t1bTime: 0,
  t1cTime: 0,
  t2Time: 0,
  t3Time: 0,
  t4Time: 0
}' genesis.json
```

After adding those fields, the zone behavior matched the expected post-fork behavior.

### Logs

```text

```

### Platform(s)

Linux (x86)

### Container Type

Not running in a container

### What version/commit are you on?

Built from source in a local checkout

### If you've built from source, provide the full command you used

_No response_

### Code of Conduct

- [x] I agree to follow the Code of Conduct

betterclever:
## Summary
- add Tempo fork activation fields to generated Zone genesis config
- set `t0Time` through `t8Time` to `0`, matching the existing genesis-time Ethereum fork activations

## Verification
- `cargo test -p tempo-xtask`
- generated a Zone genesis locally and confirmed `t0Time` through `t8Time` are present at `0`

Fixes #459

### DIFF
--- a/xtask/src/generate_zone_genesis.rs
+++ b/xtask/src/generate_zone_genesis.rs
@@ -308,7 +308,7 @@ impl GenerateZoneGenesis {
         genesis_alloc.entry(self.admin).or_default().balance =
             U256::from(1_000_000_000_000_000_000_000u128);
 
-        let chain_config = ChainConfig {
+        let mut chain_config = ChainConfig {
             chain_id: self.chain_id,
             homestead_block: Some(0),
             eip150_block: Some(0),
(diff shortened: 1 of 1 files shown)


betterclever:
@jenpaff is this a non-issue or the fix was wrong?

PR #579, opened 2026-07-02
fix: Redact gasUsed, blockSize, log_index etc... that leak tx counts from zones

adityapk00:
- Redact gasUsed, blockSize, log_index and other fields from zone block headers and logs that might allow an observer to guess zone activity and allow side channel attacks
- Move redaction from ws.rs upstream, so the ws.rs is now just plumbing making it easier to maintain (i.e., all info going into ws.rs will already be redacted)
- re-order log_index so user's wallets can continue to rely on (txhash, log_index) but not leak block activity
- update spec

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -11,12 +11,12 @@ use std::{
     time::Duration,
 };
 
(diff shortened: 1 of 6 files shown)


mattsse (approved):
LGTM

PR #580, opened 2026-07-02
fix: Make RPC calls take constant time

adityapk00:
- To prevent timing attacks the following methods should take at least 100ms (already defined in the spec)

```
 "eth_getTransactionByHash"
        | "eth_getTransactionReceipt"
        | "eth_getLogs"
        | "eth_getFilterLogs"
        | "eth_getFilterChanges"
```


adityapk00:
Closing since the changes are not worth the timing risk + the methods don't leak much anyway

PR #581, opened 2026-07-03
feat(zone): skip txpool during L1 catch-up

### LINKED ISSUE #214: feat(zone): skip pool transactions during L1 catch-up (no_tx_pool)
When a zone node restarts after downtime, the L1Subscriber backfills all missed L1 blocks into the DepositQueue and the ZoneEngine processes them sequentially. During this catch-up phase, every block currently includes pool transactions, which is wrong — they're evaluated against stale state and slow down sync.

## Proposal

Add a `no_tx_pool: bool` field to `ZonePayloadAttributes` (like OP Stack's `noTxPool`). The engine sets it based on timestamp drift, and the builder skips pool transactions when the flag is set.

**Drift detection**: Compare the L1 block's timestamp to wall clock time. If `now - l1_block_timestamp > threshold`, set `no_tx_pool = true`. The threshold should be configurable via CLI arg with a sensible default (e.g. ~24s / 2 L1 block times).

**When `no_tx_pool` is true**, the zone block contains only system transactions (`advanceTempo` + `finalizeWithdrawalBatch`), no pool transactions. Once the zone catches up and the L1 block timestamp is within the threshold of wall clock time, pool transactions resume.

## Files to modify

- `payload.rs` — add `no_tx_pool: bool` to `ZonePayloadAttributes` and `ZonePayloadBuilderAttributes`
- `builder.rs` — guard pool tx inclusion with `if !attributes.no_tx_pool()`
- `engine.rs` — compute `no_tx_pool` from timestamp drift in `advance()`
- CLI args — add `--no-tx-pool-drift-threshold` flag

betterclever:
## Summary
- add a `no_tx_pool` payload attribute for stale L1 catch-up blocks
- skip user txpool transactions when L1 wall-clock drift exceeds the configured threshold
- derive drift from `TempoHeader::timestamp_millis()` so sub-second L1 timestamps are preserved
- default the drift threshold to 24 seconds for Tempo L1 catch-up; set it to 0 to disable the heuristic

Fixes #214

## Root Cause
During L1 backfill, zone payload building still attempted user txpool inclusion against stale L1 precompile state, so pool transactions could fail repeatedly and slow catch-up.

### DIFF
--- a/crates/node/src/node.rs
+++ b/crates/node/src/node.rs
@@ -73,6 +73,9 @@ use zone_sequencer::{BatchAnchorConfig, ZoneSequencerConfig, spawn_zone_sequence
 /// Network primitives for Zone Nodes
 type ZoneNetworkPrimitives = BasicNetworkPrimitives<TempoPrimitives, TempoTxEnvelope>;
 
+/// Default L1 timestamp drift before catch-up blocks skip txpool transactions.
+pub const DEFAULT_NO_TX_POOL_DRIFT_THRESHOLD: Duration = Duration::from_secs(1);
+
 /// Configuration for the sequencer background tasks
 #[derive(Debug, Clone)]
 pub struct ZoneSequencerAddOnsConfig {
@@ -88,6 +91,8 @@ pub struct ZoneSequencerAddOnsConfig {
     pub batch_anchor_config: BatchAnchorConfig,
     /// How often the withdrawal processor polls the L1 queue.
     pub withdrawal_poll_interval: Duration,
+    /// Maximum L1 timestamp drift before the engine skips txpool transactions.
+    pub no_tx_pool_drift_threshold: Duration,
 }
 
 /// Configuration for the Zone private RPC server extension.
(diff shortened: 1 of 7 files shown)


crates/node/src/node.rs
```diff
@@ -73,6 +73,9 @@ use zone_sequencer::{BatchAnchorConfig, ZoneSequencerConfig, spawn_zone_sequence
 /// Network primitives for Zone Nodes
 type ZoneNetworkPrimitives = BasicNetworkPrimitives<TempoPrimitives, TempoTxEnvelope>;
 
+/// Default L1 timestamp drift before catch-up blocks skip txpool transactions.
+pub const DEFAULT_NO_TX_POOL_DRIFT_THRESHOLD: Duration = Duration::from_secs(24);
```
betterclever, line 77: @mattsse Curious about this threshold. In the original issue the threshold is supposed to be 2 L1 block times. 24 doesn't fit that number based on current L1 block production time. Could you please explain?

PR #582, opened 2026-07-03
fix: seed zone token policies at local l1 height

mattsse:
## What changed
- Added `ZonePortal::enabled_tokens_at` so callers can resolve enabled tokens against a specific L1 block.
- Updated zone startup token discovery to read the portal at `policy_cache.last_l1_block()`, matching the block used by `seed_token_policies`.
- Made startup policy-cache seeding best-effort: discovery or seed failures now log a warning and continue instead of aborting node launch.
- Added regressions for stale-height token discovery and for startup continuing when policy seeding hits an uninitialized future token.

## Root cause
Startup discovered enabled portal tokens from latest L1 state, then seeded each token's `transferPolicyId()` at the node's persisted local L1 height. If a token had been enabled after that local height, the latest portal list included a future token whose TIP20 state was still uninitialized at the historical block, causing the node to shut down during policy seeding.

## Impact
A zone node with stale local L1 progress can restart without mixing future portal state into historical policy cache seeding. If the startup seed cannot be populated, the node logs the failure and continues; policy resolution can still fall back to L1 RPC on demand.

### DIFF
--- a/crates/node/src/node.rs
+++ b/crates/node/src/node.rs
@@ -56,7 +56,7 @@ use tempo_transaction_pool::{
 use tempo_zone_contracts::{
     TEMPO_STATE_ADDRESS, ZONE_INBOX_ADDRESS, ZONE_OUTBOX_ADDRESS, ZonePortal,
 };
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 use zone_evm::ZoneEvmConfig;
 use zone_l1::{
     DepositQueue, L1Subscriber, L1SubscriberConfig, PolicyCache, TempoStateExt,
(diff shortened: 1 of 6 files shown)


crates/node/src/node.rs
```diff
@@ -438,16 +438,47 @@ where
...
+        {
+            warn!(
+                target: "reth::cli",
+                %err,
+                count = tracked_tokens.len(),
+                %portal,
+                "Failed to seed token policies from L1; continuing with RPC fallback"
+            );
+            return Ok(());
+        }
```
0xKitsune, line 481: Should we fail here instead of continuing?

0xKitsune (approved):

crates/node/src/node.rs
mattsse: this can be fallible, we'd just encounter a cache miss on the first interaction

PR #584, opened 2026-07-06
fix: Read pending withdrawals directly from ZoneOutbox instead of recomputing

adityapk00:
- builder.rs was recomputing the pending withdrawals, but it can simply read from ZoneOutbox and get the authoritative list directly. Add view functions in ZoneOutbox that allow builder.rs to read the pending state directly, and delete ~300 lines of confusing builder.rs code
- finalizeWithdrawalBatch should pass the exact count of withdrawals instead of letting it silently clamping it.


0xKitsune:
cyclops audit fast

mattsse (approved):
LGTM

PR #586, opened 2026-07-06
chore(precompiles): remove `TempoStateReader`

0xKitsune:
This PR removes `TempoStateReader` now that `TempoState` is precompile for both checkpointing and Tempo storage reads. 

`TempoState` now directly serves `readTempoStorageSlot`/`readTempoStorageSlots` using the finalized Tempo block number.


mattsse (approved):
LGTM

PR #587, opened 2026-07-06
chore: Remove RPC timing side channels from spec

adityapk00:
Delete the side channel timing from the RPC section. The differences are not meaningful enough to matter

```
- eth_getTransactionByHash
      - Missing hash: mean 0.690ms, p99 0.854ms
      - Existing hash as non-owner, still returns null: mean 0.654ms, p99 0.805ms
      - Paired delta hidden-existing minus missing: mean -0.036ms, p99 0.062ms

  - eth_getTransactionReceipt
      - Missing hash: mean 0.662ms, p99 0.744ms
      - Existing hash as non-owner, still returns null: mean 0.673ms, p99 0.757ms
      - Paired delta hidden-existing minus missing: mean 0.011ms, p99 0.086ms
```

### DIFF
--- a/specs/spec.md
+++ b/specs/spec.md
@@ -55,7 +55,6 @@
     - [Method Access Control](#method-access-control)
     - [Block Responses](#block-responses)
     - [Event Filtering](#event-filtering)
-    - [Timing Side Channels](#timing-side-channels)
     - [WebSocket Subscriptions](#websocket-subscriptions)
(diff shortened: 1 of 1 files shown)


0xKitsune (approved):

0xKitsune (dismissed):
Should we add a note in the spec detailing why we do not need this? Might be useful to document this.

0xKitsune (approved):

PR #590, opened 2026-07-06
fix: Handle crash/timeout during process_queue

adityapk00:
when `process_queue` is handling withdrawals, if there's a crash/timeout, the next cycle was restarting at slot 0, which caused the resubmission to revert, causing a `repair_notify` which triggered a full portal resync.

Instead, we already have `find_processed_offset`, which can pinpoint the last processed slot, so we can recover and process properly. 

The code was technically correct, but was causing un-necessary reverts and re-processing during timeouts (which probably was causing further delays). Fix so we recover properly.

Additionally, clean up the L1 submission into `submit_and_confirm` so it is easier to understand and follow.


mattsse (approved):
LGTM

PR #594, opened 2026-07-07
feat: Wire tempoGasRate and maxWithdrawalsPerBlock to config

adityapk00:
Allow setting the zone gas rate and max withdrawals (`--zone.tempo-gas-rate` and `--zone.max-withdrawals-per-block`) and sync them to the ZoneOutbox during block building by injecting system txns in the zone block. 

Linear Issue: https://linear.app/tempoxyz/issue/CHAIN-952/wire-depositwithdrawal-fee-configuration-into-deployment

Note: This PR doesn't change defaults, which is 0 fee (free withdrawals) set at genesis.

### DIFF
--- a/crates/payload/src/builder.rs
+++ b/crates/payload/src/builder.rs
@@ -53,18 +53,49 @@ use crate::{ZonePayloadAttributes, ZonePayloadTypes};
 
 pub const DEFAULT_WITHDRAWAL_BATCH_INTERVAL: Duration = Duration::from_secs(60);
 
+/// Sequencer-configured `ZoneOutbox` fee parameters, applied on-chain via
+/// system transactions during block building.
+///
+/// The outbox's `tempoGasRate` and `maxWithdrawalsPerBlock` are zero at zone
+/// genesis (withdrawals are free). When configured, the payload
+/// builder compares the on-chain values at the start of each block and injects
+/// setter system txs to ensure they match.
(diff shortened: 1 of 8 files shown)


crates/payload/src/builder.rs
```diff
@@ -538,6 +576,154 @@ where
     }
 }
 
+impl<Provider, EvmConfig> ZonePayloadBuilder<Provider, EvmConfig> {
+    /// Compare on-chain `ZoneOutbox` fee parameters with the configured values
+    /// and inject setter system txs for any that differ.
+    fn sync_outbox_fee_config<B>(
```
mattsse, line 582: hmm do we even need this, can we not sync these changes via events just like the other updates?

adityapk00: The event paths are consuming L1 facts that have already happened (they're part of the next zone block's input). This is the opposite direction: the desired value comes from CLI arg, so there's no event yet. The `TempoGasRateUpdated` is emitted by the setter tx this fn decides whether to inject. I guess Events could tell the builder the current value,but the view call is a staticcall against the in-flight builder's own state  (which is the source of truth). The read is just an idempotence check so we only inject the setter when config and chain actually differ.

adityapk00: (Unless im misunderstanding your question)

PR #600, opened 2026-07-08
feat: zone fee manager

0xKitsune:
This PR implements a `ZoneFeeManager `that accepts any enabled TIP20 token for fees and settles fees directly in that token, without FeeAMM routing. This PR also removes the previous executor hack that overwrote validator fee token preferences per transaction.


crates/evm/src/fee_manager.rs
```diff
@@ -0,0 +1,210 @@
...
+        token.transfer_fee_post_tx(fee_payer, refund_amount, actual_spending)?;
+
+        if !actual_spending.is_zero() {
+            token.transfer(
+                TIP_FEE_MANAGER_ADDRESS,
+                ITIP20::transferCall {
+                    to: beneficiary,
+                    amount: actual_spending,
+                },
+            )?;
```
mattsse: I believe this is problematic, in case the token gets paused, then this fee collection would fail.

we have a test for this behaviour here

https://github.com/tempoxyz/tempo/blob/ac33b5b86a79edb1b9284450ba1a87fc645c4bb7/crates/node/tests/it/tip20.rs#L779-L930

so I guess we need some force transfer or expose _transfer

adityapk00: If the token is paused / TIP 403 blocked from paying the block builder, then the whole tx reverts and get skipped from the block, right? Maybe we are OK with that?

adityapk00: Oh wait, I think i misunderstood the evm code. I think the fee failure will actually cause the block builder to revert, causing the sequencer to stall. ignore my previous comment

PR #602, opened 2026-07-08
fix: Allow bounceback recipients to view deposit status

adityapk00:
Updates `zone_getDepositStatus` so the deposit `bouncebackRecipient` can see deposits where they are the refund target.

- This is needed for flows like `SwapAndDepositRouter`, where the L1 sender is the router and the target zone deposit bounces. Here,  the bounceback recipient needs to track whether the deposit processed successfully or failed and requires a refund claim.

  - Add `bouncebackRecipient` in portal deposit records decoded from `DepositMade` and `EncryptedDepositMade`.
  - Allow `bouncebackRecipient` to query  regular + encrypted deposit status 
  

Note:  For encrypted deposits, `bouncebackRecipient` is already a public portal event field, this change only lets that same party query status through private RPC.

Linear issue: https://linear.app/tempoxyz/issue/ZONE-27/bounceback-deposits-from-swapanddepositrouter-cannot-be-claimed

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -273,6 +273,7 @@ impl<Api: EthApiTypes + 'static> ZoneRpc<Api> {
                         deposit_hash: event.newCurrentDepositQueueHash,
                         sender: event.sender,
                         recipient: event.to,
+                        bounceback_recipient: event.bouncebackRecipient,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
makes sense

PR #606, opened 2026-07-09
feat(evm): disable CREATE and CREATE2

0xrusowsky:
this PR creates a new `ZoneEvm` wrapper around `TempoEvm` which disables the `CREATE` and `CREATE2` opcodes


crates/evm/src/lib.rs
```diff
@@ -52,10 +62,119 @@ use tempo_precompiles::{
...
+        // Ensure that none of the transaction top-level calls attempt contract creation.
+        if tx.kind().is_create()
+            || tx
+                .tempo_tx_env
+                .as_ref()
+                .is_some_and(|aa| aa.aa_calls.iter().any(|call| call.to.is_create()))
+        {
+            return Err(EVMError::Custom(
+                "contract creation not supported on zones".to_string(),
+            ));
```
pepyakin: Shouldn't this be CallsValidation? My clanker says this could potentially lead to an issue in a builder.

0xrusowsky: makes sense, my bad

0xrusowsky: addressed in [`7d6df90` (this PR)](https://github.com/tempoxyz/zones/pull/606/commits/7d6df900122e3041b5e2d2cf596df028b691f363)

crates/evm/src/lib.rs
```diff
@@ -52,10 +62,119 @@ use tempo_precompiles::{
...
+impl<DB: Database, I> ZoneEvm<DB, I> {
+    /// Creates a new `ZoneEvm` by disabling `CREATE` and `CREATE2` opcodes.
+    fn new(mut evm: TempoEvm<DB, I>) -> Self {
+        fn disabled<DB: Database>(_: ZoneInstructionCtx<'_, DB>) -> Result<(), InstructionResult> {
+            Err(InstructionResult::NotActivated)
+        }
+
+        let instructions = &mut evm.inner_mut().inner.instruction;
+        instructions.insert_instruction(CREATE, Instruction::new(disabled::<DB>), 0);
+        instructions.insert_instruction(CREATE2, Instruction::new(disabled::<DB>), 0);
```
mattsse: this will be a bit painful to maintain in case we need to whitelist deployments.

could we make this optional.
e.g. what if we want to allow deployments from a specific address, then we'd not a way to not disable them

mattsse (changes requested):
lets put this in another module and not

crates/evm/src/lib.rs
```diff
@@ -52,10 +62,119 @@ use tempo_precompiles::{
...
+    fn chain_id(&self) -> u64 {
+        self.inner.chain_id()
+    }
+
+    fn transact_raw(
+        &mut self,
+        tx: Self::Tx,
+    ) -> Result<ResultAndState<Self::HaltReason>, Self::Error> {
+        // Ensure that none of the transaction top-level calls attempt contract creation.
+        if tx.kind().is_create()
```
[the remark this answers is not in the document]
0xrusowsky: addressed in [`33e809c` (this PR)](https://github.com/tempoxyz/zones/pull/606/commits/33e809c50b33ecdc8afd875939a033b9539d1762)

crates/evm/src/lib.rs
0xrusowsky: addressed with [`33e809c` (this PR)](https://github.com/tempoxyz/zones/pull/606/commits/33e809c50b33ecdc8afd875939a033b9539d1762)

crates/rpc/src/policy.rs
```diff
@@ -52,19 +52,52 @@ pub fn enforce_from(
...
+        ));
+    }
+
+    Ok(())
+}
+
+/// Reject Ethereum and Tempo authorization lists that can mutate account code.
+pub fn enforce_no_code_authorization(
+    request: &TempoTransactionRequest,
+) -> Result<(), JsonRpcError> {
```
mattsse: doesnt this disable 7702 entirely?

why do we wan to disable them?

mattsse (commented):
The txpool still enforces the old creation rule. `ZonePoolBuilder` builds the Tempo validator with `TempoEvmConfig`, and the additional stateless check only calls `tx.is_create()`. For Tempo AA transactions, that only reflects the first call, so later create calls and both forms of authorization list can enter the pool even though `ZoneEvm::transact_raw` will reject them. Conversely, the unconditional check rejects every plain or first-call AA create before the new sender allowlist can apply, so an allowlisted deployer cannot submit that transaction through the normal pool path.

Suggested fix: expose the new pure transaction-policy function from `zone-evm` and call it here with `tx.tx_env()` instead of checking `tx.is_create()`. The pooled transaction already has the recovered sender, and `tx_env()` exposes the full AA call list plus Ethereum and Tempo authorization lists, so the same helper can enforce the allowlist and code-authorization rules in both `ZoneEvm::transact_raw` and pool admission. Map `TempoInvalidTransaction::CallsValidation` to an invalid-pool result and add pool tests for an allowlisted plain create, a create in a non-first AA call, and EIP-7702/Tempo authorization lists.

Simply passing `ZoneEvmConfig` to the builder is not sufficient with the current upstream implementation because `TempoTransactionValidator` constructs a `TempoEvm` internally; alternatively, that validator could be generalized to construct the zone EVM.

crates/rpc/src/policy.rs
0xrusowsky: initially i didn't modify it, but cyclops flagged it as a concern and i overindexed...

reverted since all deployed contracts will be "allowed" (either explicitly via protocol upgrade or implicitly via trusted deployer), so 7702 delegation shouldn't be a concern [`1086eab` (this PR)](https://github.com/tempoxyz/zones/pull/606/commits/1086eab035e5cb094090c91979a111cb60237f9e)

mattsse (dismissed):
LGTM

0xKitsune (dismissed):
Can we add a few E2E tests here? Otherwise lgtm.

0xKitsune (approved):

PR #612, opened 2026-07-10
docs(spec): use vectors for MPT node pools

decofe:
## Summary
- change `ZoneStateWitness.node_pool` and `BatchStateProof.node_pool` from `HashMap<B256, Vec<u8>>` to `Vec<Vec<u8>>`
- specify that the prover recomputes node hashes and builds its own hash-to-node index
- update all keyed-node-pool references across the spec

## Motivation
Prover inputs are untrusted, so the prover must recompute each node hash before relying on it regardless of whether a hash is supplied alongside the node. Omitting those redundant keys keeps the witness smaller without reducing verification work or security.

## Validation
- audited all MPT node-pool references in `specs/spec.md`
- ran `git diff --check`

Prompted by: @rkrasiuk

### DIFF
--- a/specs/spec.md
+++ b/specs/spec.md
@@ -1211,7 +1211,7 @@ pub struct ZoneStateWitness {
     pub state_root: B256,
 
     /// Deduplicated pool of all zone-state MPT nodes
-    pub node_pool: HashMap<B256, Vec<u8>>,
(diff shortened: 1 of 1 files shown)


adityapk00 (approved):
Makes Sense

PR #614, opened 2026-07-10
docs(spec): add `zone_id` to prover input

rkrasiuk:
## Summary

Adds `zone_id` to the prover's `PublicInputs`. The stateless verifier uses the zone ID to derive the zone’s EVM chain ID during execution. This ensures the EVM is initialized with the correct chain context for transaction validation and `CHAINID` opcode semantics.


adityapk00 (approved):
Makes sense

PR #617, opened 2026-07-10
fix: Use block numbers instead of wall clock for batch boundaries

adityapk00:
- Use block numbers to decide when to create and submit batches (if no withdrawals). Use block numbers makes the batch boundaries deterministic (instead of relying on wall clock), which makes it easier to debug + allows multiple instances to co-ordinate properly.


mattsse (approved):
lgtm

PR #618, opened 2026-07-10
feat: Make authenticated-withdrawal sender encryption deterministic

adityapk00:
Multi sequencer setups will become much easier if block building is deterministic. That is, if the block hash is a pure function of `(parent state, L1 block, ordered user transaction list)`. Given the same inputs, each sequencer should produce the exact same block hash, state root, receipts, batch boundary, and withdrawal data.

This PR changes the encrypted-sender ephemeral key derivation to be derived from the `private key, zone id and tx hash` so the ciphertext is reproducible by all sequencers in the zone


mattsse (dismissed):
LGTM

specs/spec.md
```diff
@@ -707,6 +707,8 @@ The `txHash` is the hash of the `requestWithdrawal` transaction on the zone. Sin
...
 
+Unlike user-created encrypted deposits, authenticated-withdrawal sender reveals are sequencer-created data that is hashed into the withdrawal queue. To keep zone blocks deterministic, the sequencer must not use fresh randomness when producing `encryptedSender`. It first derives a purpose-specific authenticated-withdrawal HMAC key from the registered sequencer encryption private key. Using that subkey, it derives the ECIES ephemeral scalar deterministically from the zone id, `revealTo`, `sender`, and `txHash`, retrying with a counter if the derived value is not a valid secp256k1 scalar. It derives the AES-GCM nonce from the same context plus the resulting ephemeral public key. For the same registered encryption key, zone id, reveal key, sender, and withdrawal transaction hash, `encryptedSender` is therefore byte-for-byte reproducible.
```
dankrad: I think the spec should include the formula for the derivation of the HMAC

mattsse (approved):

PR #625, opened 2026-07-13
feat: upgradability

legion2002:
Stacked on #609 
Moves per-zone metadata from implementation immutables into preserved slots 17-18 and replaces constructor arguments with factory-only, one-time initialization. `ZoneFactory.createZone` deploys and initializes atomically, while proxy tests cover unauthorized and repeated initialization.

Validated with `forge build --force -q`, `forge test --match-contract ZonePortalProxyStorageTest -vv`, `forge test --match-contract ZonePortalGasLimitTest -vv`, `forge inspect ZonePortal storage-layout`, and `git diff --check`.


specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -62,15 +62,14 @@ contract ZonePortal is IZonePortal {
     /// @notice Maximum allowed gas fee rate to prevent overflows
     uint128 public constant MAX_GAS_FEE_RATE = 1e18;
 
+    /// @dev The contract that deployed this implementation is the only initializer authority.
+    ///      This is protocol-wide code configuration and does not consume proxy storage.
+    address internal immutable _factory = msg.sender;
```
mattsse: [P1] Bind initialization to the actual factory, not the logic deployer

_factory = msg.sender only works while ZoneFactory itself deploys each ZonePortal. In the proxy path this change is meant to support, TIP-1091 etches one shared logic runtime and the enshrined factory later calls initialize through each proxy. The address that produced or patched the logic runtime is therefore not necessarily the initializer caller, so createZone can revert with NotFactory. The proxy test misses this because the test contract both deploys the implementation and calls the initializer.

Please make the authorized factory explicit, align it with the enshrined factory used by TIP-1091, and test with separate logic-deployer and factory addresses.

legion2002: addressed here - https://github.com/tempoxyz/zones/pull/625/changes/30162ec1e078dea3a6d97da7685f96e88636a56c

mattsse (dismissed):
LGTM

0xKitsune (dismissed):
Lgtm

Looks like tests are failing due to the dev-genesis needing to be regenerated. 

Opened a PR to fix: https://github.com/tempoxyz/zones/pull/655

0xalpharush:
see suggestions in https://github.com/tempoxyz/tempo/pull/6841/ and https://github.com/tempoxyz/zones/pull/656

0xKitsune (approved):

PR #626, opened 2026-07-13
feat(l1+precompiles): add L1-backed zone storage provider

0xrusowsky:
## High-level goal

before this PR stack, Zones had two representations of Tempo L1 state:
  1. A semantic `PolicyCache` / `PolicyProvider`, populated from typed TIP-403 and TIP-20 events with RPC fallback. Custom `ZoneTip20Token` and `ZoneTip403ProxyRegistry` implementations used it to reproduce upstream policy behavior.
  2. A generic block-versioned raw storage cache keyed by `(address, slot)`, used by `TempoState` and Zone system contracts for portal storage reads.

this duplicated policy semantics and left two L1-state paths that had to remain aligned in block anchoring, event decoding, invalidation, RPC fallback, and upstream behavior. The goal is to remove the semantic policy implementation from the EVM path and execute upstream Tempo precompiles against an L1-aware storage provider. The provider wraps the normal EVM storage provider and:
  - reads the Tempo L1 anchor from local `TempoState`
  - serves TIP-403 registry storage from L1 at that anchor
  - overlays only the L1-owned `transferPolicyId` field in TIP-20’s packed slot, preserving Zone-local fields
  - delegates ordinary storage to local Zone state
  - rejects writes to L1-owned mirrored state

from the upstream precompile's perspective these remain ordinary storage operations, but selected slots are virtualized to finalized L1 state. This makes upstream Tempo code the source of execution semantics rather than maintaining Zone-specific policy implementations.

## Raw L1 cache design

the existing generic cache has this shape: `(address, storage slot) → BTreeMap<L1 block number, storage value>`. A lookup at block `N` returns the latest cached value at or before `N`. The intended benefit is that mirrored policy state changes infrequently, so one fetched value can serve many later Tempo blocks without another synchronous L1 RPC call.

I added contract-level mutation barriers, which basically track when the cache for a given contract is invalidated: A value fetched at block `V` may serve block `N` only if there is no barrier in `(V, N]`. The subscriber installs barriers for ZonePortal logs, TIP-403 registry logs, and tracked TIP-20 policy updates. Exact sequencer-slot updates are applied after the portal barrier, so same-block values represent post-block state.

the cache now has two independent progress markers:
  - `anchor`: latest confirmed L1 block observed by the subscriber; it can run ahead while blocks are queued
  - `block_floor`: latest L1 block consumed by the Zone engine

the engine advances `block_floor` in O(1). Slot and barrier histories are pruned lazily when that specific slot/address is next mutated: retain the latest pre-floor baseline, retain entries at or above the floor, and remove older redundant history. Reads below the floor bypass inherited cache state because the pruned barrier history can no longer answer historical inheritance safely. Historical fallback results are not inserted back into the forward cache. Reorgs clear values, barriers, and the subscriber anchor while preserving the monotonic engine floor.

## Precompile Execution

`storage.rs` adds `ZonePrecompileStorageProvider`, a wrapper around Tempo's normal EVM precompile storage provider. It resolves the finalized L1 block once from local `TempoState`, serves TIP-403 storage reads from that exact L1 block, and overlays only the L1-owned `transferPolicyId` field in TIP-20's packed slot while preserving Zone-local fields. All other storage remains local. Mirrored reads still perform the local `SLOAD` first so warming, gas accounting, and storage-action recording are preserved, and writes to L1-owned state are rejected.

`execution.rs` centralizes local and L1-backed precompile execution. It installs the appropriate storage provider, preserves the active composed Tempo hardfork and storage accounting, and applies `CallRules` in two phases:
- cheap Zone-local admission checks run before resolving L1 state
- checks that need policy state run against the anchored overlay

`CallRules` also centralizes direct-call enforcement, fixed-gas handling, calldata gas, and forwarding of the original calldata and caller.

Together these changes let us drop the Zone-specific L1 precompile implementations from the execution path. TIP-20, TIP-403, and `TipFeeManager` now execute the upstream Tempo implementations directly against anchored L1-aware storage; the remaining Zone-specific behavior is expressed only as `CallRules` around that upstream execution.

## Follow-up work

`PolicyProvider` is no longer used by EVM precompile execution, but it is still retained for encrypted-deposit authorization in the engine/payload-builder path. To remove it completely, that path must be migrated to perform the same checks from the block-anchored raw L1 storage provider. Once that cutover is complete, we can remove the `PolicyProvider`/`PolicyCache` wiring from the node and subscriber, the policy prefetch task and typed policy-cache event plumbing, and the associated legacy tests and documentation.


0xKitsune:
cyclops audit

crates/l1/src/subscriber.rs
```diff
@@ -466,10 +472,10 @@ impl L1Subscriber {
         // A block is only flushed to the deposit queue once the NEXT block
         // arrives with a matching parent hash, proving the buffered block
         // is on the canonical chain.
+        #[allow(clippy::type_complexity)]
         let mut unconfirmed_tip: Option<(
             SealedHeader<TempoHeader>,
-            L1PortalEvents,
-            Vec<PolicyEvent>,
+            (L1PortalEvents, Vec<PolicyEvent>, HashSet<Address>),
```
0xKitsune: nit: seems like we can make an alias for this.

crates/l1/src/state/cache.rs
```diff
@@ -81,9 +96,30 @@ impl L1StateCacheInner {
...
+    }
+
+    /// Invalidates inherited slot values for `address` starting at `block_number`.
+    ///
+    /// Values subsequently inserted at the same block are post-block state and remain valid.
+    pub fn invalidate(&mut self, address: Address, block_number: u64) {
+        self.invalidations
+            .entry(address)
+            .or_default()
+            .insert(block_number);
```
0xKitsune: `invalidate` adds one entry to the `BTreeSet` for every block that mutates the cache, but `prune_before` is never called in production. This makes the invalidation history unbounded, can we connect pruning to the engine's L1 advancement path using the oldest L1 block the zone can still query as the boundary?

crates/l1/src/state/cache.rs
```diff
@@ -94,6 +130,64 @@ impl L1StateCacheInner {
...
+        block_number: u64,
+        activations: impl IntoIterator<Item = (u64, TempoHardfork)>,
+    ) {
+        self.hardfork_schedule.extend(activations);
+        self.hardfork_schedule_head = Some(
+            self.hardfork_schedule_head
+                .unwrap_or_default()
+                .max(block_number),
+        );
+    }
```
0xKitsune: Can you remind me why we need to track hardforks this way? If we make the zone chainspec wrap the tempo chainspec, can we just bump tempo deps after activation timestamps have been merged since the ZoneEVM wraps the TempoEVM?

crates/l1/src/state/cache.rs
0xrusowsky: addressed in [`16e88dd` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/16e88dd02be8c0bf0d87154eb541f2a4c00f5eca)

crates/l1/src/subscriber.rs
0xrusowsky: addressed in [`b35a82a` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/b35a82acb5408c96e38860412205ecae1a1dc634)

crates/l1/src/state/cache.rs
0xrusowsky: addressed in [`5ba8103` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/5ba81038733861aa7b6fd9ccaba97e120df2ceb8)

0xrusowsky:
cyclops audit

crates/l1/src/subscriber.rs
```diff
@@ -535,24 +536,25 @@ impl L1Subscriber {
...
         block_number: u64,
         receipts: &[tempo_alloy::rpc::TempoTransactionReceipt],
-    ) -> (L1PortalEvents, Vec<PolicyEvent>) {
+    ) -> SubscriberEventsOutput {
         use tempo_contracts::precompiles::{ITIP20::TransferPolicyUpdate, TIP403_REGISTRY_ADDRESS};
 
         let portal_address = self.config.portal_address;
         let mut portal_events = L1PortalEvents::default();
         let mut policy_events = Vec::new();
+        let mut mutated_accounts = HashSet::new();
```
0xrusowsky, line 557: addressed in [`8836842` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/8836842b2698c5249e8ee08a3b6b8b7a5711bae5)

crates/node/src/engine.rs
```diff
@@ -277,6 +281,11 @@ impl ZoneEngine {
         // could return wrong results.
         self.policy_provider.cache().advance(l1_num_hash.number);
 
+        // The engine has committed this L1 height, so no subsequent zone block can query an
+        // earlier height while building. Preserve only the baseline needed at this boundary and
+        // discard older slot, invalidation, and hardfork history accumulated by the subscriber.
+        self.l1_state_cache.write().prune_before(l1_num_hash.number);
```
0xrusowsky: i tried to address it by making the cache track a minimum floor block that advances with the engine (independent from the anchor of the subscriber).

then we lazily prune the cache slots based on the floor.

fix in [`de4fe49` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/de4fe49d7658bb6941ba8a874325a19dce5ff6e4)

0xrusowsky:
cyclops audit note="verify whether the proposed fix addresses the concern about 3rd party attacks via eth_call https://github.com/tempoxyz/zones/pull/626#discussion_r3590524183"

crates/node/src/engine.rs
0xrusowsky: alternatively i think we would have to distinguish untrusted (3rd-party) from trusted (builder) l1 rpc calls, and only update the cache in the latter

crates/l1/src/state/cache.rs
```diff
@@ -109,30 +158,38 @@ impl L1StateCacheInner {
...
+
+    let mut retained = history.split_off(&min_block);
+    if let Some(baseline) = history.pop_last() {
+        retained.insert(baseline.0, baseline.1);
+    }
+    *history = retained;
+}
+
+/// Retains the latest pre-floor barrier and every newer barrier.
+fn prune_invalidation_history(history: &mut BTreeSet<u64>, min_block: u64) {
```
0xrusowsky, line 197: addressed in [`01d536d` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/01d536da0ccba5c2616d0d1b7a92517b7854611c)

crates/l1/src/state/cache.rs
```diff
@@ -81,25 +100,55 @@ impl L1StateCacheInner {
...
     pub fn set(&mut self, address: Address, slot: B256, block_number: u64, value: B256) {
-        self.slots
-            .entry((address, slot))
-            .or_default()
-            .insert(block_number, value);
+        if block_number < self.block_floor {
+            return;
+        }
+
+        let history = self.slots.entry((address, slot)).or_default();
```
0xrusowsky, line 144: my understanding is that unbounded cache growth is a pre-existing issue.

while this PR initially amplified it by calling `prun_before()` on every block, now that it uses floor-driven lazy pruning, engine advancement is O(1), and only the slot/address being mutated is pruned.

IMO this PR achieves its objectives without adding extra cache work, so i think we should tackle the unbounded cache growth issue in a follow-up PR

0xrusowsky: closing after discussing it with mattsse and agreeing to address in a follow-up

crates/precompiles/src/lib.rs
```diff
@@ -40,20 +45,183 @@ pub mod dispatch {
...
+    // zone privacy, bridge-authorization, fixed-gas, and anchored policy-storage rules.
+    precompiles.set_precompile_lookup(move |address: &alloy_primitives::Address| {
+        if is_tip20_prefix(*address) {
+            Some(create_tip20_precompile(
+                *address,
+                &l1_env,
+                sequencer.clone(),
+            ))
+        } else if *address == TIP_FEE_MANAGER_ADDRESS {
+            Some(execution::create_l1_backed_precompile(
```
mattsse, line 132: **[P1] Apply the L1 overlay to protocol fee collection** — This wraps only external calls to the FeeManager address. Tempo transaction handling invokes `ProtocolFeeManager::collect_fee_pre_tx` directly under the ordinary EVM storage provider, so that path never reaches this lookup or `ZonePrecompileStorageProvider`. Zone tokens keep local `transferPolicyId = 1`; consequently a sender blacklisted by the anchored L1 policy can still pay the fee-token transfer and execute a transaction whose body does not call TIP-20. Please route the internal fee hooks through the same anchored overlay and add an empty-transaction regression test for an L1-blacklisted sender.

crates/evm/src/lib.rs
```diff
@@ -91,67 +72,16 @@ impl ZoneEvmFactory {
...
-        // must also handle the non-TIP-20 Tempo precompiles that are zone-relevant
-        // (FeeManager, NonceManager, AccountKeychain).
-        // Zone-specific overrides (TIP20Factory, TIP403Proxy) are in the
-        // static map via `apply_precompile` and take priority over this.
-        let zone_cfg = cfg.clone();
-        let zone_env = PrecompileEnv::new(
+        extend_zone_precompiles(
+            precompiles,
             &cfg,
+            self.l1_provider.clone(),
```
mattsse: **[P1] Do not install the dead L1 reader in executable CLI configs** — `new_without_l1()` supplies `127.0.0.1:1`; this provider now reaches every migrated L1-backed precompile through this argument, while `get_storage()` retries forever. `import`, `stage`, or `re-execute` will hang on the first replayed block with a TIP-20/TIP-403 storage read. Please give maintenance execution a usable historical L1 source or fail before execution instead of installing an indefinitely retrying reader.

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -1,334 +1,187 @@
...
+                decode_and_check::<ITIP20::allowanceCall>(args, |decoded| {
+                    self.check_allowance_read(decoded.owner, decoded.spender, call.caller)
+                })
+            }
+            IRolesAuth::hasRoleCall::SELECTOR => {
+                decode_and_check::<IRolesAuth::hasRoleCall>(args, |decoded| {
+                    self.check_balance_read(decoded.account, call.caller)
+                })
+            }
+            _ => Ok(()),
```
mattsse, line 106: **[P1] Preserve privacy for reward getters** — Both `userRewardInfo(address)` and `getPendingRewards(address)` now fall through this default branch, allowing an authenticated outsider to query another account reward recipient, index, and pending balance. The previous wrapper gated both using the balance-read rule, and the associated tests were removed. Add both selectors through `decode_and_check(... check_balance_read)` and restore owner/sequencer/outsider coverage.

crates/node/tests/it/tip403_policy.rs
```diff
@@ -100,14 +109,15 @@ async fn test_policy_proxy_whitelist_authorization() -> eyre::Result<()> {
...
-        let cache = zone.policy_cache();
-        let mut w = cache.write();
-        w.set_policy_type(5, ITIP403Registry::PolicyType::WHITELIST);
-        w.set_policy_status(5, alice, 1, true);
-        w.set_policy_status(5, bob, 1, false);
-    }
+    // Populate raw L1 state: policy 5 = WHITELIST, Alice is in the set, Bob is not.
+    seed_raw_tip403_policy(
+        zone.l1_state_cache(),
+        1,
```
mattsse, line 193: **[P1] Seed before advancing the cache floor** — This test waits until `tempoBlockNumber == 3` and then seeds at block 1. `L1StateCacheInner::set` discards writes below `block_floor` (now 3), so the first registry call misses and the local dummy L1 reader retries forever. The same ordering appears in the blacklist, compound, and block-versioned tests. Seed versions before advancing the floor or use the current accepted anchor; the helper should also surface rejected insertions.

crates/node/tests/it/utils.rs
```diff
@@ -3199,10 +3346,51 @@ impl L1Fixture {
...
         cache.update_anchor(NumHash {
             number: num_blocks,
             hash: B256::ZERO,
         });
+        drop(cache);
+        self.caches.lock().unwrap().push(cache_handle.clone());
+    }
+
+    /// Seed the absence of an address-level TIP-403 receive policy for the next fixture block.
+    pub(crate) fn seed_no_receive_policy(&self, recipient: Address) {
```
mattsse: **[P1] Seed receive-policy slots for all transfer fixtures** — Only the migrated policy test calls this helper. The unchanged transfer suite (`test_deposit_then_transfer`, sequential transfers, events, and memo) executes T6+ transfers to Bob or Charlie without seeding their receive-policy slots; estimates or execution miss the cache and retry the dummy RPC forever. Seed recipients before estimate/send, or make fixture misses finite and explicit.

crates/node/src/engine.rs
```diff
@@ -277,6 +281,13 @@ impl ZoneEngine {
         // could return wrong results.
         self.policy_provider.cache().advance(l1_num_hash.number);
 
+        // The engine has committed this L1 height, so advance the forward-cache floor.
+        // Individual slot and mutation histories are compacted lazily when next touched, so
+        // RPC-created slots cannot amplify engine work.
+        self.l1_state_cache
```
mattsse: **[P2] Advance the floor on follower imports too** — This is only called from `ZoneEngine::advance`, but that engine is spawned only when `sequencer_config` is present. The subscriber still runs on non-sequencer nodes and adds invalidations and slot versions while the floor remains zero, so follower memory grows for the lifetime of the process. Drive the floor from the common canonical import/commit path for every node role and initialize it from local `TempoState` on startup.

mattsse (commented):
Holistic pass after the stacked changes were consolidated into #626. I found four approval blockers and one follower-lifetime issue; inline details below. The previously discussed global cache-cardinality issue remains accepted follow-up work.

mattsse (commented):
Follow-up holistic review at `2ec6aab0`. The bounded retry, fixture seeding, receive-policy seeding, and canonical follower-floor fixes look good. Two P1 approval blockers remain; inline comments include concrete implementation suggestions.

crates/precompiles/src/lib.rs
mattsse, line 132: [P1] Apply the anchored L1 overlay to internal fee collection

This lookup only wraps external calls to `TIP_FEE_MANAGER_ADDRESS`. Tempo's transaction handler invokes `TipFeeManager::collect_fee_pre_tx` and `collect_fee_post_tx` directly against the ordinary journal-backed `StorageCtx`, so those fee-token transfers never see `ZonePrecompileStorageProvider`. An L1-blacklisted fee payer can therefore submit an otherwise empty transaction and still execute it.

Suggested implementation: factor the anchored policy-storage setup out of `create_l1_backed_precompile` and make it available to `ZoneBlockExecutor`/the Tempo fee hooks. Resolve the `TempoState` anchor once for the transaction, then run both pre- and post-fee collection through the same `ZonePrecompileStorageProvider` (preserving local balances/allowances while sourcing TIP-403 policy slots from L1). Add an end-to-end empty-call transaction test with the sender blacklisted at the anchor; it should be rejected before the body executes.

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -1,334 +1,187 @@
...
+            }
+            ITIP20::burnCall::SELECTOR | ITIP20::burnWithMemoCall::SELECTOR => {
+                self.check_burn_auth(call.caller)
             }
+            ITIP20::balanceOfCall::SELECTOR => {
+                decode_and_check::<ITIP20::balanceOfCall>(args, |decoded| {
+                    self.check_balance_read(decoded.account, call.caller)
+                })
+            }
+            ITIP20::allowanceCall::SELECTOR => {
```
mattsse, line 96: [P1] Restore privacy checks for reward getters

`userRewardInfo(address)` and `getPendingRewards(address)` still fall through to upstream dispatch without applying the account/sequencer authorization used by the other private reads. The previous wrapper protected both selectors, so this exposes per-account reward state to arbitrary callers.

This compiles against the current `ITIP20` bindings:

```suggestion
            ITIP20::userRewardInfoCall::SELECTOR => {
                decode_and_check::<ITIP20::userRewardInfoCall>(args, |decoded| {
                    self.check_balance_read(decoded.account, call.caller)
                })
            }
            ITIP20::getPendingRewardsCall::SELECTOR => {
                decode_and_check::<ITIP20::getPendingRewardsCall>(args, |decoded| {
                    self.check_balance_read(decoded.account, call.caller)
                })
            }
            ITIP20::allowanceCall::SELECTOR => {
```

Please also restore owner/sequencer/outsider coverage for both selectors; outsiders should receive `Unauthorized`.

crates/precompiles/src/lib.rs
0xrusowsky: pending upstream https://github.com/tempoxyz/tempo/pull/6879 to unlock the fix

0xrusowsky: pending upstream https://github.com/tempoxyz/tempo/pull/6879 to unlock the fix

crates/node/src/engine.rs
0xrusowsky: addressed in [`2ec6aab` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/2ec6aab0fd066248fac9d6a0048c15e574ddc502)

crates/node/tests/it/tip403_policy.rs
0xrusowsky: addressed in [`bd7cf42` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/bd7cf429a9b208b67a89b4bf8faa4a8be346f828)

crates/node/tests/it/utils.rs
0xrusowsky: addressed in [`bd7cf42` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/bd7cf429a9b208b67a89b4bf8faa4a8be346f828)

crates/evm/src/lib.rs
0xrusowsky: addressed in [`c8b557d` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/c8b557d860ba23465ae70085616c27737a3346d0)

crates/precompiles/src/ztip20/mod.rs
0xrusowsky: i did it on purpose cause rewards are disabled on TIP20s now but the selectors were preserved as a no-op.

IMO fine to leave as is (cause they won't like anything and simply return 0), and simply update the spec.

wdyt @0xKitsune @mattsse?

crates/precompiles/src/lib.rs
0xrusowsky: tentatively fixed with [`1e130d4` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/1e130d4acd7c27a2eb31ea7106dd91ac2adf4074) which assumes upstream PR gets merged as is

0xrusowsky: tentatively fixed with [`1e130d4` (this PR)](https://github.com/tempoxyz/zones/pull/626/commits/1e130d4acd7c27a2eb31ea7106dd91ac2adf4074) which assumes upstream PR gets merged as is

0xrusowsky:
cyclops audit note="we decided we will address unbounded cache growth in a follow-up PR, so don't report this issue"

PR #628, opened 2026-07-13
feat(precompiles): l1-backed storage provider

0xrusowsky:
stacked on top of https://github.com/tempoxyz/zones/pull/626

this PR adds a composable precompile storage overlay for finalized Tempo L1 state without changing active precompile dispatch.

introduces `L1StorageReader` and `ZonePrecompileStorageProvider`, which:
 - anchor all mirrored reads to the finalized Tempo block recorded in `TempoState`
 - resolve Tempo hardfork rules at that same block
 - overlay all TIP-403 registry + TIP-20 transfer policy id from L1 storage
 - reject persistent writes, increments, and decrements targeting mirrored L1 state


crates/precompiles/src/storage.rs
```diff
@@ -0,0 +1,526 @@
...
+fn is_l1_slot(address: Address, key: U256) -> bool {
+    address == TIP403_REGISTRY_ADDRESS || is_tip20_policy_id_slot(address, key)
+}
+
+fn is_tip20_policy_id_slot(address: Address, key: U256) -> bool {
+    address.is_tip20() && key == tip20_slots::TRANSFER_POLICY_ID
+}
+
+fn l1_write_err(address: Address, key: U256) -> TempoPrecompileError {
+    TempoPrecompileError::Fatal(format!(
```
mattsse: [P1] Please do not use `TempoPrecompileError::Fatal` for this user-triggerable write rejection. Once this provider is installed under the TIP-20/TIP-403 dispatch, a policy mutator can reach this path. The payload builder only skips specially tagged `[zone rpc]` fatal errors; an untagged fatal aborts payload construction rather than reverting the transaction, and the same pool transaction can be retried. Return a normal precompile revert/error here (or reject the mutating selector before dispatch), and cover the behavior through `PrecompileResult`.

crates/precompiles/src/storage.rs
```diff
@@ -0,0 +1,526 @@
...
+}
+
+// TODO(rusowsky): Remove TIP20 policy-slot detection, write protection, merge logic,
+// and related tests once Tempo L1 migrates transfer policy IDs into TIP403.
+fn is_l1_slot(address: Address, key: U256) -> bool {
+    address == TIP403_REGISTRY_ADDRESS || is_tip20_policy_id_slot(address, key)
+}
+
+fn is_tip20_policy_id_slot(address: Address, key: U256) -> bool {
+    address.is_tip20() && key == tip20_slots::TRANSFER_POLICY_ID
```
mattsse, line 271: [P2] This treats the entire physical `TRANSFER_POLICY_ID` slot as L1-owned, but Tempo's TIP-20 layout packs `nextQuoteToken` at offset 0 and `transferPolicyId` at offset 20 in the same slot. An upstream `set_next_quote_token` performs a read-modify-write of this slot and will be rejected by `sstore`, even though it only changes zone-local bits. Please make writes field-aware—preserving/validating the L1 policy bits while allowing the local packed field—and add a `setNextQuoteToken` regression test.

crates/precompiles/src/storage.rs
```diff
@@ -0,0 +1,526 @@
...
+    ))
+}
+
+pub(super) fn trace_err(
+    err: PrecompileError,
+    address: Address,
+    key: U256,
+    block_number: u64,
+) -> TempoPrecompileError {
+    TempoPrecompileError::Fatal(format!(
```
mattsse, line 288: [P2] This decoration moves an original `[zone rpc] ...` marker away from byte zero (`Tempo L1 ...: [zone rpc] ...`). `is_zone_rpc_error` uses `starts_with`, so a finite/custom `L1StorageReader` failure loses its transient classification and the payload builder aborts instead of skipping the affected transaction. Preserve the marker at the start or use structured error classification, with a regression assertion after conversion.

mattsse (changes requested):
The L1-backed overlay direction makes sense, but the new provider is not safe to wire into active precompile dispatch yet. The protected-write path turns user-triggerable operations into fatal block-building errors, TIP-20 write protection treats a packed slot as wholly L1-owned, and storage-read error decoration breaks the existing transient RPC classification. Please address these integration blockers and add dispatch-level regression coverage.

crates/precompiles/src/storage.rs
0xrusowsky: addressed in [`648c049` (this PR)](https://github.com/tempoxyz/zones/pull/628/commits/648c0492f3c8353e342a2e39605467adc89fcafc)

crates/precompiles/src/storage.rs
0xrusowsky: addressed in [`648c049` (this PR)](https://github.com/tempoxyz/zones/pull/628/commits/648c0492f3c8353e342a2e39605467adc89fcafc)

crates/precompiles/src/storage.rs
0xrusowsky: addressed in [`0adb20c` (this PR)](https://github.com/tempoxyz/zones/pull/628/commits/0adb20cc85d1890b6d7e4d72fe2e18f12a29b7dc)

0xrusowsky:
cyclops audit fast

crates/precompiles/src/storage.rs
```diff
@@ -0,0 +1,544 @@
...
+        slot: B256,
+        block_number: u64,
+    ) -> core::result::Result<B256, PrecompileError>;
+}
+
+/// Precompile storage that overlays finalized Tempo L1 policy state onto zone-local EVM state.
+///
+/// TIP-403 reads use L1 values, while TIP-20 policy reads replace only the policy-ID field.
+/// Ordinary operations remain local, and persistent writes to mirrored state are rejected.
+pub struct ZonePrecompileStorageProvider<'a, P> {
```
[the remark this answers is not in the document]
0xrusowsky: the wiring is done in follow-up PR #629 which registers via `L1BackedPrecompileEnv` + `create_l1_backed_precompile` using `ZonePrecompileStorageProvider`

PR #640, opened 2026-07-14
fix(evm): require current transaction hash

mattsse:
## Summary

- remove the `synthetic_tx_hash` fallback from `ZoneTxContext`
- revert when `current_tx_hash` is unavailable
- add regression coverage for both populated and missing transaction-hash context

## Why

`currentTxHash()` must expose the hash of the transaction actually being executed. Deriving a replacement from call and block fields when executor context is missing produces a non-canonical value and hides invalid execution setup. The precompile now fails closed instead.

## Impact

The precompile ABI and normal transaction execution are unchanged. Calls made without the executor-provided transaction-hash context now revert with empty data.

Linear: [ZONE-29](https://linear.app/tempoxyz/issue/ZONE-29/require-current-tx-hash-and-stop-using-synthetic-hash)

### DIFF
--- a/crates/evm/src/tx_context.rs
+++ b/crates/evm/src/tx_context.rs
@@ -7,8 +7,8 @@
 
 use std::{cell::RefCell, thread_local};
 
-use alloy_evm::precompiles::{DynPrecompile, PrecompileInput};
-use alloy_primitives::{B256, Bytes, keccak256};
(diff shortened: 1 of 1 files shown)


0xrusowsky (approved):
the change makes sense to me.

CI requires a foundry bump to be aware of the new `CurrentCommittee` precompile

PR #641, opened 2026-07-14
refactor(l1): follow finalized L1 head

mattsse:
## Summary

- preserve both RPC transports: WebSocket uses `newHeads` and HTTP uses block-filter polling as wake-up signals
- select and ingest blocks only through the L1 `finalized` tag
- replace the live-tip confirmation buffer and subscriber reorg recovery with one deterministic finalized-range sync step
- separate the trigger loop from the sync operation and cover finalized catch-up plus the HTTP fallback with mocked providers
- remove obsolete subscriber reorg and stream-wait metrics and align related cache/engine documentation
- patch Reth to the rebased `dev.finality-depth` change and configure the embedded Tempo L1 to finalize each block

## Why

Certified Tempo follower nodes now expose the finalized head correctly. The one-block buffer and reorg path existed for the previous follower behavior; keeping it now made the subscriber harder to reason about and test without adding reachable safety.

The subscriber creates its transport-appropriate notification stream before the initial catch-up. On startup, reconnect, and every notification it reads `finalized` and ingests the complete missing range, so delayed or coalesced notifications do not skip L1 blocks. HTTP and WebSocket URLs remain supported.

The test L1 uses Reth's standard local miner with a finality depth of one. This makes finalized-head tests prompt without a copied miner implementation or test-only consensus RPC stubs.

## Validation

- `cargo test -p zone-node --test it l1_e2e::test_zone_advances_with_real_l1 -- --nocapture`

Linear: ZONE-25

Related: #488

Depends on: https://github.com/paradigmxyz/reth/pull/26451


crates/l1/src/subscriber.rs
```diff
@@ -278,54 +204,103 @@ impl L1Subscriber {
...
+        let mut next_block = self.next_block_to_sync(l1_provider).await?;
+
+        // The caller subscribes before this initial sync so a head published
+        // while catching up remains queued as another trigger.
+        next_block = self.sync_finalized_once(l1_provider, next_block).await?;
+
+        while let Some(trigger) = triggers.next().await {
+            trigger?;
+            next_block = self.sync_finalized_once(l1_provider, next_block).await?;
         }
```
klkvr, line 277: iiuc trigger only fires when a new block is mined so in case when we get a unfinalized block and then L1 stalls but still finalizes it we won't wake up to sync it

also usually blocks are finalized faster than parent is produced i think so this always has a bit of a lag

idk if it's a big deal but i guess ideally we want to avoid delays where possible

klkvr: probably doesn't matter because we assume the L1 rpc to be following finalized blocks only

klkvr (approved):

decofe (approved):

klkvr (approved):

PR #655, opened 2026-07-15
fix: update dev genesis

0xKitsune:
This PR regenerates the bundled dev genesis with the updated proxy storage ZoneFactory bytecode and adds a regression test to keep the bundled bytecode in sync with the Foundry artifact.


0xalpharush:
why are we checking these artifacts in? seems to create a lot of churn

mattsse (approved):

legion2002:
@0xalpharush I think we should check atleast the ZonePortal in for 1091, so that we can commit to a specific etched bytecode

PR #658, opened 2026-07-15
feat(precompiles): add `ZonePrecompileError`

0xrusowsky:
this PR introduces a wrapper type `ZonePrecompileError` on top of `TempoPrecompileError` so that we can preserve tempo's devex and propagate errors naturally with `?`, rather than manually encoding precompile outputs

> [!NOTE]
> the tempo bump introduced an unpublished commonware version. It looks lik cargo does not inherit patches from git deps, so i (my clanker) had to port them over.


0xrusowsky:
cyclops audit fast

0xKitsune (approved):

PR #659, opened 2026-07-15
refactor(payload): delegate gas accounting to executor

mattsse:
## Summary

- remove the payload builder's duplicate cumulative gas counter and pre-execution fit check
- treat executor-reported block gas exhaustion as a non-fatal pool skip
- run outbox view simulations as zero-gas system envelopes, relying on Tempo's fixed internal system-call allowance

## Why

The payload builder maintained a second gas-accounting model that could diverge from the block executor's fork-aware rules, including transaction gas caps and Amsterdam regular/state gas semantics. Tempo's payment and non-payment gas lanes also do not apply to zones, which intentionally use the simplified zone executor.

## Impact

The block executor is now the single source of truth for gas capacity. Transactions that do not fit are marked invalid for the current pool iteration while other eligible transactions can still be considered. Withdrawal finalization view calls remain executable even after user transactions fill the block because their outer system envelope no longer claims the remaining block gas.

### DIFF
--- a/crates/payload/src/builder.rs
+++ b/crates/payload/src/builder.rs
@@ -255,7 +255,6 @@ where
 
         let block_gas_limit = parent_header.gas_limit();
 
-        let mut cumulative_gas_used = 0u64;
         let total_fees = U256::ZERO;
 
         let next_block_env_attributes = TempoNextBlockEnvAttributes {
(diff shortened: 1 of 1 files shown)


0xKitsune:
cyclops audit fast

0xKitsune (approved):

PR #661, opened 2026-07-15
perf: drop always-zero `pendingWithdrawalHead`

0xrusowsky:
this PR proposes getting rid of the always-zero `_pendingWithdrawalsHead`. Since partial batch finalization is rejected, every successful batch consumes and clears the entire pending withdrawal array, making the head redundant.

additionally, given that block numbers are capped at `u64::MAX` and that our rust impl can only deal with `u32::MAX` elems, i have downcasted some internal vars to pack them in a single slot.

i verified that all `ZoneOutbox.t.sol` tests continue to pass unchanged.


0xKitsune (dismissed):
Looks like dev genesis needs to be regenerated for tests to pass, otherwise lgtm.

specs/ref-impls/src/zone/ZoneOutbox.sol
```diff
@@ -77,19 +77,18 @@ contract ZoneOutbox is IZoneOutbox {
...
     /// @notice Pending withdrawals waiting to be batched
     PendingWithdrawal[] internal _pendingWithdrawals;
-    uint256 internal _pendingWithdrawalsHead;
 
     /// @notice Maximum number of withdrawal requests allowed per zone block (0 = unlimited)
     /// @dev Sequencer-configurable cap to prevent DoS via mass withdrawal requests.
     ///      This limits the number of requestWithdrawal() calls per block, complementing
     ///      the gas fee mechanism which already provides economic rate-limiting.
-    uint256 public maxWithdrawalsPerBlock;
+    uint32 internal _maxWithdrawalsPerBlock;
```
0xKitsune: Do we want to be a public function? Or is there a reason this is internal now?

specs/ref-impls/src/zone/ZoneOutbox.sol
```diff
@@ -145,11 +144,18 @@ contract ZoneOutbox is IZoneOutbox {
 
     /// @notice Set maximum withdrawal requests per zone block. Only callable by sequencer.
     /// @dev Set to 0 for unlimited. Provides rate-limiting in addition to the gas fee mechanism.
-    /// @param _maxWithdrawalsPerBlock The maximum number of requestWithdrawal() calls per block
-    function setMaxWithdrawalsPerBlock(uint256 _maxWithdrawalsPerBlock) external {
+    /// @param maxWithdrawals The maximum number of requestWithdrawal() calls per block
+    function setMaxWithdrawalsPerBlock(uint256 maxWithdrawals) external {
```
0xKitsune: If we are changing the type on `_maxWithdrawalsPerBlock`, why do we leave this as `uint256`?

specs/ref-impls/src/zone/ZoneOutbox.sol
0xrusowsky: external fns get an autogenerated getter, so i made it internal and manually implemented the getter to not break the ABI

specs/ref-impls/src/zone/ZoneOutbox.sol
0xrusowsky: same reason as mentioned above, to not break ABI

0xrusowsky: addressed in [`bad55dd` (this PR)](https://github.com/tempoxyz/zones/pull/661/commits/bad55dd59e9350c2087997db2d6f9fd2460152f0)

specs/ref-impls/src/zone/ZoneOutbox.sol
0xrusowsky: addressed in [`bad55dd` (this PR)](https://github.com/tempoxyz/zones/pull/661/commits/bad55dd59e9350c2087997db2d6f9fd2460152f0)

0xKitsune (approved):

PR #669, opened 2026-07-15
feat: add sequencer quorum settlement

legion2002:
## What changed

- require every zone to provide a sorted sequencer set and settlement threshold at creation
- initialize portals directly at sequencer-set version 1 instead of creating a privileged primary sequencer and activating quorum settlement later
- give every registered sequencer the same portal permissions; any member can submit a valid threshold-certified settlement or process withdrawals
- bind threshold signatures to the full settlement effect and remove the legacy representative address from the certificate and verifier inputs
- split sequencer fees evenly across the active set and version every admin-controlled set or threshold rotation

The legacy `sequencer()` storage slot/getter remains only for storage and ABI compatibility and grants no distinct portal authority.

## Why

There should be no primary sequencer in the multi-sequencer model. The threshold certificate authorizes settlement, while the submitting sequencer is interchangeable with every other registered member.

## Validation

- `forge fmt --check` on all changed Solidity files
- `forge build --sizes` (`ZonePortal`: 19,871-byte runtime; `ZoneFactory`: 24,089-byte runtime)
- focused Forge tests: 6 passed
- `cargo test -p tempo-zone-contracts`: 7 passed

The full local Forge suite compiles and runs 250 tests successfully, but 9 suites remain blocked in setup by the existing missing `BlockHashHistory` precompile in this local Foundry environment.

Design: https://docs.google.com/document/d/1xOo-0Ez1lz33Xkmwa9jT6uhN9xprcyRAD2M8TGykh-k/edit?tab=t.118i8icwgof9

### DIFF
--- a/specs/ref-impls/src/tempo/ZonePortal.sol
+++ b/specs/ref-impls/src/tempo/ZonePortal.sol
@@ -67,8 +67,8 @@ contract ZonePortal is IZonePortal {
     );
     bytes32 internal constant NAME_HASH = keccak256("ZonePortal");
     bytes32 internal constant VERSION_HASH = keccak256("1");
-    bytes32 internal constant ZONE_BLOCK_ATTESTATION_TYPEHASH = keccak256(
-        "ZoneBlockAttestation(uint32 zoneId,uint64 sequencerSetVersion,uint256 zoneHeight,bytes32 parentBlockHash,bytes32 zoneBlockHash)"
+    bytes32 internal constant SETTLEMENT_ATTESTATION_TYPEHASH = keccak256(
+        "SettlementAttestation(uint32 zoneId,uint64 sequencerSetVersion,uint256 zoneHeight,uint256 withdrawalBatchIndex,address sequencer,address verifier,uint64 tempoBlockNumber,uint64 anchorBlockNumber,bytes32 anchorBlockHash,bytes32 blockTransitionHash,bytes32 depositQueueTransitionHash,bytes32 withdrawalQueueHash,bytes32 verifierConfigHash)"
     );
     /*//////////////////////////////////////////////////////////////
                                 STORAGE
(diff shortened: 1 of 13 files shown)


specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -974,4 +1117,80 @@ contract ZonePortal is IZonePortal {
...
+            uint8 v = 0;
+            assembly ("memory-safe") {
+                r := mload(add(signature, 0x20))
+                s := mload(add(signature, 0x40))
+                v := byte(0, mload(add(signature, 0x60)))
+            }
+            if (v < 27) v += 27;
+            if (v != 27 && v != 28) return false;
+            if (uint256(s) == 0 || uint256(s) > SECP256K1N_DIV_2) return false;
+
```
0xalpharush: there is already a lib to do this

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -221,6 +251,56 @@ contract ZonePortal is IZonePortal {
...
+        if (newQuorum == sequencerQuorum && length == _sequencers.length) {
+            bool unchanged = true;
+            for (uint256 i = 0; i < length; ++i) {
+                if (newSequencers[i] != _sequencers[i]) {
+                    unchanged = false;
+                    break;
+                }
+            }
+            if (unchanged) revert SequencerSetUnchanged();
+        }
```
0xalpharush: idt this actually enforces changing the sequencer set as you can change the quorom and keep the old set

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/4bbafd72

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/4bbafd72

legion2002:
cyclops audit

legion2002:
Superseded by #740.

legion2002:
multi sequencers hae been merged in other PRs

PR #670, opened 2026-07-15
feat: Handle unexpected p2p shutdown for followers

adityapk00:
Handle unexpected commonware p2p shutdown by shuting down the follower node as well (so no zombie followers)

+ add test


mattsse (approved):
LGTM

PR #674, opened 2026-07-15
feat(portal): make bounceback gas configurable

0xKitsune:
This PR makes the deposit bounceback gas configurable by the sequencer. It defaults to 0, so bounceback fees are 0 until configured.


legion2002 (approved):
makes sense

PR #676, opened 2026-07-15
feat(chainspec): add `ZoneChainSpec`

0xKitsune:
This PR adds `ZoneChainSpec` and updates the Zone node to use it while preserving existing Tempo behavior.

It also bumps Tempo dependencies to the generic chain-spec changes in https://github.com/tempoxyz/tempo/pull/6862.

Depends on https://github.com/tempoxyz/tempo/pull/6862.

### DIFF
--- a/crates/evm/src/lib.rs
+++ b/crates/evm/src/lib.rs
@@ -31,6 +31,7 @@ use alloy_evm::{
     revm::{Inspector, context::DBErrorMarker, inspector::NoOpInspector},
 };
 use alloy_provider::{Provider, ProviderBuilder};
+use reth_chainspec::EthChainSpec;
 use reth_evm::{
     ConfigureEngineEvm, ConfigureEvm, EvmEnvFor, ExecutableTxIterator, ExecutionCtxFor,
     block::StateDB,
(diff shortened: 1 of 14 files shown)


crates/evm/src/lib.rs
```diff
@@ -192,9 +194,9 @@ pub struct ZoneBlockAssembler {
 
 impl ZoneBlockAssembler {
     /// Create a new [`ZoneBlockAssembler`] with the given chain spec.
-    pub fn new(chain_spec: Arc<TempoChainSpec>) -> Self {
+    pub fn new(chain_spec: Arc<ZoneChainSpec>) -> Self {
         Self {
-            inner: TempoBlockAssembler::new(chain_spec),
+            inner: TempoBlockAssembler::new(Arc::new(chain_spec.inner.clone())),
```
mattsse: this seems expensive because I believe we need to clone the entire chainspec object on every block, so we should consider wrapping the TempoChainSpec in an arc in the ZoneChainSpec as well

0xKitsune: Updated in [aef4dd4](https://github.com/tempoxyz/zones/pull/676/commits/aef4dd47bb1fb45fc6d4956e4384a4fa9b360710)

mattsse (approved):
LGTM

PR #688, opened 2026-07-16
feat: spf

rkrasiuk:
## Summary

Introduces the initial stateless proof function (SPF) for Tempo Zones. The SPF verifies Zone batch execution from witness data and produces commitments that can be used to validate the resulting Zone and withdrawal state transitions.

**Overview**

- Adds the `zone-spf` crate with the public witness, input, output, and configuration types for stateless Zone batch verification.
- Replays Zone blocks using the production Tempo/Zone EVM behavior, including Tempo checkpoint imports, deposits, token enablement, user transactions, and withdrawal finalization.
- Adds witness-backed MPT validation for Zone and Tempo state, including post-execution state-root calculation.
- Validates block continuity, Tempo checkpoint ancestry, public inputs, and withdrawal batch commitments.
- Adds the `tempo-zone-spf` CLI for discovering a Zone, collecting the required blocks and state proofs, locally validating the generated witness, and optionally writing it as JSON.


bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1365 @@
...
+                    slot = %slot,
+                    "SPF replay discovered an additional Tempo storage proof"
+                );
+                let additional_reads = BTreeMap::from([(
+                    block_number,
+                    BTreeMap::from([(account, BTreeSet::from([slot]))]),
+                )]);
+                let additional =
+                    tempo_state_witness(&tempo_provider, &initial_tempo_header, additional_reads)
+                        .await?;
```
shekhirin, line 360: what if we introduce `debug_zoneExecutionWitness` that returns an additional `tempo_reads` field?
```rust
struct ExecutionWitness {
    state: Vec<Bytes>,
    codes: Vec<Bytes>,
    headers: Vec<Bytes>,
    tempo_reads: Vec<TempoStorageRead>,
}

struct TempoStorageRead {
    block_number: u64,
    account: Address,
    slot: B256,
}
```

on the Zone RPC side, it will get populated by having a custom `L1StorageReader` implementation that records the accessed L1 states

this will allow us to remove:
1. Manual portal deposit-queue slot insertion
2. Retry loop that fetches additional witnesses
3. Block tracing

bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1365 @@
...
+                .client()
+                .request::<_, ExecutionWitness>(
+                    "debug_executionWitness",
+                    (BlockNumberOrTag::Number(number),),
+                )
+                .await
+                .wrap_err_with(|| format!("debug_executionWitness for Zone block {number}"))?;
+            if witness.headers.len() > 1 {
+                bail!(
+                    "Zone block {number} reads an older BLOCKHASH, which the current SPF witness cannot represent"
```
shekhirin, line 913: do we intend to disable `BLOCKHASH` in Zones EVM, or is it pending support in SPF? IIUC we currently can't prove an out-of-batch `BLOCKHASH` access

bin/tempo-zone-spf/src/main.rs
rkrasiuk: good idea, tho block number should be unnecessary

bin/tempo-zone-spf/src/main.rs
rkrasiuk: to be discussed

bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1365 @@
...
+        for code in witness.codes {
+            codes.entry(keccak256(&code)).or_insert(code);
+        }
+        traces.insert(number, trace);
+    }
+
+    Ok((
+        ZoneStateWitness {
+            node_pool: state.into_values().collect(),
+            bytecodes: codes.into_values().collect(),
```
shekhirin: let's store this one as a hashmap, because we then recalculate the keccaks again here https://github.com/tempoxyz/zones/blob/bb3b1516cb4d8a00707d0834dc99545b5c97b00a/crates/spf/src/execution/database.rs#L79-L82

bin/tempo-zone-spf/src/main.rs
shekhirin: BLOCKHASH opcode doesn't use system contract

crates/spf/src/execution/database.rs
```diff
@@ -0,0 +1,323 @@
...
+            node_pool,
+            missing_read: Arc::default(),
+        })
+    }
+
+    /// Return a reader rooted at a Tempo header imported by the current Zone
+    /// block. `ZoneInbox.advanceTempo` validates that the header is the next
+    /// checkpoint before any Tempo-dependent system work executes.
+    pub(crate) fn with_imported_checkpoint(
+        &self,
```
shekhirin:
```suggestion
        self,
```

can then avoid clones below

crates/spf/src/execution/database.rs
```diff
@@ -0,0 +1,323 @@
...
+        })
+    }
+
+    /// Calculate the post-state root from the initial witness and cumulative
+    /// in-memory execution changes.
+    pub(crate) fn state_root(
+        &self,
+        bundle_state: &BundleState,
+    ) -> Result<B256, StatelessSparseTrieError> {
+        let mut trie = StatelessSparseTrie::new(self.pre_state_root, &self.node_pool)?;
```
shekhirin, line 104: I think we can do `self.state.clone()` here, because it's never modified since creation. This will allow us to save on revealing the whole witness again, and to remove `self.pre_state_root` and `self.node_pool` fields.

This will require Reth to implement `Clone` for `SparseStateTrie`

crates/spf/src/execution/evm.rs
```diff
@@ -0,0 +1,427 @@
...
+    execute_recovered_transaction(
+        executor,
+        recovered,
+        Error::AdvanceTempoExecution { block_index },
+        true,
+    )?;
+    Ok(transaction)
+}
+
+fn execute_finalize_withdrawal_batch<'a, 'db, I>(
```
shekhirin, line 274: would be nice to deduplicate with the builder function https://github.com/tempoxyz/zones/blob/fdb2c5a95916026c4c8c3440f213927878852c23/crates/payload/src/builder.rs#L591-L630

crates/spf/src/execution/evm.rs
```diff
@@ -0,0 +1,427 @@
...
+        },
+        general_gas_limit: 0,
+        shared_gas_limit: gas_limit,
+        validator_set: None,
+        consensus_context: None,
+        subblock_fee_recipients: HashMap::new(),
+    }
+}
+
+fn execute_advance_tempo<'a, 'db, I>(
```
shekhirin, line 236: same https://github.com/tempoxyz/zones/blob/fdb2c5a95916026c4c8c3440f213927878852c23/crates/payload/src/builder.rs#L670-L710

crates/spf/src/execution/database.rs
```diff
@@ -0,0 +1,323 @@
...
+/// When the shared node pool includes that checkpoint's root, it owns a fully
+/// revealed immutable sparse trie. Otherwise it remains inactive and rejects
+/// any Tempo storage read as an incomplete witness.
+#[derive(Clone, Debug)]
+pub struct TempoWitnessDatabase {
+    state: Option<Arc<StatelessSparseTrie>>,
+    tempo_block_hash: B256,
+    tempo_block_number: u64,
+    node_pool: Arc<Vec<Bytes>>,
+    missing_read: Arc<Mutex<Option<MissingTempoStorageRead>>>,
```
shekhirin, line 168: I'd prefer to propagate this error properly, I think it's cleaner than checking it like that https://github.com/tempoxyz/zones/blob/fdb2c5a95916026c4c8c3440f213927878852c23/crates/spf/src/lib.rs#L158-L165

bin/tempo-zone-spf/src/main.rs
shekhirin: won't fix: main.rs is a debugging tool, and `ZoneStateWitness` is a part of verifier API that needs to have lean types https://github.com/tempoxyz/zones/blob/30d665214319e0a4425374c43e3ebc46aef675b3/specs/spec.md?plain=1#L1304

shekhirin (dismissed):
LGTM, nothing blocking. Can iterate on comments in follow-up PRs

crates/spf/src/lib.rs
```diff
@@ -0,0 +1,1097 @@
...
+        witness.parent_header.state_root(),
+    )?;
+    let mut zone_state = State::builder()
+        .with_database(zone_database)
+        .with_bundle_update()
+        .build();
+
+    // Capture the pre-batch deposit state from the parent Zone state. The
+    // transition output commits to this exact pair.
+    let previous_processed_hash = B256::from(
```
shekhirin, line 59: I think there's an opportunity to cleanup and make this less error prone by reading storage fields via `StorageCtx`

bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1374 @@
...
+                    {
+                        reads
+                            .entry(checkpoint)
+                            .or_default()
+                            .entry(call.account)
+                            .or_default()
+                            .insert(call.slot);
+                    } else if let Ok(call) =
+                        TempoStateStorageReads::readTempoStorageSlotsCall::abi_decode(&input)
+                    {
```
klkvr: we've removed those methods recently so i don't think this traces parsing is useful anymore https://github.com/tempoxyz/zones/pull/959

crates/spf/src/types.rs
```diff
@@ -0,0 +1,141 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct ZoneBlock {
+    pub number: u64,
+    pub parent_hash: B256,
+    pub timestamp: u64,
+    pub beneficiary: Address,
+    /// RLP-encoded Tempo header passed to `ZoneInbox.advanceTempo`, when the
+    /// block imports a new Tempo checkpoint.
+    pub tempo_header_rlp: Option<Bytes>,
```
klkvr: should this field be required? we kinda already assume every block has an `advanceTempo` call in a bunch of places

shekhirin: checks out https://github.com/tempoxyz/zones/blob/9e04a1d914e7bd986e190fd0dd1ceccd386585a9/crates/evm/src/executor.rs#L233-L238, making it non-optional

bin/tempo-zone-spf/src/main.rs
shekhirin: yup, removed

klkvr:
cyclops audit

bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1274 @@
...
+        parent_header,
+        zone_blocks: extracted.iter().map(|block| block.input.clone()).collect(),
+        zone_state_witness,
+        tempo_state_witness: initial_tempo_state_witness,
+        tempo_ancestry_headers,
+    };
+
+    let started = start_phase("SPF validation");
+    let output = loop {
+        match prove_zone_batch(&spf_config, witness.clone()) {
```
klkvr, line 329: worst case for this loop is quite bad anyway so should we remove the `initial_tempo_state_witness` stuff altogether and just keep the loop until we have custom witness rpc?

bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1274 @@
...
+    withdrawal_batch_index: u64,
+    tempo_block_number: u64,
+    block_hash: B256,
+}
+
+#[derive(Debug)]
+struct ExtractedBlock {
+    input: ZoneBlock,
+    block_hash: B256,
+    checkpoint_number: Option<u64>,
```
klkvr: `checkpoint_number` should also not be optional iiuc given that it's coming from `advanceTempo` call which we enforce to be present

bin/tempo-zone-spf/src/main.rs
```diff
@@ -0,0 +1,1274 @@
...
+    for block in &extracted {
+        if let Some(imported) = block.checkpoint_number {
+            checkpoint = imported;
+        }
+        checkpoint_by_zone_block.insert(block.input.number, checkpoint);
+    }
+    let mut reads = L1Reads::new();
+    for block in &extracted {
+        let checkpoint = checkpoint_by_zone_block[&block.input.number];
+        // advanceTempo always authenticates the portal deposit-queue head.
```
klkvr: we should be able to get rid of `checkpoint_by_zone_block` if we make checkpoint non-optional for `ExtractedBlock`

bin/tempo-zone-spf/src/main.rs
shekhirin: good catch

bin/tempo-zone-spf/src/main.rs
shekhirin: yup, done

crates/spf/src/execution/evm.rs
```diff
@@ -0,0 +1,405 @@
...
+        general_gas_limit: 0,
+        shared_gas_limit: block_gas_limit,
+        timestamp_millis_part: header.timestamp_millis_part,
+        consensus_context: None,
+        subblock_fee_recipients: HashMap::new(),
+    })
+}
+
+pub(crate) fn next_block_execution_context(
+    chain_spec: &zone_chainspec::ZoneChainSpec,
```
klkvr, line 210: a bit worried about those 2 fns, they are effectively copy-pasted from tempo/zones and could be hard to keep track of

any way we could unify?

klkvr (approved):

PR #689, opened 2026-07-16
fix: cache ancestry headers across batches

decofe:
## Summary
- cache validated, RLP-encoded Tempo headers across overlapping ancestry-mode batches
- fetch only missing headers, typically the newly produced suffix
- use a fixed-capacity `schnellru::LruMap` capped at 262,144 headers (~150 MiB payload)
- add a mocked-RPC regression proving the second overlapping range fetches only one new header and the cache respects its capacity

Anchor reuse is intentionally deferred to a separate PR.

## Validation
- `cargo metadata --offline --no-deps --format-version 1`
- `git diff --check`
- updated `schnellru` revision is running CI

Prompted by: @rkrasiuk

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13438,6 +13438,7 @@ dependencies = [
  "eyre",
  "futures",
  "k256",
+ "lru 0.16.4",
  "metrics",
  "parking_lot",
  "reth-metrics",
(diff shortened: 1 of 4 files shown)


0xalpharush:
maybe some changes in https://github.com/tempoxyz/zones/pull/692 worth pulling in

mattsse (approved):

PR #693, opened 2026-07-16
perf: pack withdrawal queue bounds into one slot

0xalpharush:
## Summary

- pack withdrawal queue `head` and `tail` as `uint64` values in one storage slot
- add `withdrawalQueueBounds()` so the sequencer reads both values with one `eth_call`
- update the shifted ZonePortal slot constants and regenerate the storage layout and dev genesis

## Deployment

This intentionally does not preserve the previous ZonePortal storage layout. Deploy a fresh ZoneFactory/ZonePortal before updating sequencers.

## Validation

- `forge test` — 555 passed
- `cargo test -p zone-sequencer --lib` — 45 passed
- bundled genesis artifact tests — 2 passed


crates/contracts/src/precompiles/zone_portal.rs
```diff
@@ -180,8 +180,9 @@ crate::sol! {
         function blockHash() external view returns (bytes32);
         function currentDepositQueueHash() external view returns (bytes32);
         function lastSyncedTempoBlockNumber() external view returns (uint64);
-        function withdrawalQueueHead() external view returns (uint256);
-        function withdrawalQueueTail() external view returns (uint256);
+        function withdrawalQueueHead() external view returns (uint64);
+        function withdrawalQueueTail() external view returns (uint64);
+        function withdrawalQueueBounds() external view returns (uint64 head, uint64 tail);
```
legion2002, line 185: do we need this new fn? `withdrawalQueueBounds` 
maybe we can just remove the individual head and tail fns?

legion2002 (approved):
wow big diff for a small change :)

crates/contracts/src/precompiles/zone_portal.rs
0xalpharush: it is used by `process_queue` to retrieve both without multiple RPC calls

0xalpharush: but fine to drop if you think we should

jenpaff:
superseded by https://github.com/tempoxyz/zones/pull/823

PR #696, opened 2026-07-17
feat: enforce closed-loop zone flows

0xalpharush:
## Summary

- add per-zone closed-loop membership and callback-only gateway configuration
- enforce membership for Tempo-side deposit callers, refund recipients, plain withdrawals, and bounce-backs while allowing arbitrary zone deposit recipients
- restrict callback withdrawals to registered gateways and require a synchronous return deposit
- define the shared `CallbackData` ABI with `Deposit` and `Redeem` flows, decoded by the outbox, messenger, and gateway
- keep the active messenger and gateways outside the plain-recipient membership set
- update Rust precompile bindings, creation options, storage layout, specification, and reference tests

## Why

The existing portal and callback paths did not distinguish ordinary closed-loop participants from infrastructure contracts. That allowed unrestricted callback targets and left several deposit, refund, and withdrawal paths outside one consistent membership boundary.

This change introduces explicit member and gateway roles at the Tempo custody boundary. Plaintext and encrypted zone deposit recipients remain unrestricted. Callback safety still depends on the correctness of the admin-configured `ZoneGateway`; the portal currently proves only that the callback synchronously appended a return deposit.

## Impact

Zone creation now requires initial allowed-account and gateway lists, and the `createZone` ABI changes accordingly. Callback payloads use the new `CallbackData` schema and only `Deposit` and `Redeem` enum variants. Legacy router callback flows are not compatible without migration.

## Validation

- `forge test -q` — passes
- focused factory, portal, messenger, outbox, bridge, and integration suites — pass
- `forge fmt --check` — passes
- `git diff --check` — passes
- `cargo test -p zone-node --test advance_tempo` — fails 2 of 4 tests because the bundled ZoneFactory and ZoneConfig bytecode is stale

## Known blockers / follow-ups

- enforce the closed-loop sequencer-fee invariant onchain; the implementation currently retains `FIXME(closed-loop-fees)`
- regenerate `crates/node/assets/zone-dev-genesis.json`
- deploy a compatible Moderato ZoneFactory and update the default address and deployment documentation
- wire required creation arguments through Just recipes and documented commands with a usable gateway
- remove or migrate legacy router integration flows that still register a dummy gateway and use the old callback contract


0xalpharush:
cyclops audit

specs/ref-impls/src/tempo/ZoneFactory.sol
```diff
@@ -60,10 +60,36 @@ contract ZoneFactory is IZoneFactory {
...
-        // Validate initial token is a TIP-20
+        // Validate initial token is a TIP-20.
         if (!ITIP20Factory(StdPrecompiles.TIP20_FACTORY_ADDRESS).isTIP20(params.initialToken)) {
             revert InvalidToken();
         }
+        if (params.zoneGateways.length == 0 || params.allowedAccounts.length == 0) {
+            revert InvalidClosedLoopConfig();
+        }
+
+        for (uint256 i; i < params.allowedAccounts.length; ++i) {
```
legion2002, line 71: don't think these messenger, account 0 checks, or duplication checks give us much, we can just remove both these loops?

specs/ref-impls/src/tempo/ZoneFactory.sol
```diff
@@ -60,10 +60,36 @@ contract ZoneFactory is IZoneFactory {
...
+            address account = params.allowedAccounts[i];
+            if (account == address(0) || account == _messenger) {
+                revert InvalidClosedLoopConfig();
+            }
+            for (uint256 j; j < i; ++j) {
```
legion2002: the duplication checks here are not necessary, because it would just set the same account to true multiple times.

specs/ref-impls/src/tempo/ZoneFactory.sol
```diff
@@ -60,10 +60,36 @@ contract ZoneFactory is IZoneFactory {
...
+        }
+
+        for (uint256 i; i < params.zoneGateways.length; ++i) {
+            address gateway = params.zoneGateways[i];
+            if (gateway == address(0)) revert InvalidClosedLoopConfig();
+            for (uint256 j; j < params.allowedAccounts.length; ++j) {
+                // A gateway is a withdrawal-and-call destination, never a plain recipient.
+                if (params.allowedAccounts[j] == gateway) revert InvalidClosedLoopConfig();
+            }
+            for (uint256 j; j < i; ++j) {
```
legion2002: same here

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -747,8 +804,12 @@ contract ZonePortal is IZonePortal {
 
         bool success;
         if (withdrawal.gasLimit == 0) {
+            if (zoneGateway[withdrawal.to]) revert InvalidCallbackTarget();
+            _requireAllowed(withdrawal.to);
```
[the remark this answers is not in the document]
legion2002: think this is valid? but the bounceback solution might make this slightly more complicated

specs/ref-impls/src/tempo/ZoneMessenger.sol
```diff
@@ -48,6 +51,12 @@ contract ZoneMessenger is IZoneMessenger {
         ZoneInfo memory zone = zoneFactory.zones(zoneId);
         if (zone.portal != msg.sender) revert UnauthorizedPortal();
 
+        if (!IZonePortal(msg.sender).zoneGateway(target)) revert InvalidCallbackTarget();
+
+        // Decoding CallbackData rejects any flow other than Deposit or Redeem before
+        // funds are transferred to the configured gateway.
+        abi.decode(data, (CallbackData));
```
legion2002: can you explain why this is needed?

legion2002: codex suggests that the security benefit is pretty narrow. Do we have to keep the `CallbackData` in sync just because of this?

0xKitsune (approved):
Lgtm once @legion2002's comments are addressed and CI is fixed.

0xalpharush:
replaced by https://github.com/tempoxyz/zones/pull/709

PR #697, opened 2026-07-17
feat: migrate token policies on zone enablement

legion2002:
> Requires TIP-1092 to be live in the environment; only merge after testnet redeployment.

Makes `ZonePortal` check a TIP-20's TIP-403 policy binding before enabling the token. If the binding is unset, it calls `migrateTransferPolicyIds`, rechecks, and fails closed unless TIP-403 reports `isSet == true`.

Covers both the Solidity initializer and later `enableToken` calls. Native `ZoneFactory::create_zone` bypasses the Solidity portal enablement path; proactive migration for the supported provisioning flow is handled separately by #767.

Uses the canonical `(bool isSet, uint64 policyId)` interface from tempoxyz/tempo-std#132 and complements tempoxyz/tempo#6846.

Validated by the full GitHub Specs, Rust lint, Docker build, storage-layout, and sharded test workflows on the rebased head.


crates/precompiles/src/tip403_proxy/dispatch.rs
```diff
@@ -112,6 +116,21 @@ impl<P: PolicyCheck> ZoneTip403ProxyRegistry<P> {
...
 
+    /// Handle `tokenTransferPolicyId(token) -> (bool, uint64)`.
+    fn handle_token_transfer_policy_id(&self, token: Address) -> PrecompileResult {
+        let policy_id = self.resolve_transfer_policy_id(token)?;
+        let result = ITIP403Registry::tokenTransferPolicyIdReturn {
+            isSet: true,
+            policyId: policy_id,
+        };
+        let mut storage = StorageCtx::default();
+        if storage.deduct_gas(POLICY_DATA_GAS).is_err() {
```
0xalpharush: We should charge gas before RPC call. This is pre-existing and can be done in follow-up

0xalpharush (dismissed):

crates/precompiles/src/tip403_proxy/dispatch.rs
legion2002: think this has been addressed with https://github.com/tempoxyz/zones/pull/698/changes which has been merged.

so rebasing against it and merging

adityapk00 (dismissed):
lgtm

0xalpharush (approved):

legion2002:
cyclops audit fast

legion2002:
cyclops comments addressed in - https://github.com/tempoxyz/zones/pull/800

PR #698, opened 2026-07-17
feat(l1+evm+precompiles): add L1-backed zone db

0xrusowsky:
replaces https://github.com/tempoxyz/zones/pull/626

## High-level goal

before this PR stack, Zones had two representations of Tempo L1 state:
  1. A semantic `PolicyCache` / `PolicyProvider`, populated from typed TIP-403 and TIP-20 events with RPC fallback. Custom `ZoneTip20Token` and `ZoneTip403ProxyRegistry` implementations used it to reproduce upstream policy behavior.
  2. A generic block-versioned raw storage cache keyed by `(address, slot)`, used by `TempoState` and Zone system contracts for portal storage reads.

this duplicated policy semantics and left two L1-state paths that had to remain aligned in block anchoring, event decoding, invalidation, RPC fallback, and upstream behavior. The goal is to remove the semantic policy implementation from the EVM path and execute upstream Tempo code against one factory-installed L1-aware database. The database wraps the normal caller-provided EVM database and:
  - lazily reads the Tempo L1 anchor from local `TempoState`
  - serves TIP-403 registry storage from L1 at that anchor
  - overlays only the L1-owned `transferPolicyId` field in TIP-20’s packed slot, preserving Zone-local fields
  - delegates ordinary account, code, block-hash, and storage reads to local Zone state
  - rejects TIP-403 writes and strips mirrored TIP-20 fields from canonical state transitions

from upstream Tempo's perspective these remain ordinary EVM storage operations, but selected slots are virtualized to finalized L1 state. This makes upstream Tempo code the source of execution semantics rather than maintaining Zone-specific policy implementations or installing a separate storage provider around each precompile call.

## EVM Database

`database.rs` adds `L1OverlayDB`, a database adapter installed by `ZoneEvmFactory` for every EVM created for payload building, canonical import, RPC calls, tracing, transaction-pool validation, and maintenance execution. The factory uses the alloy-evm database-context extension so the internal Tempo context can execute over `AnchoredZoneDb<DB>` while `ZoneEvm` continues to expose and return the original caller-provided `DB`.

this keeps Reth’s state ownership and commit path unchanged. Reads pass through the adapter during execution, while successful state transitions are validated and sanitized before Reth commits them through the original database. Inner database failures preserve their original error type; anchor, L1-read, and mirrored-write failures are propagated as typed custom EVM errors.

the adapter and native `TempoState` precompile share one execution-local anchor controller. It tracks whether execution is still at the parent Tempo anchor or has advanced to its direct child, rejects reads at an inconsistent anchor, rejects parent-anchor reads before advancement, and rejects duplicate or non-contiguous advancement. Noncommitting payload simulations restore the controller snapshot so they cannot leak anchor state into included transaction execution.

until TIP-20 policy storage moves fully into TIP-403, the adapter retains compatibility with the current packed layout. It reads the local packed word first, replaces only the L1-owned `transferPolicyId` field for execution, and restores the local field before canonical transitions are committed. TIP-403 account or storage mutations fail closed.

## Precompile Execution

`execution.rs` now has one execution path for Zone-native and upstream Tempo precompiles. Every wrapper receives the same EVM configuration, storage-action recorder, and non-creditable-slot set. Whether a storage read is local or L1-backed is entirely a database concern rather than a choice between local and L1-aware precompile providers.

`CallRules` contains only Zone-specific admission behavior:
- delegate calls are denied by default and explicitly enabled only for precompiles that support them
- fixed-gas selectors are checked before execution
- pure selector/caller admission returns either `Continue` or ABI-encoded revert data
- the common wrapper applies calldata gas, reservoir handling, and fixed-price output consistently
- admitted calls preserve the original calldata and caller when forwarding upstream

TIP-20, TIP-403, and `TipFeeManager` now execute the upstream Tempo implementations directly over ordinary `EvmPrecompileStorageProvider` storage. The remaining Zone-specific behavior is limited to privacy checks, bridge mint/burn authorization, read-only TIP-403 selectors, fixed gas, and delegate-call policy.

this removes `ZonePrecompileStorageProvider`, the separate local/L1-backed execution orchestration, and the custom L1-aware `ZoneFeeManager`. Tempo’s ordinary protocol fee manager now observes the same anchored policy state through the EVM database adapter.

## Follow-up work

- https://github.com/tempoxyz/zones/pull/701
- harden L1 state cache


crates/evm/src/zone_evm/mod.rs
```diff
@@ -2,47 +2,99 @@
...
+            self.anchor_controller().restore(snapshot);
+            return Ok(result);
+        }
+        if let Err(error) = self.inner.db().sanitize_state(&mut result.state) {
+            self.anchor_controller().restore(snapshot);
+            return Err(error.into_evm_error());
+        }
+        Ok(result)
+    }
+}
```
klkvr: can we not avoid this logic? we always have `advanceTempo` as first tx in the block so i think entire block's execution would use the same L1 state always?

crates/evm/src/database.rs
```diff
@@ -0,0 +1,343 @@
...
+        if let Some(account) = state.get(&TIP403_REGISTRY_ADDRESS) {
+            if account.info != account.original_info() {
+                return l1_write(TIP403_REGISTRY_ADDRESS, U256::ZERO);
+            }
+            for (slot, value) in &account.storage {
+                if value.is_changed() {
+                    return l1_write(TIP403_REGISTRY_ADDRESS, *slot);
+                }
+            }
+        }
```
klkvr, line 154: is this actually reachable given that we ban all mutating selectors? it would be kinda unfortunate if txs can execute all the way to here and only then get rejected

0xrusowsky:
cyclops audit

crates/evm/src/executor.rs
```diff
@@ -99,15 +103,33 @@ where
...
+        f: impl FnOnce(&Self::Result) -> CommitChanges,
+    ) -> Result<Option<GasOutput>, BlockExecutionError> {
+        let snapshot = self.evm().anchor_controller().phase();
+        let output = self.execute_transaction_without_commit(tx)?;
+        if !f(&output).should_commit() {
+            self.evm().anchor_controller().restore(snapshot);
+            return Ok(None);
+        }
+        Ok(Some(self.commit_transaction(output)))
+    }
```
klkvr: `execute_transaction_without_commit` should not advance any state either. it is expected that `execute_transaction_without_commit` might be called multiple times in a row with different transactions

crates/node/src/node.rs
```diff
@@ -669,6 +692,72 @@ where
         Ok(())
     }
 
+    /// Track the canonical Zone anchor and advance the raw L1 cache floor for every node role.
+    fn spawn_l1_state_cache_floor_task(&self, ctx: &AddOnsContext<'_, N>) {
```
klkvr: is this relevant for this PR?

crates/evm/src/database.rs
0xrusowsky: should be unreachable yeah, i only added it cause it feels like an invariant worth enforcing?

crates/evm/src/zone_evm/mod.rs
0xrusowsky: simplified state machine logic [`b62c49f` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/b62c49fcfdd01c23a0ae4beac7e7f845b3e2371e)

crates/evm/src/executor.rs
0xrusowsky: addressed in [`b62c49f` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/b62c49fcfdd01c23a0ae4beac7e7f845b3e2371e)

crates/node/src/node.rs
0xrusowsky: dropped cache changes from this PR and we can take in a follow-up [`73dfe07` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/73dfe078f62f498ac23480794f846fa18b1040f3)

crates/evm/src/database.rs
```diff
@@ -0,0 +1,366 @@
...
+}
+
+impl<DB: fmt::Debug, L1> fmt::Debug for AnchoredZoneDb<DB, L1> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("AnchoredZoneDb")
+            .field("inner", &self.inner)
+            .field("controller", &self.controller)
+            .finish_non_exhaustive()
+    }
+}
```
0xKitsune: nit: can we move this to after the main `AnchoredZoneDb` impl block?

crates/evm/src/database.rs
```diff
@@ -0,0 +1,366 @@
...
+        source: PrecompileError,
+    },
+    /// A transaction attempted to persist mirrored Tempo-owned state.
+    #[error("write to mirrored Tempo storage address={address} slot={slot}")]
+    L1Write { address: Address, slot: U256 },
+}
+
+impl<E: DBErrorMarker> DBErrorMarker for AnchoredZoneDbError<E> {}
+
+impl<E: DBErrorMarker> AnchoredZoneDbError<E> {
```
0xKitsune: nit: can we move this after the main `AnchoredZoneDb` logic? Same thing for `PackedPolicySlot`.

crates/evm/src/database.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// Rejects writes to the L1-mirrored slots and removes the mirrored field from packed TIP-20 transitions.
+    pub fn sanitize_state(
+        &self,
+        state: &mut AddressMap<Account>,
+    ) -> Result<(), AnchoredZoneDbError<DB::Error>> {
+        let l1_write = |address, slot| Err(AnchoredZoneDbError::L1Write { address, slot });
+
+        if let Some(account) = state.get(&TIP403_REGISTRY_ADDRESS) {
+            if account.info != account.original_info() {
+                return l1_write(TIP403_REGISTRY_ADDRESS, U256::ZERO);
```
0xKitsune: nit: its a bit nicer IMO to use `Err` directly rather than a closure. Makes it a bit easier to read.

```rust
use AnchoredZoneDbError::L1Write;

// --snip--
return Err(L1Write { address, slot });
```

crates/evm/src/database.rs
0xKitsune: Yeah seems like a reasonable defense in depth.

crates/evm/src/database.rs
```diff
@@ -0,0 +1,366 @@
...
+        let local = self
+            .inner
+            .storage(address, slot)
+            .map_err(AnchoredZoneDbError::Inner)?;
+        if address != TIP403_REGISTRY_ADDRESS && !storage::is_tip20_policy_id_slot(address, slot) {
+            return Ok(local);
+        }
+
+        let anchor = self.anchor()?;
+        let l1 = self.l1_storage(address, slot, anchor)?;
```
0xKitsune, line 208: This can return stale policy state after the first read. `L1StateProvider::get_storage` treats the latest cached value at or before `anchor` as valid, but policy events still update only `PolicyCache`, they do not write or invalidate the corresponding raw `L1StateCache` slots. 

For example, after a slot is cached at `N` and changed on L1 at `N+1`, this read at `N+1` returns the `N` value without making an RPC request. Cache warmth can then make nodes execute the same block with different policy state.

crates/evm/src/database.rs
```diff
@@ -0,0 +1,366 @@
...
+        // TODO(rusowsky): remove once TIP-1092 is implemented
+        for ((address, slot), observed) in &self.packed_policy_slots {
+            let Some(value) = state
+                .get_mut(address)
+                .and_then(|account| account.storage.get_mut(slot))
+            else {
+                continue;
+            };
+            if storage::merge_transfer_policy_id(U256::ZERO, value.present_value)
+                != storage::merge_transfer_policy_id(U256::ZERO, observed.l1)
```
0xKitsune: This rejects the mandatory `enableToken` path whenever the token already has a non-default L1 policy. 

`ZoneTokenFactory::enable_token` calls upstream `TIP20Token::initialize`, which writes transfer policy ID 1 into this packed slot. If the overlaid L1 field is anything other than 1, this comparison returns an error, causing the entire `advanceTempo` system transaction to fail and stalling block production. 


Initialization needs an exception that discards/restores the mirrored field without treating the upstream default write as an attempted L1 mutation.

crates/precompiles/src/lib.rs
```diff
@@ -40,20 +42,183 @@ pub mod dispatch {
...
+            ))
+        } else if *address == STABLECOIN_DEX_ADDRESS {
+            None
+        } else if *address == NONCE_PRECOMPILE_ADDRESS {
+            Some(NonceManager::create_precompile(&tempo_env))
+        } else if *address == ACCOUNT_KEYCHAIN_ADDRESS {
+            Some(AccountKeychain::create_precompile(&tempo_env))
+        } else {
+            None
+        }
```
0xKitsune: This replacement lookup drops the upstream `RECEIVE_POLICY_GUARD_ADDRESS` precompile. With L1 backed TIP403 execution, upstream TIP20 can enforce a receive policy, move blocked funds into the guard, and record a claim receipt internally. 

However, users cannot call the guard's public `claim` or `burnBlockedReceipt` methods because this lookup returns `None` for its address, leaving those funds permanently locked. We should either explicitly reject receive policy resolution or find a different sollution.

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -1,334 +1,174 @@
...
+    ITIP20::transferFromWithMemoCall::SELECTOR,
+    ITIP20::approveCall::SELECTOR,
+];
+
+fn decode_and_check<C: SolCall>(args: &[u8], check: impl FnOnce(C) -> CallCheck) -> CallCheck {
+    match C::abi_decode_raw_validate(args) {
+        Ok(decoded) => check(decoded),
+        Err(_) => CallCheck::Continue,
+    }
+}
```
0xKitsune, line 43: Protected selectors must not continue when strict ABI validation fails. Upstream dispatch accepts some encodings that `abi_decode_raw_validate` rejects.

For example, setting a nonzero byte in an address word's upper padding makes this helper return `Continue`, while upstream still decodes the address and successfully serves `balanceOf`.

crates/evm/src/executor.rs
```diff
@@ -99,13 +100,20 @@ where
...
 
-        // Override the validator's fee token preference to match this
-        // transaction's resolved fee token, so the handler skips FeeAMM.
-        self.override_validator_token();
+        // System txs do not pay protocol fees. Resolving their validator token here could read L1
+        // policy state at N before `advanceTempo` moves the controller to N+1, causing the anchor
+        // advancement to fail. Thus, FeeAMM override is restricted to regular fee-paying txs.
+        if !tx_env.is_system_tx {
+            self.override_validator_token();
+        }
```
0xKitsune: Lets add a `TODO` to remove this once the ZoneFeeManager lands.

crates/l1/src/subscriber.rs
```diff
@@ -762,48 +762,24 @@ pub(crate) fn apply_sequencer_events_to_cache(
...
-                    block_number,
-                    address_to_storage_value(current_sequencer),
-                );
-                cache.set(
-                    portal_address,
-                    PORTAL_PENDING_SEQUENCER_SLOT,
-                    block_number,
-                    address_to_storage_value(pending_sequencer),
-                );
+                set_cache_slot(PORTAL_SEQUENCER_SLOT, current_sequencer.into_word());
```
0xKitsune, line 804: Can we derive this slot (as well as other slots) from the `sol!` bindings rather than using a const?

crates/precompiles/src/tip403_proxy/mod.rs
```diff
@@ -1,92 +1,182 @@
...
-use tempo_contracts::precompiles::TIP403_REGISTRY_ADDRESS;
-use zone_primitives::policy::AuthRole;
+use alloy_sol_types::{SolCall, SolError};
+use tempo_contracts::precompiles::{ITIP403Registry, TIP403_REGISTRY_ADDRESS};
 
-use crate::policy::PolicyCheck;
-
-/// The precompile address — same as the L1 TIP403Registry.
+/// Canonical TIP-403 registry address, shared with Tempo L1.
 pub const ZONE_TIP403_PROXY_ADDRESS: Address = TIP403_REGISTRY_ADDRESS;
```
0xKitsune: Do we need a separate const for this or can we just use `TIP403_REGISTRY_ADDRESS` directly?

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -1,334 +1,174 @@
...
 
-use crate::{
-    policy::PolicyCheck,
-    tip403_proxy::{AUTH_CHECK_GAS, ZoneTip403ProxyRegistry},
-};
+use crate::execution::{CallCheck, CallRules};
 
-const FIXED_TRANSFER_GAS: u64 = 100_000;
+/// Fixed gas charged for TIP20 transfer and approval selectors on the zone.
+pub const TIP20_FIXED_TRANSFER_GAS: u64 = 100_000;
```
0xKitsune, line 28: Why do we need this rather than using upstream pricing?

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -1,334 +1,174 @@
...
+}
 
-        match registry.is_authorized(policy_id, to, AuthRole::MintRecipient) {
-            Ok(true) => None,
-            Ok(false) => {
-                trace!(target: "zone::precompile", %to, policy_id, "mint recipient not authorized");
-                Some(Ok(Self::policy_forbids_output()))
-            }
-            Err(e) => Some(Err(e)),
+impl TIP20Rules {
```
0xKitsune, line 108: All of these functions are essentially the same thing. We could write one `check_auth` function and pass in the authorized address. For example:


```rust

fn check_auth(&self, caller: Address, authorized: Address) -> CallCheck {
        if caller == authorized {
            CallCheck::Continue
        } else {
            unauthorized()
        }
}

// --snip--
            ITIP20::mintCall::SELECTOR | ITIP20::mintWithMemoCall::SELECTOR => {
                self.check_auth(caller, ZONE_INBOX_ADDRESS)
            }
            ITIP20::burnCall::SELECTOR | ITIP20::burnWithMemoCall::SELECTOR => {
                self.check_auth(caller, ZONE_OUTBOX_ADDRESS)
            }
// --snip--
```

crates/evm/src/zone_evm/mod.rs
```diff
@@ -2,47 +2,101 @@
...
+{
+    fn execute_inner(
+        &mut self,
+        execute: impl FnOnce(
+            &mut TempoEvm<AnchoredZoneDb<DB, L1>, I>,
+        ) -> Result<TempoResult, AdaptedEvmError<DB::Error>>,
+    ) -> Result<TempoResult, ZoneEvmError<DB::Error>> {
+        let result = match execute(&mut self.inner) {
+            Ok(mut result) => {
+                if result.result.is_success()
```
0xKitsune: Reverts and halts still return state that the block executor commits, including the nonce and Tempo fee transition, so sanitization cannot be success only. B

Before execution, `collect_fee_pre_tx` authorizes the fee token transfer through TIP403 and mutates that same token's balances/reward state. Those fee effects remain when the user call later reverts.

The returned touched token account contains the L1 overlaid policy bits, but this branch skips `sanitize_state`, and `commit_transaction` persists them into canonical Zone storage. 

Should we sanitize every `Ok(result)` regardless of `ExecutionResult`? The reverted user-frame writes have already been rolled back, while the persistent fee/nonce transition still needs its mirrored fields removed.

crates/evm/src/database.rs
```diff
@@ -0,0 +1,366 @@
...
+        let l1_write = |address, slot| Err(AnchoredZoneDbError::L1Write { address, slot });
+
+        if let Some(account) = state.get(&TIP403_REGISTRY_ADDRESS) {
+            if account.info != account.original_info() {
+                return l1_write(TIP403_REGISTRY_ADDRESS, U256::ZERO);
+            }
+            for (slot, value) in &account.storage {
+                if value.is_changed() {
+                    return l1_write(TIP403_REGISTRY_ADDRESS, *slot);
+                }
```
0xKitsune, line 148: Checking only `is_changed()` validates that the call did not modify the L1 value, but it does not remove that value from the returned transition. The adapter supplied the L1 value as both original_value and present_value, and revm commits every storage entry of a touched account.

For example, with canonical Zone slot 0 equal to 5 and the mirrored L1 value equal to 99, a `policyIdCounter()` transaction succeeds, `sanitize_state` returns Ok, and the touched registry account in result.state contains 99. Committing that result overwrites the Zone slot even though the transaction was read only.

cc @klkvr to double check if Im overlooking something here.

crates/evm/src/test_utils.rs
```diff
@@ -0,0 +1,3 @@
+//! Shared test utilities for Zone EVM tests.
+
+pub(crate) use zone_precompiles::test_utils::MockL1Reader as TestL1;
```
0xKitsune: nit: do we need this?

crates/precompiles/src/lib.rs
```diff
@@ -65,10 +230,11 @@ pub fn zone_rpc_error(msg: impl core::fmt::Display) -> PrecompileError {
     PrecompileError::Fatal(alloc::format!("{ZONE_RPC_ERROR_PREFIX} {msg}"))
 }
 
-/// Returns `true` if the error string was produced by [`zone_rpc_error`].
+/// Returns `true` if the error chain contains a failure produced by [`zone_rpc_error`].
 pub fn is_zone_rpc_error(err: &str) -> bool {
-    err.starts_with(ZONE_RPC_ERROR_PREFIX)
+    err.contains(ZONE_RPC_ERROR_PREFIX)
 }
```
0xKitsune: What is the motivation for this? Should we have a proper error rather than `[zone rpc]` prefix?

crates/precompiles/src/execution.rs
```diff
@@ -0,0 +1,348 @@
...
+pub(crate) struct DirectCallOnly;
+impl CallRules for DirectCallOnly {}
+
+/// Rules for precompiles that allow delegate calls without additional admission checks.
+pub(crate) struct NoCallRules;
+impl CallRules for NoCallRules {
+    fn is_delegate_call_allowed(&self) -> bool {
+        true
+    }
+}
```
0xKitsune: Can you remind me why/when we need to delegate call with precompiles? We explictly restrict this upstream in Tempo via the `tempo_precompile` macro.

0xrusowsky: i saw that the L2 crypto precompiles (which are "pure") allowed delegate calls, so i thought u wanted to preserve behavior.

IMO it would be nicer to never allow it, as it would simplify impl

crates/evm/src/test_utils.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46)

crates/precompiles/src/ztip20/mod.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46)

crates/precompiles/src/tip403_proxy/mod.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46)

crates/evm/src/executor.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46)

crates/evm/src/database.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46)

crates/evm/src/database.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46) + [`71c5c6b` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/71c5c6be104133e2a415fd7d49ca277862e26011)

crates/evm/src/database.rs
0xrusowsky: addressed in [`77c41f4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/77c41f4edd651b11d978f6fb090026b962791e46)

crates/precompiles/src/ztip20/mod.rs
0xrusowsky: this is a spec constraint (which i preserved from the original ZTIP20 impls):
> Fixed gas for transfers. All TIP-20 transfer and approve operations charge a fixed 100,000 gas regardless of storage layout. This eliminates a side channel where variable gas costs reveal whether a recipient has previously received tokens.

0xrusowsky: documented explicitly [`cdff0eb` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/cdff0eb61e803462bef13e299a68b9529afec161)

crates/precompiles/src/execution.rs
0xrusowsky: addressed in [`0abf462` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/0abf46255322a4efc9a510cc87d7f5c6025199d5)

crates/precompiles/src/storage.rs
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+    }
+}
+
+/// The execution-local Tempo anchor used by the database adapter.
+#[derive(Clone, Default)]
+pub struct L1AnchorController {
+    state: Rc<Cell<L1AnchorPhase>>,
+}
```
klkvr: is there any way to avoid doing this or simplify? iiuc this is the main trade off vs `ZoneStorageProvider` which can read `tempoBlockNumber` from the state directly on every L1 read

in practice during execution zone's EVM only observes L1 state at a single anchor because `advanceTempo` is called at the beginning of the block so there must be a way to enforce this somehow

crates/evm/src/database.rs
```diff
@@ -0,0 +1,368 @@
...
+use zone_primitives::constants::TEMPO_STATE_ADDRESS;
+
+/// Resolves mirrored L1 reads at the active Tempo anchor and forwards all other database
+/// operations to the caller-provided Zone database.
+pub struct AnchoredZoneDb<DB, L1> {
+    inner: DB,
+    l1: L1,
+    controller: L1AnchorController,
+    // TODO(rusowsky): remove once TIP-1092 is implemented
+    packed_policy_slots: HashMap<(Address, U256), PackedPolicySlot>,
```
klkvr: i guess technically we don't need this cache because we could just call `is_tip20_policy_id_slot` in `sanitize_state`?

crates/evm/src/executor.rs
```diff
@@ -99,13 +101,20 @@ where
...
 
         let _tx_hash_guard = tx_context::set_current_tx_hash(*recovered.tx().tx_hash());
-        self.inner
-            .execute_transaction_without_commit((tx_env, recovered))
+        let result = self
+            .inner
+            .execute_transaction_without_commit((tx_env, recovered));
+
+        self.evm_mut().reset_transaction_state();
```
klkvr: do we need to call it again here given that `ZoneEvm::transact_raw` handles it already?

crates/evm/src/executor.rs
```diff
@@ -99,13 +101,20 @@ where
...
 
-        // Override the validator's fee token preference to match this
-        // transaction's resolved fee token, so the handler skips FeeAMM.
-        self.override_validator_token();
+        // System txs do not pay protocol fees. Resolving their validator token here could read L1
+        // policy state at N before `advanceTempo` moves the controller to N+1, causing the anchor
+        // advancement to fail. Thus, FeeAMM override is restricted to regular fee-paying txs.
+        if !tx_env.is_system_tx {
+            self.override_validator_token();
+        }
```
klkvr: why would validator token resolution here read L1 state?

crates/evm/src/lib.rs
```diff
@@ -76,85 +67,27 @@ where
...
-        // one, it still applies privacy, fixed-gas, and bridge-auth rules.
-        //
-        // This replaces the upstream `extend_tempo_precompiles` lookup, so we
-        // must also handle the non-TIP-20 Tempo precompiles that are zone-relevant
-        // (FeeManager, NonceManager, AccountKeychain).
-        // Zone-specific overrides (TIP20Factory, TIP403Proxy) are in the
-        // static map via `apply_precompile` and take priority over this.
-        let zone_cfg = cfg.clone();
-        let zone_env = PrecompileEnv::new(
+        extend_zone_precompiles(
```
klkvr: nitpick but i'd be in favor of keeping this as is without refactoring into `extend_zone_precompiles` just to keep diff smaller

crates/evm/src/database.rs
0xrusowsky: nice catch, was too aggressive when reverting L1 state cache changes to minimize the diff.

Restored a minimal version which addresses the soundness with [`e21f5a5` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/e21f5a58ae81930c6de14ba94ad3e51800f43196)

crates/precompiles/src/lib.rs
0xrusowsky: addressed in [`6fa5a19` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/6fa5a19e59aec5f94b661881b5660122bcf68fd7)

crates/evm/src/database.rs
0xrusowsky: totally missed it, nice catch. Rather than weakening the DB adapter with an initialization exception, enable_token now snapshots the anchored policy field, runs upstream initialization, and restores that field in the journal while preserving the adjacent initialized fields [`18a7ded` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/18a7ded1541fc8806156188bd2832ddaca2b7d6f)

crates/evm/src/database.rs
0xrusowsky: addressed in [`dcb4dea` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/dcb4dea491791c71275757f9178f251b81c3e2bf)

crates/l1/src/subscriber.rs
0xrusowsky: will do it later in a followup, since it was already the case

crates/precompiles/src/ztip20/mod.rs
0xrusowsky: mmm i guess it makes sense. Is there a reason why we were using `abi_decade_raw_validate` here rather than the upstream `abi_decade_raw`?

if not the easiest fix seems to simply change the fn and continue delegating errors to upstream impl

0xrusowsky: addressed in [`4f647c4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/4f647c4d56e85c82f43379dea82f878707121909)

crates/evm/src/database.rs
0xrusowsky: addressed in [`5c7c61b` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/5c7c61b5f7d49e47943448c321871409d29b3c9e)

crates/precompiles/src/storage.rs
0xrusowsky: simplified the state machine to a an `Rc<Cell<Option<u64>>>` and absorbed `L1AnchorController` into a `L1State<P>` type [`2ebaf89` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/2ebaf896e9dfef00fc5da92ccb34c2b97c687f4c)

crates/evm/src/executor.rs
0xrusowsky: was just being extra careful, but i think u are right. On the flip side, i think it doesn't hurt?

crates/evm/src/zone_evm/mod.rs
0xrusowsky: addressed in [`97b3d25` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/97b3d254e78a846067dcceb2f85d8dba3d2b2c75)

crates/evm/src/lib.rs
0xrusowsky: reverted in [`171b637` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/171b63708f0ff82c23dcbe61b227a1a536b91ebf) and will do as a follow-up

crates/evm/src/executor.rs
0xrusowsky: yup, restored [`12fb0bd` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/12fb0bda36f9597b9d733db42afb7057ff155166).

we will enable fee amm soonish, so this will eventually go either way

crates/precompiles/src/lib.rs
0xrusowsky: i missed this too. Decided to enable in the L2 for now [`f6a1d93` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/f6a1d93c0eb87f2e60c4d907e1177d744881af43).

Pending conversation on how we address the UX challenges.

crates/l1/src/state/cache.rs
```diff
@@ -55,15 +55,24 @@ impl L1StateCache {
...
 #[derive(Debug, Default)]
 pub struct L1StateCacheInner {
     tracked_contracts: HashSet<Address>,
     /// Per-slot value history: `(address, slot) → { block_number → value }`.
     /// The `BTreeMap` enables efficient range lookups for "latest value at or before block N".
     slots: HashMap<(Address, B256), BTreeMap<u64, B256>>,
-    /// Latest L1 block the cache has received data for, used for reorg detection.
+    /// Per-address mutation barriers. A value at V cannot serve a read at N when a barrier exists
+    /// in `(V, N]`.
+    invalidations: HashMap<Address, BTreeSet<u64>>,
```
klkvr, line 72: can we just clear the invalidated addresses from `slots` mapping instead of tracking an extra map?

klkvr (approved):
lgtm, pending @0xKitsune

klkvr (dismissed):

crates/evm/src/zone_evm/mod.rs
```diff
@@ -2,47 +2,99 @@
...
+
+impl<DB, I, L1> ZoneEvm<DB, I, L1>
+where
+    DB: Database,
+    L1: L1StorageReader,
+    I: Inspector<TempoCtx<AnchoredZoneDb<DB, L1>>>,
+{
+    fn execute_inner(
+        &mut self,
+        execute: impl FnOnce(
```
0xKitsune: nit: can we add doc comments on this function?

crates/evm/src/zone_evm/mod.rs
```diff
@@ -2,47 +2,99 @@
...
     }
 
     /// Provides a mutable reference to the EVM context.
-    pub fn ctx_mut(&mut self) -> &mut TempoCtx<DB> {
+    pub fn ctx_mut(&mut self) -> &mut TempoCtx<AnchoredZoneDb<DB, L1>> {
         self.inner.ctx_mut()
     }
+
+    /// Clears database-adapter bookkeeping left by the current transaction attempt.
+    pub(crate) fn reset_transaction_state(&mut self) {
```
0xKitsune: nit: consider renaming to `clear_l1_anchor` or similar for clarity as this is only clearing the db adapter state.

crates/evm/src/zone_evm/mod.rs
```diff
@@ -2,47 +2,99 @@
...
+    }
+}
+
+impl<DB, I, L1> ZoneEvm<DB, I, L1>
+where
+    DB: Database,
+    L1: L1StorageReader,
+    I: Inspector<TempoCtx<AnchoredZoneDb<DB, L1>>>,
+{
+    fn execute_inner(
```
0xKitsune: nit: consider renaming to `execute_and_sanitize` or similar for clarity. Right now this is a bit generic.

crates/l1/src/state/cache.rs
0xKitsune: @0xrusowsky thoughts?

crates/l1/src/state/provider.rs
```diff
@@ -200,6 +207,14 @@ impl L1StateProvider {
                     return Ok(value);
                 }
                 Err(rpc_err) => {
+                    if self
+                        .max_sync_attempts
+                        .is_some_and(|max_attempts| attempt >= max_attempts)
+                    {
+                        return Err(eyre::eyre!(
```
0xalpharush, line 215: should add metric for grafana?

crates/node/src/node.rs
```diff
@@ -295,6 +295,21 @@ impl ZoneNode {
...
+    /// Bound L1 state-provider retries for callers that must fail finitely on cache misses.
+    pub fn with_l1_state_provider_retry_limits(
+        mut self,
+        transport_retries: u32,
+        sync_attempts: u32,
+    ) -> Self {
+        assert!(
+            sync_attempts > 0,
+            "at least one synchronous attempt is required"
+        );
```
0xalpharush: Nonzero?

crates/node/src/node.rs
```diff
@@ -1015,20 +1030,8 @@ where
...
-                &self.l1_state_provider_config.l1_rpc_url,
-                rpc_connection_config(self.l1_state_provider_config.retry_connection_interval),
-            )
-            .await?
-            .erased();
-
-        let policy_provider = PolicyProvider::new(self.policy_cache, policy_l1, runtime_handle);
-        evm_config = evm_config.with_policy_provider(policy_provider);
-        info!(target: "reth::cli", "Zone EVM initialized with TempoState + TIP-403 proxy precompiles");
+        let evm_config = ZoneEvmConfig::new(ctx.chain_spec(), tempo_chain_spec, l1_provider);
```
0xKitsune, line 1020: Block execution now uses the anchored L1 overlay, but pool construction still replaces this with a plain `TempoEvmConfig` at line 1078. When anchored L1 policy differs from Zone-local state, `eth_sendRawTransaction` can reject a transaction with `PolicyForbids` even though block execution would accept it.

0xKitsune (dismissed):

crates/l1/src/state/provider.rs
0xrusowsky: we already emit structured warn logs for every failed attempt and an info log with the final attempt count when the read succeeds.

since finite exhaustion is only enabled in tests (prod retries indefinitely), i don’t think a separate metric is needed.

crates/node/src/node.rs
0xrusowsky: addressed in [`848589e` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/848589ee2a93b4c9178b3f0e17816237a0380e40)

crates/evm/src/zone_evm/mod.rs
0xrusowsky: addressed in [`17a3ca4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/17a3ca4a57205c67c4c3a2aa5d5a4b9016b6e626)

crates/evm/src/zone_evm/mod.rs
0xrusowsky: addressed in [`17a3ca4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/17a3ca4a57205c67c4c3a2aa5d5a4b9016b6e626)

crates/evm/src/zone_evm/mod.rs
0xrusowsky: addressed in [`17a3ca4` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/17a3ca4a57205c67c4c3a2aa5d5a4b9016b6e626)

crates/l1/src/state/cache.rs
0xrusowsky: i think this is not secure to do it, as if anchor is at `N` and tempo is at `N + i` (where i > 1) we could listen to a new event which invalidates at `N + i`. Thus, removing directly would lose the info between `N` and `N + i`. Then we would do the RPC fallback at `N + 1` and be left with a stale value.

crates/payload/src/builder.rs
```diff
@@ -512,6 +518,20 @@ where
     Ok(PoolExecutionOutcome::Complete)
 }
 
+fn is_l1_storage_unavailable(error: &(dyn Error + 'static)) -> bool {
```
legion2002, line 524: 🚨 **[SECURITY — High] F1: Dead transient-error skip — precompile path stringifies `L1StateError` before it reaches this downcast**

`is_l1_storage_unavailable` walks `Error::source()` looking for a typed `L1StateError::StorageUnavailable`. That works only on the direct DB-adapter error path (`EVMError::CustomAny`). For every TIP-20/TIP-403 read that occurs **inside a native precompile** (the dominant path), the error chain is: `L1StateError` → `EvmInternalsError::Database(ZoneDbError)` → `TempoPrecompileError::Fatal(e.to_string())` → `PrecompileError::Fatal(String)` → `EVMError::Custom(String)`. `EVMError::Custom(_)` has `source() == None` in revm, so this function returns `false`, and the caller (line 514) falls through to `Err(err) => return Err(PayloadBuilderError::evm(err))`, **aborting the whole block build** instead of skipping just the offending pool tx.

The skip mechanism is effectively dead for all policy-reading pool transactions.

**Fix:** Either match the stringified form (`error.to_string().contains("Tempo L1 storage unavailable")`), which survives stringification as the old `is_zone_rpc_error` did, or route precompile fatals through `EVMError::CustomAny` to preserve the typed source chain.

crates/l1/src/subscriber.rs
```diff
@@ -571,20 +583,25 @@ impl L1Subscriber {
                         }
                     }
                 } else if addr == TIP403_REGISTRY_ADDRESS {
+                    invalidated.insert(addr);
```
legion2002, line 587: 🚨 **[SECURITY — High] F2: `ZonePortal` mutations are not added to `invalidated` — stale cache values can be inherited across portal events**

This hunk adds `invalidated.insert(addr)` for `TIP403_REGISTRY_ADDRESS` (here, line 586) and for TIP-20 `TransferPolicyUpdate` logs (line 593). However, the `addr == portal_address` branch above in `extract_events` (which handles `TokenEnabled`, sequencer-transfer events, etc.) does **not** insert `portal_address` into `invalidated`. `update_l1_state_anchor` (line 723) then publishes receipt coverage through this block without a portal mutation barrier.

`L1StateCacheInner::get` inherits cached values when `block_number <= self.anchor.number && !invalidated`. A node that cached a portal slot (e.g. `_tokenConfigs[token] = false`) before an `enableToken()` call will continue serving that stale `false` value to `ZoneConfig.isEnabledToken()` and `ZoneOutbox.requestWithdrawal()` — permanently blocking withdrawals for the newly enabled token until a restart. Nodes with different cache histories diverge.

**Fix:** Add `invalidated.insert(portal_address)` inside the `addr == portal_address` branch of `extract_events` whenever any portal log is decoded (or, minimally, for storage-mutating portal events like `TokenEnabled`).

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -1,334 +1,154 @@
...
+                decode_and_check::<ITIP20::allowanceCall>(args, |call| {
+                    self.check_auth_with_sequencer(caller, &[call.owner, call.spender])
+                })
+            }
+            IRolesAuth::hasRoleCall::SELECTOR => {
+                decode_and_check::<IRolesAuth::hasRoleCall>(args, |call| {
+                    self.check_auth_with_sequencer(caller, &[call.account])
+                })
+            }
+            _ => CallCheck::Continue,
```
legion2002, line 103: 🚨 **[SECURITY — Medium] F3: `userRewardInfo` and `getPendingRewards` fall through ungated — zone privacy regression**

This catch-all (`_ => CallCheck::Continue`) passes every selector not explicitly matched to upstream Tempo without a caller check. Upstream Tempo exposes `userRewardInfo(address)` (returns `rewardRecipient`, `rewardPerToken`, `rewardBalance`) and `getPendingRewards(address)` (returns claimable amount, and on pre-T8 reads the holder balance) as ordinary view calls with no caller authorization.

The **previous** zone dispatch (`ztip20/dispatch.rs`, deleted by this PR) gated both selectors via `enforce_balance_of` with owner-or-sequencer checks. This PR's refactor dropped those two cases and their tests. Any authenticated zone user can now call `userRewardInfo(victim)` or `getPendingRewards(victim)` and receive another user's financial state.

**Fix:** Add `ITIP20::userRewardInfoCall::SELECTOR` and `ITIP20::getPendingRewardsCall::SELECTOR` to the `match` with `check_auth_with_sequencer(caller, &[call.account])` (same pattern as `balanceOf` above). Restore the deleted regression tests.

crates/precompiles/src/execution.rs
```diff
@@ -0,0 +1,335 @@
...
+        let mut result = StorageCtx::enter(&mut storage, || match rules.admit(data, caller) {
+            CallCheck::Continue => execute(data, caller),
+            CallCheck::Revert(output) => {
+                let s = StorageCtx::default();
+                let output = s.revert_output(output);
+                add_input_cost(s, data, Ok(output))
+            }
+        });
+        if let (Ok(output), Some(gas)) = (&mut result, fixed_gas) {
+            output.gas_used = gas;
```
legion2002, line 130: 🚨 **[SECURITY — Medium] F5: Fixed-gas wrapper sets `gas_used` but not `gas_refunded` — storage-dependent costs leak through refunds**

`output.gas_used = gas` correctly reports the fixed 100,000 charge, but `output.gas_refunded` is left as whatever upstream Tempo returned. Upstream Tempo fills `gas_refunded` from EIP-3529 slot-clearing refunds (`dispatch.rs:145-148`), and revm records that refund independently when processing precompile output. A `transfer` that clears the sender's balance slot returns `gas_used=100000, gas_refunded=19900` — making the net gas cost storage-layout-dependent despite the fixed `gas_used`.

Observers (block explorers, gas estimators, cross-tx timing) can distinguish clearing from non-clearing operations; callers of clearing paths also pay less than the intended fixed charge.

**Fix:**
```rust
if let (Ok(output), Some(gas)) = (&mut result, fixed_gas) {
    output.gas_used = gas;
    output.gas_refunded = 0;  // normalize: privacy depends on fixed net cost
}
```
If Amsterdam/EIP-8037 state-gas is enabled, also normalize `state_gas_used` and `reservoir`.

crates/l1/src/state/cache.rs
```diff
@@ -78,23 +87,61 @@ impl L1StateCacheInner {
...
+            .range(..=block_number)
+            .next_back()?;
+        let invalidated = self
+            .invalidations
+            .get(&address)
+            .and_then(|blocks| blocks.range(..=block_number).next_back())
+            .is_some_and(|&invalidated_at| invalidated_at > cached_block);
+
+        let is_exact = cached_block == block_number;
+        let can_inherit = block_number <= self.anchor.number && !invalidated;
```
legion2002, line 106: ⚠️ **[ISSUE — Medium] F7 (context): Cache inheritance requires receipt-coverage anchor — followers never advance it**

`can_inherit = block_number <= self.anchor.number && !invalidated` (this line) is the gate that allows historical values to be reused without an RPC round-trip. This anchor is only advanced by `L1Subscriber::update_l1_state_anchor`. Follower nodes skip the L1 deposit subscriber and have no path that advances `self.anchor`. Therefore on followers, a value cached at exactly block N can be served for N, but **cannot** be inherited for N+1 — every new Tempo anchor causes a fresh exact-block L1 RPC call on the critical serial import path (`engine.new_payload`). A transient L1 failure traps the follower in the unbounded retry loop; a missed broadcast is irrecoverable (no backfill).

**Fix:** Split the subscriber into (a) deposit enqueueing (leader-only) and (b) receipt coverage advancement + mutation barriers (both roles). Run (b) on followers so they establish and advance `self.anchor` matching the leader.

crates/l1/src/state/provider.rs
```diff
@@ -52,6 +55,7 @@ impl Default for L1StateProviderConfig {
             max_retries: 10,
             initial_backoff_ms: 20,
             retry_connection_interval: std::time::Duration::from_millis(100),
+            max_sync_attempts: None,
```
legion2002, line 58: 🚨 **[SECURITY — High] F6: Default `max_sync_attempts: None` combined with historical `eth_call` creates an unbounded L1 RPC retry loop**

With `max_sync_attempts: None` (this line), `get_storage` retries indefinitely when the upstream L1 node cannot serve a request. PR #698 installs `L1OverlayDB` in **every** EVM context — including the historical-state databases Reth constructs for `eth_call`. The private RPC accepts a caller-supplied `BlockId`. An authenticated user calling an open TIP-403 view at an old zone block triggers an exact-block L1 lookup at that historical anchor.

After startup, `initialize_floor(tempo_block_number)` discards values below the floor, so historical anchors can **never** become cache hits (per `L1StateCacheInner::set`). With `max_sync_attempts: None`, a single historical `eth_call` to a pruned L1 state permanently occupies a Tokio task and an upstream connection. The private Axum server has no request timeout or concurrency limit, so repeated calls can exhaust runtime capacity and threaten block production.

**Fix:** In simulation/RPC contexts, reject or clamp anchor lookups to within verified raw-cache coverage. Apply per-request deadlines and server-side concurrency limits to the private RPC server. Separate the indefinitely-retrying consensus path from the bounded simulation path.

crates/node/src/node.rs
```diff
@@ -464,7 +474,11 @@ where
         let sp = ctx.node.provider().latest()?;
         let tempo_block_number = sp.tempo_block_number()?;
         self.policy_cache.set_last_l1_block(tempo_block_number);
-        info!(target: "reth::cli", tempo_block_number, "Read local tempoBlockNumber for L1 subscriber");
+        self.l1_config
+            .l1_state_cache
+            .write()
+            .initialize_floor(tempo_block_number);
```
legion2002, line 480: ⚠️ **[ISSUE — Medium] F7: Floor initialized here for leaders; followers have no path to advance the raw-cache anchor**

`initialize_floor(tempo_block_number)` seeds the cache floor so historical reads below the current Zone state are not admitted. However, the code that **advances** the coverage anchor (`L1Subscriber::update_l1_state_anchor`) is only invoked from the L1 deposit subscriber, which follower nodes skip. Follower raw-cache anchor stays at zero throughout operation, so cache inheritance (`block_number <= self.anchor.number`) never applies: every imported anchor triggers a synchronous L1 RPC read in the critical `engine.new_payload` path.

See also `crates/l1/src/state/cache.rs:106` for the inheritance predicate and `crates/l1/src/subscriber.rs:723` for the anchor-advancement call site that followers miss.

crates/evm/src/database.rs
```diff
@@ -0,0 +1,408 @@
...
+        anchor: u64,
+    ) -> Result<U256, ZoneDbError<DB::Error>> {
+        self.l1
+            .read_l1_storage(address, B256::from(slot), anchor)
+            .map(Into::into)
+            .map_err(ZoneDbError::L1State)
+    }
+
+    /// Rejects writes to the L1-mirrored slots and removes the mirrored field from packed TIP-20 transitions.
+    pub fn sanitize_state(
```
legion2002, line 131: 🛡️ **[DEFENSE-IN-DEPTH — Low] F9: `sanitize_state` write rejection becomes a block-fatal error, not a per-tx revert**

`ZoneDbError::L1Write` (and anchor error variants) returned from `sanitize_state` are mapped to `EVMError::CustomAny` (fatal). The builder's pool-tx loop at `crates/payload/src/builder.rs:514` treats any non-`StorageUnavailable` fatal error as an abort — the entire block build fails instead of reverting only the offending transaction.

Currently non-exploitable (native value disabled chain-wide; all mutating registry selectors blocked; `changeTransferPolicyId` requires `DEFAULT_ADMIN_ROLE` held only by `ZONE_INBOX`). **However:** `TIP403_MUTATING_SELECTORS` is a hand-maintained list with no compile-time cross-check against the upstream registry ABI. An upstream hardfork adding a new mutating selector, or any future permission change, would make this a **permissionless block-production halt**.

**Fix:** Convert L1-slot write rejections to per-transaction reverts rather than fatal `CustomAny`, or broaden the builder skip-classifier to treat `ZoneDbError::L1Write` and anchor conflicts as "drop tx, continue building". Add a compile-time assertion that `TIP403_MUTATING_SELECTORS` matches the upstream registry ABI.

legion2002 (commented):
:eye: **Cyclops Review** — PR #698 (`rus/l1-aware-db`)

> **Head drift:** Workers audited `133e2bf`. Current head is `ae31693`. Finding F8 (pool EVM divergence via plain `TempoEvmConfig`) appears fixed in the newer commit — the pool now receives `ZoneEvmConfig`. All other findings below remain applicable to the current head. **Confirm** that the upstream `TempoTransactionPool<Provider, ZoneEvmConfig>` generic correctly threads `ZoneEvmFactory` through to fee-manager policy reads, and review the updated integration test comment ("policy-invalid transactions are allowed to remain pending") to verify it is not masking a remaining gap.

## Summary

The L1-overlay architecture is well-constructed and its core anchor/sanitize invariants hold. Six security issues and one latent robustness issue were confirmed across three independent audit workers:

| # | Severity | Title |
|---|----------|-------|
| F1 | 🚨 High | Transient L1 RPC failure aborts block production (skip mechanism dead for precompile path) |
| F2 | 🚨 High | `ZonePortal` mutations not invalidated in raw L1 cache; stale state can be inherited |
| F3 | 🚨 Medium | `userRewardInfo`/`getPendingRewards` bypass zone privacy gates |
| F4 | 🚨 High | Encrypted deposits check only `MintRecipient`; `Recipient` policy bypassed |
| F5 | 🚨 Medium | Fixed-gas wrapper normalizes `gas_used` but not `gas_refunded` |
| F6 | 🚨 High | Historical `eth_call` can pin unlimited synchronous L1 RPC retry loop |
| F7 | ⚠️ Medium | Follower nodes never establish raw-cache coverage; policy re-fetched every anchor |
| F9 | 🛡️ Low | L1-owned-slot write rejection is block-fatal, not per-tx revert (latent) |

## Finding F4 — Encrypted deposits bypass `Recipient` policy (file not in diff)

**File:** `crates/l1/src/block.rs:105-112`  
**Severity:** High  

`L1BlockDeposits::prepare` calls `policy_provider.is_authorized_async(d.token, dec.to, l1_block_number, AuthRole::MintRecipient)` only. Regular portal deposits enforce **both** `AuthRole::Recipient` and `AuthRole::MintRecipient`. When a compound TIP‑403 policy has a restrictive `recipient_policy_id` but a permissive `mint_recipient_policy_id`, regular deposits to a blocked address fail while encrypted deposits succeed. The zone then mints to that address, and upstream TIP-20 mint validation also checks only `MintRecipient`, so the missing constraint is never enforced.

**Fix:** Require both `AuthRole::MintRecipient` AND `AuthRole::Recipient` in `prepare` after decryption.

<details><summary>Reviewer Callouts</summary>

- ⚡ **advanceTempo ordering:** `ZoneInbox.advanceTempo`'s `OnlySequencer` guard calls `config.sequencer()` (pins anchor to pre-advance checkpoint) **before** `finalizeTempo`. On the sequencer-direct path (non-zero `msg.sender`), `advance_anchor(prev, prev+1)` would fail with `AnchorConflict`. Confirm `TEMPO_SYSTEM_TX_SENDER = Address::ZERO` short-circuit is the only production path.
- ⚡ **Offline CLI execution broken:** `ZoneEvmConfig::new_without_l1` points to `127.0.0.1:1` with `max_sync_attempts: Some(1)`. Reth CLI import/stage/re-execute will fail when any block reads a TIP-403 or TIP-20 policy slot. Recovery tooling needs a real L1 provider or policy witness.
- ⚡ **ReceivePolicyGuard hardfork gating:** Zone registers `ReceivePolicyGuard` unconditionally; upstream Tempo gates it at `is_t6()`. Historical/pre-T6 execution may diverge.
- ⚡ **Sequencer read non-determinism (pre-existing):** `TIP20Rules::is_sequencer` reads `latest_sequencer()` at the subscriber's current anchor, not the deterministic `tempoBlockNumber`. Narrow divergence risk during sequencer handoff.

</details>

crates/node/src/node.rs
0xrusowsky: addressed in [`ae31693` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/ae31693980652154970a2f1d6cb1ff86d1db334c)

crates/precompiles/src/ztip20/mod.rs
0xrusowsky: they have been disabled so not an issue

crates/l1/src/state/provider.rs
0xrusowsky: pre-existing issue that will be addressed in a follow-up PR

crates/l1/src/subscriber.rs
0xrusowsky: addressed in [`3cfe183` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/3cfe1839756f947a55e67059993067b4ed0e1903)

crates/precompiles/src/execution.rs
0xrusowsky: i think it's fine to leave as is for now, as it mirrors `main`'s behavior

crates/evm/src/database.rs
0xrusowsky: seems fine to leave as is for now

crates/payload/src/builder.rs
0xrusowsky: addressed in [`991a2df` (this PR)](https://github.com/tempoxyz/zones/pull/698/commits/991a2df1e8eccd4127aa9fec5996b9363ae30af5)

crates/l1/src/state/cache.rs
0xrusowsky: we can address in a follow-up as for now we will only have 1 sequencer

crates/node/src/node.rs
0xrusowsky: we can address in a follow-up as for now we will only have 1 sequencer

0xKitsune (approved):

PR #700, opened 2026-07-17
fix: align messenger with native factory ABI

legion2002:
Updates `ZoneMessenger` to decode the flattened `zones(uint32)` return ABI exposed by the TIP-1091 native factory.

The reference `IZoneFactory`, Solidity factory, and `SwapAndDepositRouter` use the same getter shape so the repository remains internally consistent. A focused regression test exercises `relayMessage()` against the flattened payload.

This addresses https://github.com/tempoxyz/tempo/pull/6874#discussion_r3596182128.

The bundled zone dev genesis is regenerated because the factory creation bytecode embeds the messenger creation bytecode.

Validation:
- `FOUNDRY_HARDFORK=cancun forge fmt --check`
- `FOUNDRY_HARDFORK=cancun forge build --no-lint`
- `FOUNDRY_HARDFORK=cancun forge test --match-contract ZoneMessengerFactoryAbiTest`
- `cargo test -p zone-node zone_test_genesis_factory_bytecode_matches_foundry_artifact -- --nocapture`
- `git diff --check`


legion2002:
Superseded by #740.

PR #701, opened 2026-07-17
feat: drop `PolicyProvider`

0xrusowsky:
stacked on top of https://github.com/tempoxyz/zones/pull/698

fully drops the `PolicyProvider`


crates/l1/src/block.rs
```diff
@@ -96,42 +92,9 @@ impl L1BlockDeposits {
...
-                        } else {
-                            warn!(
-                                target: "zone::engine",
-                                sender = %d.sender,
-                                recipient = %dec.to,
-                                token = %d.token,
-                                amount = %d.amount,
-                                "Encrypted deposit recipient unauthorized; queuing deposit bounce-back"
-                            );
-                            queued.rejected = true;
```
klkvr, line 130: do we now never reject deposits? is it safe to just remove this logic?

0xrusowsky: they are rejected at execution when the 403 logic runs

0xrusowsky: basically it would be this code which calls `mint()` on the zone's TIP20:
- regular deposit: https://github.com/tempoxyz/zones/blob/95366b02cddd15d771ba1195c26252432a5e039b/specs/ref-impls/src/zone/ZoneInbox.sol#L236-L245
- encrypted deposit: https://github.com/tempoxyz/zones/blob/95366b02cddd15d771ba1195c26252432a5e039b/specs/ref-impls/src/zone/ZoneInbox.sol#L320-L325

0xKitsune: Yes since this is rejected at execution time, should be safe to remove here.

0xrusowsky:
cyclops audit note="focus only on ensuring that the changes in this PR are sound. the base branch is being audited in parallel"

crates/l1/src/block.rs
```diff
@@ -18,15 +16,13 @@ pub struct L1BlockDeposits {
 impl L1BlockDeposits {
     /// Prepare all deposits for the payload builder.
     ///
-    /// Decrypts encrypted deposits, checks TIP-403 policy authorization,
-    /// and ABI-encodes everything into the types the `advanceTempo` call expects.
-    /// The resulting [`PreparedL1Block`] is ready to be passed through payload
-    /// attributes to the builder.
+    /// Decrypts encrypted deposits and ABI-encodes into the types the `advanceTempo` call expects.
+    /// Mint-recipient policy is enforced by upstream TIP-20 after the L1 state is anchored.
```
0xKitsune, line 20: Flagging that this now depends on the raw-cache invalidation issue from #698. Once this precheck is removed, encrypted deposits rely on that cache for authorization, so a stale hit can mint where a cold node would bounce.

crates/node/README.md
```diff
@@ -136,8 +136,8 @@ using Tempo's upstream policy implementation over anchored raw L1 state:
    normal EVM gas, warming, and storage accounting, while persistent writes to
    L1-owned slots are rejected.
 3. **Tempo TIP-20 and TIP-403 precompiles** — execute the upstream business
-   logic against that composed storage view, replacing the zone's duplicated
-   policy dispatch. Missing or invalid anchored state fails closed, and zone
+   logic against that composed storage view, without a separate zone policy
```
0xKitsune, line 135: Can we clean up the remaining PolicyProvider docs here too? The diagram and note still describe PolicyCache/PolicyResolutionTask, and the payload attrs still say policy checks happen in the engine.

0xKitsune (dismissed):
Builder docs still claims TIP403 checks happen during preparation. `crates/payload/src/attrs.rs:39` and `crates/payload/src/builder.rs:723` say the engine has already performed policy checks. After this PR, preparation only decrypts and encodes deposits, policy enforcement occurs later during advanceTempo through the TIP-20 mint. These comments should be updated but otherwise lgtm.

0xKitsune (approved):

PR #702, opened 2026-07-17
ci: publish tempo-zone-xtask image with cast and ref-impl artifacts

Zygimantass:
Adds a `tempo-zone-xtask` image target:

- new `solidity` Docker stage compiles `specs/ref-impls` with Foundry (submodules now checked out recursively in CI)
- final image ships `tempo-xtask`, `cast`, `jq`, and the compiled artifacts at `/app/specs/ref-impls/out` (the xtask default `--specs-out`)
- bake target + metadata wired into the existing Depot build, published as `ghcr.io/tempoxyz/tempo-zone-xtask`

Needed by the `zone-regenesis` Argo workflow (tempoxyz/workflows) to re-register zones on L1 (deploy a fresh ZoneFactory, run `create-zone`, regenerate genesis) after portal ABI changes — currently blocking zone-009 withdrawals (deployed portal predates the `processWithdrawal` selector change).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,3 +3,4 @@ contrib/
 docs/
 localnet/
 scripts/
+target/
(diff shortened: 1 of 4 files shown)


Zygimantass:
+1

rkrasiuk (approved):

PR #703, opened 2026-07-17
feat: wire sequencers individual secp256k1 key into multi-sequencer manifest.

adityapk00:
Wire up the sequencer's individual secp256k1 key that the leader + followers will use to sign blocks while submitting signatures to the zone portal, similar to the ed25519 key.

Note: This is not being used yet (subsequent PR)


mattsse (approved):
lgtm

PR #706, opened 2026-07-17
refactor: name refund recipients by domain

0xalpharush:
Stack 1 of 2.

Mechanical-only recipient naming cleanup split out ahead of the closed-loop behavior changes:

- `bouncebackRecipient` / `bounceback_recipient` → `tempoRefundRecipient` / `tempo_refund_recipient`
- callback `refundRecipient` → `tempoRefundRecipient`
- `fallbackRecipient` / `fallback_recipient` → `zoneFallbackRecipient` / `zone_fallback_recipient`
- `_fallbackRecipients` → `_zoneFallbackRecipients`

Applies the mapping consistently across Solidity, the spec/reference implementation, tests, Rust bindings and models, RPC/tooling, and generated storage layout. No behavior change is intended.

Stack: this PR targets `main`; the second PR targets this branch. PR #696 and `codex/closed-loop-zone` are unchanged.

PR #707, opened 2026-07-17
feat: enforce closed-loop zone flows

0xalpharush:
Stack 2 of 3; based on #706.

Replays the closed-loop Zone changes on top of the mechanical recipient rename so naming and behavior changes are reviewable separately. It also:

- accepts duplicate and zero-address entries in initial membership arrays
- allows `setAllowedAccount` and `setZoneGateway` to enable or disable the zero address
- removes the obsolete zero-address caveat from the spec
- refreshes the bundled ZoneFactory/ZonePortal bytecode and final Portal storage layout

PR #696 and `codex/closed-loop-zone` are unchanged.


0xalpharush:
cyclops audit fast

0xrusowsky (dismissed):

crates/contracts/src/precompiles/zone_factory.rs
```diff
@@ -29,6 +29,8 @@ crate::sol! {
         }
         struct CreateZoneParams {
             address initialToken;
+            address[] allowedAccounts;
```
legion2002: This requires a corresponding tempo factory 1091 change

crates/contracts/src/precompiles/zone_portal.rs
```diff
@@ -13,6 +13,11 @@ crate::sol! {
     #[derive(Debug)]
     contract ZonePortal {
         // -- Shared types --
+        enum Role {
```
legion2002, line 17: think we should remove this enum, and just do `isAllowedAccount` `isAllowedGateway` otherwise we'll also have to cover the both account and gateway case.

specs/ref-impls/src/zone/ZoneOutbox.sol
```diff
@@ -252,13 +252,19 @@ contract ZoneOutbox is IZoneOutbox {
...
-        _validateGasLimit(gasLimit);
-
-        // Limit callback data size to prevent storage bloat and hash computation abuse
+        // Bound callback data before queueing.
         if (data.length > MAX_CALLBACK_DATA_SIZE) {
             revert CallbackDataTooLarge();
         }
 
+        if (gasLimit == 0) {
+            if (config.isZoneGateway(to)) revert IZonePortal.InvalidCallbackTarget();
```
legion2002, line 261: These changes need to be updated in the zone outbox precompile, since that got merged yesterday

specs/ref-impls/src/zone/ZoneOutbox.sol
```diff
@@ -252,13 +252,19 @@ contract ZoneOutbox is IZoneOutbox {
...
+            if (!config.isAllowedAccount(to)) revert IZonePortal.AccountNotAllowed(to);
+        } else {
+            _validateGasLimit(gasLimit);
+            if (!config.isZoneGateway(to)) revert IZonePortal.InvalidCallbackTarget();
```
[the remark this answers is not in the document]
legion2002: yeah we probably need to update this in some script somewhere?

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -278,6 +295,16 @@ contract ZonePortal is IZonePortal {
         emit AdminTransferred(previousAdmin, admin);
     }
 
+    /// @notice Assign an account's role across portal flows.
+    function setRole(address account, Role next) external onlyAdmin {
```
legion2002: prefer `setAllowedAccount` and `setGateway` over `setRole`

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -138,18 +139,24 @@ contract ZonePortal is IZonePortal {
...
+    mapping(address => Role) public role;
+
     /*//////////////////////////////////////////////////////////////
                              INITIALIZATION
     //////////////////////////////////////////////////////////////*/
 
     function initialize(
         uint32 _zoneId,
         address _initialToken,
+        address[] calldata _allowedAccounts,
```
legion2002, line 166: although we have this initialize function, we need to actually set the storage slots manually in the factory, because factory precompile cannot call an EVM function.

should be updated in the L1 PR

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -138,18 +139,24 @@ contract ZonePortal is IZonePortal {
     /// @dev These values must remain in account storage so each delegatecall proxy observes its
     ///      own metadata. Keep them after the established slots read directly by zone contracts.
     uint32 public zoneId;
+    /// @notice Fixed callback messenger assigned during initialization.
     address public messenger;
     address public verifier;
     uint64 public genesisTempoBlockNumber;
     bool internal _initialized;
 
+    /// @notice Mutually exclusive authorization role assigned to each Tempo account.
```
legion2002: is there a reason they are mutually exclusive? 
if yes, then this negates some of my other comments.

specs/ref-impls/src/tempo/ZoneFactory.sol
```diff
@@ -60,10 +60,29 @@ contract ZoneFactory is IZoneFactory {
...
         if (!ITIP20Factory(StdPrecompiles.TIP20_FACTORY_ADDRESS).isTIP20(params.initialToken)) {
             revert InvalidToken();
         }
+        if (params.zoneGateways.length == 0 || params.allowedAccounts.length == 0) {
+            revert InvalidClosedLoopConfig();
+        }
+
+        for (uint256 i; i < params.allowedAccounts.length; ++i) {
+            address account = params.allowedAccounts[i];
+            if (account == _messenger) {
```
legion2002: is there a security reason the messenger is not allowed to be the sender? if not would like to avoid this loop

crates/primitives/src/constants.rs
```diff
@@ -72,6 +72,13 @@ pub const PORTAL_PENDING_SEQUENCER_SLOT: B256 = {
     B256::new(bytes)
 };
 
+/// ZonePortal storage slot 19: `portalRole` (mapping(address => Role)).
+pub const PORTAL_ROLE_SLOT: B256 = {
```
legion2002: we should use the portal storage slots exporeted using this PR - https://github.com/tempoxyz/tempo/pull/6923

crates/contracts/src/precompiles/zone_portal.rs
0xalpharush: we had two different mapping's and @0xrusowsky recommended this
https://github.com/tempoxyz/zones/pull/707/commits/681f538d51ee57a01d0f618aecbdcc5c70ae5ec5

PR #709, opened 2026-07-17
feat: zones can restrict deposit/withdrawal and callback targets

0xalpharush:
Stack 3 of 4; based on #707.

Makes account allowlist and callback gateway enforcement independently mutable and open by default:

- stores both enforcement bits in a dedicated portal slot while retaining membership mappings
- defaults zone creation and dev flows to open/open without implicit admin/sequencer membership or a required gateway
- treats open-access callbacks as open loop, so cross-zone transfers need neither gateway registration nor source queue advancement
- propagates live modes through `ZoneConfig` and RPC, with event-driven L1 cache updates
- emits initial mode and membership state through the existing update events so creation is replayable
- seeds synthetic closed fixtures so withdrawal batching tests cannot stall on missing L1 storage

Validated with the full Forge suite (628 tests), Rust L1 tests and checks, mutable-mode and cross-zone E2Es, both former hang regressions, and storage/genesis parity.

L1 cache history coverage and follower freshness are intentionally deferred to #698 and remain required before merge.

The factory ABI, portal storage, and bundled bytecode changed, so this requires a fresh ZoneFactory deployment. No Moderato deployment is included.


0xalpharush:
cyclops audit

Welith:
The route matrix requires the VaultAdapter or recovery address to receive on-chain `VAULT` refunds while forbidding every direct `VAULT` withdrawal. In this PR, both operations use the same global `Role.Account` predicate and the same Portal → recipient TIP-20 transfer:

- making an address eligible for a `VAULT` refund also makes it a valid direct `VAULT` withdrawal recipient;
- removing `Account` prevents direct withdrawal but also parks the refund;
- TIP-403 cannot distinguish the routes because both look identical at the token layer.

There is no stable Closed-mode role assignment that represents `refund allowed / direct withdrawal forbidden`.

0xalpharush:
replaced by https://github.com/tempoxyz/zones/pull/810

PR #712, opened 2026-07-18
Draft: persist verified batch observations

decofe:
## Summary
- persist receipt-root-verified `BatchSubmitted` observations under the Zone node datadir
- use the local index for withdrawal restart recovery; retain the verified receipt scan as a cold-index fallback
- discard indexed observations at and after a replaced L1 height

## Verification
- `cargo +nightly fmt --all --check`
- `git diff --check`
- `cargo check -p zone-l1 -p zone-sequencer -p zone-node` is blocked before compilation because the pinned Commonware Git dependency returns HTTP 401

## Benchmark
- Not run: this is stacked on #711 and the available benchmark environment is currently blocked by the local chain failing to advance.

Prompted by: @0xalpharush

### DIFF
--- a/crates/l1/Cargo.toml
+++ b/crates/l1/Cargo.toml
@@ -54,6 +54,7 @@ k256.workspace = true
 metrics.workspace = true
 parking_lot.workspace = true
 serde = { workspace = true, features = ["derive"] }
+serde_json.workspace = true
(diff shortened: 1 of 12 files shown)


0xalpharush:
should probably just be soft confirmed batches

PR #713, opened 2026-07-19
feat: commonware p2p config for zones

adityapk00:
Setup commonware p2p config for the zones. Mainly reduce the connection cooldown (60s -> 1s) which allows to recover from temporary disconnects quickly.

### DIFF
--- a/crates/p2p/src/network.rs
+++ b/crates/p2p/src/network.rs
@@ -1,4 +1,4 @@
-use std::net::SocketAddr;
+use std::{net::SocketAddr, time::Duration};
 
 use alloy_primitives::Address as EthereumAddress;
 use commonware_cryptography::ed25519::{PrivateKey, PublicKey};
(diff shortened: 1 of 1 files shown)


crates/p2p/src/network.rs
```diff
@@ -40,6 +40,43 @@ impl P2pNetworkId {
...
+    // At most four legitimate remote peers are expected, so 32 leaves generous rollout headroom.
+    config.max_concurrent_handshakes = NZU32!(32);
+    // Eight attempts per second permits fast restarts while limiting a single source IP.
+    config.allowed_handshake_rate_per_ip = Quota::per_second(NZU32!(8));
+    // Thirty-two attempts per second handles the whole small peer set restarting together.
+    config.allowed_handshake_rate_per_subnet = Quota::per_second(NZU32!(32));
+    // Ten-second pings detect idle broken connections with negligible overhead.
+    config.ping_frequency = Duration::from_secs(10);
+    // One dial per 250 ms reconnects a full five-node mesh in about one second per node.
+    config.dial_frequency = Duration::from_millis(250);
```
mattsse, line 75: I believe tempo uses 1s, but has an actual distributed set and 500ms for local so this seems fine

mattsse (approved):

PR #715, opened 2026-07-20
feat(xtask): support the native ZoneFactory precompile in create-zone

Zygimantass:
The `zone-factory-hf` devnet runs the TIP-1091 native ZoneFactory precompile at `0x5aF2000000000000000000000000000000000000`. It has no bytecode, no `verifier()`/`messenger()` getters (calling them reverts `UnknownFunctionSelector`), and its `createZone` takes `CreateZoneParams` **without** the `verifier` field — verifier (`0x5A56...`) and messenger (`0x5A4D...`) are protocol constants.

`tempo-xtask create-zone` now detects the native factory by the absence of bytecode at the factory address and switches to the precompile ABI. The reference Solidity factory path is unchanged, and the `ZoneCreated` event is identical in both variants so receipt decoding is shared.

Found while pointing the `zone-regenesis` Argo workflow at the zone-factory-hf devnet:

```
Fetching verifier address from ZoneFactory...
Error: failed to create zone
Caused by: server returned an error response: error code 3: execution reverted,
data: "0xaa4bc69a2b7ac3f3..."  # UnknownFunctionSelector(verifier())
```

### DIFF
--- a/crates/node/src/node.rs
+++ b/crates/node/src/node.rs
@@ -78,10 +78,18 @@ use zone_sequencer::{BatchAnchorConfig, ZoneSequencerConfig, spawn_zone_sequence
 /// Returns a known Tempo chain spec for an L1 chain ID.
 ///
 /// Tempo Anvil uses chain ID 31337 and the same hardfork schedule as Tempo DEV (1337).
+///
+/// Additional dev-schedule L1 chain IDs (devnets that activate all Tempo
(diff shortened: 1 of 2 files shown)


mattsse (commented):
@legion2002 we merged https://github.com/tempoxyz/tempo/pull/6883
which removed caller-supplied genesis data, do we need to do the same here?

mattsse (approved):

Zygimantass:
+1

mattsse (approved):

PR #717, opened 2026-07-20
feat: Followers attest + sign blocks from leader sequencer

adityapk00:
- After recieving blocks from the leader sequencer, followers sign + attest the blocks (if execution is valid) and return them back to the leader
- Leader collects and stores the signatures, which it will use to submit to L1 zoneportal during submitBatch() (The submit signatures to L1 is not in this PR yet)


specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -164,12 +181,16 @@ contract ZonePortal is IZonePortal {
...
         admin = _admin;
-        sequencer = _sequencer;
         verifier = _verifier;
         blockHash = _genesisBlockHash;
         genesisTempoBlockNumber = _genesisTempoBlockNumber;
         rpcUrl = _rpcUrl;
 
+        uint256 initialSequencerCount =
+            _validateSequencerSet(initialSequencers, initialQuorum);
+        _writeSequencerSet(initialSequencers, initialQuorum, initialSequencerCount);
```
0xalpharush: This doesn't bump the version so initializing with a multi-sequencer setup still remains in legacy, single-sequencer mode

adityapk00:
Superseded by #727 , #751 , #788

PR #723, opened 2026-07-20
fix(evm): clear prewarm nonce context during execution

Rjected:
We have to do this in the tempo builder as well:
https://github.com/tempoxyz/tempo/blob/9cb1db7ef5d30f4f1b336184b70983ea6f38e70d/crates/evm/src/block.rs#L597-L600


crates/evm/src/executor.rs
mattsse, line 112: we only call this once, so I dont think we need an extra function for this and instead mirror what we do for the tempo executor which has this inline,

since we likely need to unify these again eventually it would be better if they look as similar as possible

mattsse (changes requested):
style nit

crates/evm/src/executor.rs
Rjected: makes sense, copied the line and comment

mattsse (approved):

PR #724, opened 2026-07-20
perf(node): use executed payload tree insertion

Rjected:
Zone block production now waits for Reth to insert the locally executed payload through its built-payload fast path, rather than converting it into `newPayload` and potentially executing it again.

The L1 queue advances only after the matching tree-insertion event. A bounded wait leaves the L1 block queued for retry if insertion does not arrive, and the subsequent forkchoice update canonicalizes the inserted block.


crates/node/src/engine.rs
```diff
@@ -245,17 +253,11 @@ impl ZoneEngine {
         };
 
         let header = payload.block().sealed_header().clone();
-        let block_number = header.number();
-        let res = self.to_engine.new_payload(payload.into()).await?;
```
mattsse, line 249: there could a race condition here in theory, where the fcu below arrives before the executed block update

we could check if the fcu returns syncing then this would be the case and then insert the block, but perhaps not even realistic

PR #725, opened 2026-07-20
fix(zone): authenticate empty-checkpoint bootstrap

legion2002:
## Summary

Lets a Zone start with an empty `TempoState` checkpoint without requiring a Tempo block hash before the Portal exists.

The first `advanceTempo` import is accepted only when the imported Tempo state proves that the configured Portal exists (`admin != 0`). After that import, ordinary parent-hash and consecutive-number validation applies.

## Bootstrap and replay safety

- Derives the replay cursor from the confirmed `createZone` receipt: a Portal created in block `X` stores anchor `X - 1` and replays starting at `X`.
- Requires `--l1.genesis-block-number` while the local checkpoint is empty; normal create/dev/zone-up and benchmark flows propagate `tempoAnchorBlock` automatically.
- Pins a replica first import to its unconsumed replay cursor.
- Requires executor authorization for the block-opening system transaction, so `eth_call` cannot drive consensus state merely by using the zero sender.
- Bounds L1-backed precompile reads by the independently observed L1 head and uses one startup snapshot for tokens, encryption keys, and leadership.
- Fails closed when the empty-checkpoint Portal proof is missing or malformed.

## Canonical genesis

The first proof must validate the canonical Zone genesis derived from authenticated creation data (`zoneId`, registered Portal, creation-time admin, initial leader, and first enabled token) plus protocol-fixed parameters.

The production `create-zone` command no longer exposes overrides for base fee, gas limit, artifact directory, or standard utility predeploys. Those fixed values live in one canonical generator profile, while the standalone low-level generator remains available for development.

## Compatibility / rollout

- **No ZonePortal ABI or storage-layout change.** The L1 Portal does not need a breaking upgrade for this PR.
- **Breaking Zone genesis/consensus change.** Existing Zone databases created from the old anchored genesis are not migrated in place. Pre-launch deployments should create a fresh Portal/Zone and regenerate genesis with this version.
- `create-zone` no longer accepts the removed genesis customization flags.

This PR incorporates the replay-cursor and observed-head protections that were previously split into follow-up work, so it is intended to merge before those changes independently.

## Validation

- `cargo clippy --workspace --all-targets -- -D warnings`
- Rust unit tests: 141 precompile, 79 L1, and 55 node tests passed
- Rust integration tests passed for synthetic Tempo advancement, delayed Portal inclusion/replay, and a real two-online-sequencer settlement certificate
- `forge build --root specs/ref-impls --no-lint`
- Full Foundry suite: 678 tests passed
- Regenerated `crates/node/assets/zone-dev-genesis.json` from the rebuilt reference contracts
- `bash -n contrib/bench/provision-topology.sh`


legion2002:
Superseded by #740.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -133,7 +133,6 @@ contract ZonePortal is IZonePortal {
     uint32 public zoneId;
     address public messenger;
     address public verifier;
-    uint64 public genesisTempoBlockNumber;
     bool internal _initialized;
```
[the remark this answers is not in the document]
legion2002: Addressed in cb791262. The rebased implementation no longer uses the removed getter: startup backfill keeps the pre-creation L1 block separately, while TempoState starts from the required empty checkpoint.

legion2002:
cyclops audit

crates/precompiles/src/tempo_state.rs
```diff
@@ -127,17 +117,25 @@ impl TempoState {
...
 
-        if header.parent_hash() != prev_block_hash {
-            return self.revert_error(TempoStateAbi::InvalidParentHash {});
-        }
-        if prev_block_number.checked_add(1) != Some(header.number()) {
-            return self.revert_error(TempoStateAbi::InvalidBlockNumber {});
-        }
+        if prev_block_hash.is_zero() {
+            let admin =
+                l1.read_l1_storage(l1.portal_address(), PORTAL_ADMIN_SLOT, header.number())?;
```
dankrad: This finding is correct but expected behaviour: All L1 blocks (not just the first imported one) are unauthenticated at the time of zone import and only authenticated when the zone settles on L1. 

We should fix RPC queries so that permanent failures (such as header.number() being out of range) fail rather than lead to retry.

xtask/src/create_zone.rs
```diff
@@ -171,7 +152,7 @@ impl CreateZone {
             "initialToken": format!("{}", self.initial_token),
             "admin": format!("{}", self.admin),
             "sequencer": format!("{}", self.sequencer),
-            "tempoAnchorBlock": anchor_header.inner.number,
+            "tempoAnchorBlock": anchor_block_number,
```
legion2002, line 304: Addressed in https://github.com/tempoxyz/zones/commit/a6a7980781a82e24d5a759d9af9f6139c5860659

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -985,8 +985,15 @@ contract ZonePortal is IZonePortal {
...
-
-        if (recentTempoBlockNumber == 0) {
+        bool isBootstrap =
+            blockHash == bytes32(0) && tempoBlockNumber == 0 && recentTempoBlockNumber == 0;
+
+        if (isBootstrap) {
+            // Bootstrap mode: the first proof derives the canonical genesis block from zoneId and
+            // proves that TempoState still has an empty checkpoint.
+            anchorBlockNumber = 0;
+            anchorBlockHash = bytes32(0);
```
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/a6a7980781a82e24d5a759d9af9f6139c5860659

crates/sequencer/src/monitor.rs
```diff
@@ -200,10 +200,41 @@ impl ZoneMonitor {
...
+            let genesis_block = provider
+                .get_block_by_number(0.into())
+                .await?
+                .ok_or_else(|| eyre::eyre!("canonical zone genesis block is unavailable"))?;
+            let genesis_hash = genesis_block.header.hash;
+            let bootstrap = BatchData {
+                zone_height: 0,
+                tempo_block_number: 0,
+                prev_block_hash: B256::ZERO,
+                next_block_hash: genesis_hash,
```
0xKitsune: Could we validate this hash before settling it?

`create-zone` builds genesis from configurable values, so `zoneId` alone does not determine the hash. The factory also does not commit a genesis hash.

The stub verifier accepts the local value today, but a real verifier would not know which genesis is canonical.

Can we make genesis deterministic or commit and verify its hash at zone creation?

crates/l1/src/state/provider.rs
```diff
@@ -247,20 +252,46 @@ impl L1StateProvider {
...
+                |error| warn!(%address, %slot, block_number, %error, "eth_getStorageAt RPC call failed"),
+            )?;
 
         let result = B256::from(value.to_be_bytes());
         debug!(%address, %slot, block_number, %result, "fetched L1 storage slot from RPC");
         Ok(result)
     }
 }
 
+fn is_permanent_rpc_failure(error: &TransportError) -> bool {
```
0xKitsune: What is this trying to protect against? If the goal is to stop an invalid block number from hanging EVM execution, let’s use one bounded retry policy in `RetryBackoffLayer` and return after it is exhausted. 

Otherwise, can we remove this classifier and the outer retry loop? As written, the two retry policies disagree some transient errors fail immediately, while Alloy’s `Max retries exceeded` wrapper can restart the loop forever.

specs/spec.md
```diff
@@ -1436,9 +1436,9 @@ The portal passes its `zoneId`, computes `anchorBlockNumber` and `anchorBlockHas
...
 
-After bootstrap, if `recentTempoBlockNumber` is `0`, the portal looks up `tempoBlockNumber` directly from EIP-2935. The proof must show that the zone's `tempoBlockHash` matches this hash. On the first Tempo import, the proof must additionally show that the portal's `sequencer` slot is non-zero in that block's state, proving that the imported block is not from before portal creation.
+After bootstrap, if `recentTempoBlockNumber` is `0`, the portal looks up `tempoBlockNumber` directly from EIP-2935. The proof must show that the zone's `tempoBlockHash` matches this hash. On the first Tempo import, the proof must additionally show that the portal's `admin` slot is non-zero in that block's state, proving that the imported block is not from before portal creation.
```
0xKitsune: nit: Looks like the proof requirements below also need a spec update. They still refer to the `sequencer` slot.

crates/l1/src/state/provider.rs
legion2002: addressed in this PR and https://github.com/tempoxyz/zones/pull/853

crates/sequencer/src/monitor.rs
legion2002: yes we can update it to be determinstic. @dankrad do you have any thoughts?

legion2002:
cyclops audit

crates/node/src/replication.rs
```diff
@@ -878,6 +878,12 @@ fn validate_l1_checkpoint_transition(
     local_hash: B256,
     zone_block_number: u64,
 ) -> eyre::Result<()> {
+    // Zero hash denotes an uninitialized checkpoint. The first finalized Tempo header may be
+    // any height and parent because a zero-genesis zone deliberately carries no L1 commitment.
+    if local_hash.is_zero() {
```
0xKitsune, line 1339: The empty-checkpoint path accepts any observed L1 height. A leader can skip the portal-creation block, causing followers to prune earlier events such as the initial `TokenEnabled`. Can we bind the first import to the next replay-cursor block?

crates/precompiles/src/tempo_state.rs
```diff
@@ -124,14 +113,21 @@ impl TempoState {
         if !header_cursor.is_empty() {
             return Err(TempoStateError::invalid_rlp_data().into());
         }
-        if header.parent_hash() != prev_block_hash {
-            return Err(TempoStateError::invalid_parent_hash().into());
-        }
-        if prev_block_number.checked_add(1) != Some(header.number()) {
-            return Err(TempoStateError::invalid_block_number().into());
+        if prev_block_hash.is_zero() {
+            let admin = l1.read_l1_storage(l1.portal(), PORTAL_ADMIN_SLOT, header.number())?;
```
0xKitsune: The first import issues an L1 read at the caller-supplied header number before that height is bound to finalized local state. This can stall RPC execution or cache non-final values. Can we validate the height before reading? #853 bounds it but does not enforce the replay cursor.

crates/precompiles/src/tempo_state.rs
```diff
@@ -56,20 +56,9 @@ impl TempoState {
...
-        })?;
-        if !cursor.is_empty() {
-            return Err(TempoPrecompileError::Fatal(
-                "invalid Tempo genesis header RLP: trailing bytes after header".into(),
-            ));
-        }
-        self.write_checkpoint(header_rlp, header.number())?;
-        Ok(())
+    /// Initializes the predeploy account code with an empty Tempo checkpoint.
+    pub fn initialize(&mut self) -> tempo_precompiles::Result<()> {
```
0xKitsune, line 60: An empty checkpoint also presents as L1 block zero, so pre-bootstrap reads query block zero and offline re-execution cannot process the first block. Can we represent checkpoint presence explicitly or fail before issuing an L1 read?

crates/precompiles/src/tempo_state.rs
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/32f37f9d91a27832f63c85cc5535d4282eef3465

legion2002: Addressed in https://github.com/tempoxyz/zones/commit/32f37f9d91a27832f63c85cc5535d4282eef3465

crates/precompiles/src/tempo_state.rs
legion2002, line 60: Addressed in https://github.com/tempoxyz/zones/commit/6d08e389901d27bc36784ad4cc5ba812c7bbc5fa

legion2002: Addressed in https://github.com/tempoxyz/zones/commit/6d08e389901d27bc36784ad4cc5ba812c7bbc5fa

crates/node/src/replication.rs
legion2002, line 1339: Addressed in https://github.com/tempoxyz/zones/commit/6d08e389901d27bc36784ad4cc5ba812c7bbc5fa

legion2002: Addressed in https://github.com/tempoxyz/zones/commit/6d08e389901d27bc36784ad4cc5ba812c7bbc5fa

crates/sequencer/src/monitor.rs
legion2002: Addressed in a341934f. Production `create-zone` now builds through a single canonical genesis profile: the variable inputs are authenticated creation data (zone ID/derived chain ID, registered Portal, creation-time admin, first sequencer/leader, and initial token), while base fee, gas limit, reference artifacts, standard predeploys, and funding are fixed protocol parameters. The spec now requires the first proof to verify those creation fields against Tempo state and reproduce the full canonical genesis; a locally supplied genesis file is not authoritative. No Portal ABI or storage-layout change is required.

legion2002:
cyclops audit

PR #727, opened 2026-07-20
feat: Forward transactions from rpc/follower to block-building sequencer

adityapk00:
Transactions submitted to the RPC (follower) sequencer nodes should be forwarded to the block building leader node.


adityapk00:
Cyclops issues have been addressed:
1. The dropped txns are all now retried by the follower nodes in case the leader is down/restarting/networking issues
2. Fix the CW limits - we don't need additional limits for txn, since the follower node will already do all its checks before admitting the submitting txn into the mempool/forwarding it to leader sequencer

legion2002 (changes requested):
## Cyclops Security Review — manual consolidation

Reviewed the 3-worker audit from `9e3c725` and re-checked every verified finding against the current head `6e03c6b`.

The new txpool reconciliation loop resolves the live-listener/drop-on-full loss, Commonware send-throttling loss, and follower-restart backup replay findings. Removing the separate 512 KiB forwarding cap also resolves the config-coupled pool-valid transaction drop. Four findings remain:

### High — transaction-channel buffering is still bounded only by the 20 MiB network-wide frame cap

`TRANSACTION_BACKLOG` remains 1,024 messages, while Commonware is configured with the block-sized `MAX_MESSAGE_SIZE = 20 MiB` for every channel. Commonware accepts and enqueues a transaction-channel frame before `run_receivers` checks the peer role or the leader decodes/revalidates it. The current revision removed the old post-dequeue 512 KiB check entirely, so one compromised manifest follower can occupy roughly 20 GiB in the shared Commonware transaction queue, plus the downstream event queues, with arbitrary near-20 MiB frames. This can OOM or severely degrade the only block-building leader and starve honest followers.

An honest follower's txpool checks do not protect this boundary: the leader explicitly treats follower validation as untrusted, and a compromised follower key can send channel frames directly. Please enforce a per-channel frame cap before Commonware's application queue (or isolate transaction transport), and derive backlog/rate limits from an aggregate byte budget with per-peer fairness.

### Medium — a transaction rejected by follower RPC can still be forwarded and executed

Reth emits `NewTransactionEvent` before it rewrites a provisionally inserted-and-discarded result to `DiscardedOnInsert`. The new immediate listener branch still calls `try_queue_transaction` without checking `pool.contains(hash)`. Under pool pressure, the follower can return `txpool is full` while the less-loaded leader accepts and executes the transaction. A client that acts on the explicit rejection may retry or replace it and perform the action twice.

Check current pool membership before the immediate queue operation (and add the `DiscardedOnInsert` regression test from the audit); reconciliation already naturally excludes non-members.

### Low — follower `Local` admission and leader `External` admission are not equivalent

Raw RPC submissions enter the follower pool as `TransactionOrigin::Local`, but the leader re-adds the bytes with `add_external_transaction`. Local transactions receive policy exemptions in Reth, including the configured minimum-priority-fee check, so a follower can return a successful hash for a transaction the leader deterministically rejects and then retries every two seconds indefinitely.

Because followers are not trusted, the safer fix is to apply the leader's external-admission policy on the follower before returning RPC success, rather than granting forwarded traffic local privileges at the leader.

### Low — the new channel is added without changing the wire version

This PR registers `TRANSACTION_CHANNEL = 3` while `WIRE_PROTOCOL_VERSION` remains `0`. During a follower-first/mixed-version rollout, a valid transaction sent to an upgraded follower reaches an old leader as an unregistered channel; pinned Commonware treats that as connection-fatal for the multiplexed peer connection, which also carries block replication.

Bump the wire namespace/version or add capability negotiation/unknown-optional-channel handling, and document the required rollout order.

All repository CI checks currently pass; these are integration/security semantics not covered by the existing tests.

adityapk00:
Addressing cyclops comments:

>a transaction rejected by follower RPC can still be forwarded and executed

Fixed

> the new channel is added without changing the wire version

This is not live, so no backward compatibility needed, no need to change wire version

> transaction-channel buffering is still bounded only by the 20 MiB network-wide frame cap
>  follower Local admission and leader External admission are not equivalent

The sequencers are inside the trust boundary (we're running all sequencers in the zone - A sequencer key compromise can halt the chain). Adding these additional bounds is not really needed IMO

.config/nextest.toml
```diff
@@ -7,4 +7,4 @@ test-threads = "num-cpus"
 # resource contention and flaky timeouts on CI runners.
 [[profile.ci.overrides]]
 filter = "test(/^(e2e|l1_e2e|restart_e2e)::/) | test(advance_tempo)"
-threads-required = 8
```
joshieDo, line 10: why?

adityapk00: sorry, this shouldn't have been changed. reverted

mattsse (dismissed):
LGTM

legion2002 (approved):

legion2002 (approved):

legion2002 (dismissed):

legion2002 (approved):

PR #729, opened 2026-07-21
refactor(portal): group transfer storage

legion2002:
Stacks on #669 and coordinates with tempoxyz/tempo#6902. Groups each active transfer authority beside its pending replacement and updates the cross-domain slot constants, generated layout, and layout regression tests.


legion2002:
not needed

PR #733, opened 2026-07-21
feat(primitives): remove `ZoneHeader`

rkrasiuk:
## Summary

Remove the redundant `ZoneHeader` type and use Tempo’s canonical `TempoHeader` for zone blocks.

## Protocol version

`protocol_version` was specific to the simplified `ZoneHeader` and is not part of Tempo’s canonical header. This change therefore removes it from the specified zone header and prover witness instead of encoding it into an unrelated Tempo header field.

Zone nodes and provers select execution rules from the imported Tempo block and the active Tempo fork schedule. Unsupported forks must cause the node or prover to halt rather than relying on a zone-specific version committed in the block hash.

## Changes

- Remove `zone_primitives::ZoneHeader` and its custom RLP/hash implementation.
- Remove `protocol_version` from the specified zone header and prover witness.
- Update the specification and invariants to define canonical Tempo header hashing.


crates/evm/src/lib.rs
```diff
@@ -241,7 +246,7 @@ impl BlockAssembler<ZoneEvmConfig> for ZoneBlockAssembler {
             ),
             None,
             None,
-            None,
+            Some(Bloom::ZERO),
```
mattsse: I dont think we should alter the bloom here, this is already redacted when it crosses the rpc boundary, but this will likely cause issues with the inner getlogs, so since we always have a different representation on the rpc I think it's fine if we keep the bloom

crates/node/tests/it/private_rpc_e2e.rs
mattsse, line 1: kinda strange that we need this

mattsse (changes requested):

mattsse (approved):

PR #734, opened 2026-07-21
feat(node): load sequencer key from file

decofe:
## Summary
- add mutually exclusive `--sequencer-key-file` / `SEQUENCER_KEY_FILE` support
- allow regular files or FIFOs so sequencer keys do not need to appear in argv
- zeroize raw key input after constructing the signer
- document the new key source

## Testing
- `git diff --check`
- `cargo test -p zone-node --features cli cli::tests --no-fail-fast` *(blocked: Cargo dependency fetch for private Commonware revision returns HTTP 401 in sandbox)*

Prompted by: @kamsz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13534,6 +13534,7 @@ dependencies = [
  "tokio-tungstenite 0.28.0",
  "tracing",
  "url",
+ "zeroize",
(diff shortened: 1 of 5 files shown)


Zygimantass:
+1

PR #738, opened 2026-07-21
feat(rpc): add authenticated encryption key read

struong:
## Summary

Adds authenticated `zone_getEncryptionKey` support to the private Zone RPC.

The endpoint reads `encryptionKeyAtBlock` from the configured ZonePortal at the current Tempo L1 head and forwards its decoded `{ x, yParity, keyIndex }` result directly. This is the latest value clients need when constructing `depositEncrypted` transactions, including when L1 is ahead of the Zone.

## Changes

- Added the private RPC method, allowlist entry, and dispatch wiring.
- Pinned the portal read to the current L1 block and returned the generated contract response without a custom wrapper.
- Added focused coverage showing an L1 key rotation is returned without waiting for Zone synchronization.
- Documented the method and response shape.


crates/node/src/rpc.rs
```diff
@@ -319,6 +331,55 @@ impl<Api: EthApiTypes + 'static> ZoneRpc<Api> {
...
+        &self,
+    ) -> Result<EncryptionKeyResponse, JsonRpcError> {
+        let zone_head = self
+            .zone_provider
+            .get_block_number()
+            .await
+            .map_err(internal)?;
+        let zone_block = BlockId::number(zone_head);
+        let tempo_block_number_call = self.tempo_state.tempoBlockNumber().block(zone_block);
+        let tempo_block_hash_call = self.tempo_state.tempoBlockHash().block(zone_block);
```
0xKitsune: I dont think we need to fetch the block hash here?

crates/node/src/rpc.rs
```diff
@@ -319,6 +331,55 @@ impl<Api: EthApiTypes + 'static> ZoneRpc<Api> {
...
+        let (tempo_block_number, tempo_block_hash) =
+            tokio::try_join!(tempo_block_number_call.call(), tempo_block_hash_call.call(),)
+                .map_err(internal)?;
+        let portal_address = self.l1_state_provider.portal_address();
+
+        let (key, key_index) = self
+            .l1_state_provider
+            .encryption_key_at(BlockId::hash(tempo_block_hash))
```
0xKitsune: Could we check that this key is still valid at the current Tempo head before returning it? If the Zone falls behind during a key rotation and stays behind past the grace period, this will return the old key even though `depositEncrypted` will reject it. The existing test covers the grace period, but not what happens after the old key expires.

0xKitsune (dismissed):

0xKitsune:
Pending @mattsse

mattsse (approved):

PR #739, opened 2026-07-21
refactor(node): forward built payload execution data

mattsse:
## Summary

- consume `TempoBuiltPayload` directly when submitting through the consensus engine handle
- avoid cloning and manually rewrapping the built block
- preserve payload-carried execution data such as a future block access list

The zone payload builder already attaches `Some(executed_block)` to `TempoBuiltPayload`, allowing Reth's built-payload subscriber to insert the locally executed block through its engine-tree fast path.

## Validation

- `RUSTFLAGS='-A semicolon-in-expressions-from-macros' cargo check -p zone-node`

The lint override is required for an unrelated existing future-incompatibility lint in `crates/l1/src/state/tip403/provider.rs:490` on the current nightly toolchain.

cc @rjected

### DIFF
--- a/crates/node/src/engine.rs
+++ b/crates/node/src/engine.rs
@@ -44,7 +44,7 @@ use eyre::OptionExt;
 use reth_chainspec::EthereumHardforks;
 use reth_node_builder::ConsensusEngineHandle;
 use reth_payload_builder::PayloadBuilderHandle;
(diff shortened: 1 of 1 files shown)


mattsse:
@rjected would appreciate your review on this simplification.

Rjected (approved):

PR #744, opened 2026-07-21
feat(rpc): require enabled token balance

0xKitsune:
## Summary
- enforce a non-zero balance in at least one enabled zone token during transaction-pool admission
- seed a dedicated enabled-token registry from the startup L1 snapshot and update it from confirmed `TokenEnabled` events
- read that registry during pool admission, avoiding per-transaction portal-array decoding or L1 RPC
- return a descriptive pool rejection when the sender has no enabled-token balance
- document the check as admission policy, not consensus validity

## Testing
- `cargo check -p zone-node -p zone-l1`
- `cargo test -p zone-l1`
- `cargo test -p zone-node --lib`
- `cargo test -p zone-node --test it test_send_raw_transaction_requires_enabled_token_balance -- --nocapture`
- `cargo clippy -p zone-node -p zone-l1 --all-targets`


crates/node/src/rpc.rs
```diff
@@ -331,6 +333,22 @@ impl<Api: EthApiTypes + 'static> ZoneRpc<Api> {
...
+                .from(account)
+                .call()
+                .await
+                .map_err(internal)?;
+            if !balance.is_zero() {
+                return Ok(());
+            }
+        }
+
+        Err(JsonRpcError::transaction_rejected())
```
legion2002: should we have a more descriptive error here

0xalpharush (dismissed):

legion2002 (dismissed):

0xalpharush:
should we put it in executor?

0xalpharush:
Alternative implementation in #745:

The change centralizes the zone-specific validations at three boundaries:

1. **Stateless transaction validation, shared by the pool and EVM**
   - preserves the contract-deployer allowlist
   - parses direct calls and every call in an AA batch
   - permits direct TIP-20 calls only for `transferFrom` and `approve`
   - rejects malformed or unsupported TIP-20 calls before admission and during block re-execution

2. **Stateful transaction-pool admission**
   - tracks enabled tokens independently from TIP-403 policy entries
   - reads enabled-token balance slots from the latest local zone state
   - accepts a transaction only when its recovered sender has a non-zero balance in at least one enabled token
   - applies the rule to every pool ingress path without per-submission RPC calls

3. **Block-executor validation**
   - recognizes zone system transactions by their exact target and selector
   - requires exactly one `advanceTempo` transaction and requires it to be first
   - allows at most one `finalizeWithdrawalBatch` transaction and requires it to be last
   - rejects user transactions that impersonate state-changing system operations
   - allows `getPendingWithdrawals` for builder simulation but rejects it if committed to a block

The balance requirement remains pool-admission policy, while transaction-call restrictions and block sequencing are enforced during execution and re-execution.

klkvr (dismissed):

crates/node/src/node.rs
```diff
@@ -462,6 +476,9 @@ where
             .await?
             .erased();
 
+        self.resolve_and_seed_tokens(&l1_provider, tempo_block_number)
```
mattsse, line 484: **[P1] Handle the pre-portal genesis anchor without aborting startup.** A freshly provisioned zone stores the L1 block immediately *before* `createZone` as its `tempoBlockNumber`, specifically so the subscriber can replay the creation block. At that anchor the portal has no code, so `enabled_tokens_at(...)` returns no data and this `?` terminates node launch before replay begins. This is visible in CI as `failed to discover enabled tokens ... called address might not be a contract`.

Suggested fix: first check whether the portal has code at this exact anchor. If it does not, seed an empty registry and let the subscriber backfill `TokenEnabled` events from `anchor + 1`; if it does, keep the historical snapshot lookup. Preserve real RPC/decoding failures as errors rather than treating every failure as an empty registry. A regression test using `l1_anchored_genesis`/the normal provision flow should cover the pre-creation anchor.

mattsse (changes requested):
Requesting changes for one blocking startup regression. Fresh-zone genesis is intentionally anchored before `createZone`, so token discovery must handle the portal not existing at that historical anchor. The current call aborts launch before the L1 subscriber can replay the creation block and its initial `TokenEnabled` event. The failing CI shards reproduce this failure.

mattsse (changes requested):
Requesting changes for one blocking startup regression (the latest head still contains the issue described in the inline comment).

Fresh-zone genesis is intentionally anchored at the L1 block immediately before `createZone`, so the portal has no code at the stored `tempoBlockNumber`. `launch_add_ons` now unconditionally calls `enabled_tokens_at` at that historical block and propagates the no-data error, aborting node startup before the subscriber can replay the creation block and its initial `TokenEnabled` event. The failing CI shards reproduce this path.

Suggested fix: check whether the portal has code at the exact anchor. If it does not, seed an empty registry and let the subscriber backfill events from `anchor + 1`; if it does, keep the historical snapshot lookup. Continue surfacing genuine RPC and decoding failures. Please add a regression test using `l1_anchored_genesis`/the normal provision flow, and update the direct-queue test fixture so injected `TokenEnabled` events also reach the registry.

crates/node/src/node.rs
0xKitsune: Updated startup handling so it no longer fails when the portal has no code and there are no enabled tokens.

mattsse (approved):

PR #746, opened 2026-07-21
refactor(precompiles): centralize zone registration

0xrusowsky:
this PR aligns precompile registration with how upstream tempo does it


0xrusowsky:
cyclops audit fast

crates/precompiles/src/lib.rs
```diff
@@ -32,6 +32,22 @@
 
 extern crate alloc;
 
+macro_rules! zone_precompile {
```
legion2002, line 35: do we have to write this macro again? can't we just import it from upstream?

0xrusowsky: call rules is specific from zones, so we need a custom macro

klkvr (approved):

0xKitsune (approved):

PR #748, opened 2026-07-21
fix(zones): bound withdrawal delivery gas

decofe:
## Summary
- apply the withdrawal gas limit to the full `deliverWithdrawal` self-call
- remove the nested callback gas parameter from `ZoneMessenger`
- copy only the first four return-data bytes after successful callbacks and reject failed/short/wrong-selector responses
- document the delivery-level gas semantics and add oversized return/revert-data coverage

## Validation
- `forge fmt --check`
- `forge build --force`
- focused tests compile but cannot run under stock Forge in this pod because Tempo precompiles are unavailable (`MissingPrecompile("BlockHashHistory", ...)`)

Prompted by: @0xalpharush

### DIFF
--- a/specs/ref-impls/src/interfaces/IZone.sol
+++ b/specs/ref-impls/src/interfaces/IZone.sol
@@ -270,7 +270,7 @@ interface IAesGcmDecrypt {
 
 }
 
-// Maximum callback gas a withdrawal may request.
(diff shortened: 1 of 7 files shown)


legion2002:
cyclops audit fast

jenpaff:
superseded by #859

PR #752, opened 2026-07-21
feat(evm): `ZoneFeeManager`

0xKitsune:
This PR implements a Zone-native protocol fee manager.

Transactions can use a `TempoTransaction` to specify a valid fee token. If no fee token is specified, the fee manager defaults to the first token enabled on the Zone. The manager escrows the maximum fee before execution, refunds unused gas, and transfers the actual fee directly to the block beneficiary.

Any enabled token on the zone is accepted by the fee manager as a valid fee token.


crates/precompiles/src/zone_fee_manager.rs
```diff
@@ -0,0 +1,148 @@
...
+        fee_payer: Address,
+        fee_token: Address,
+        max_amount: U256,
+    ) -> Result<Address> {
+        let mut token = TIP20Token::from_address(fee_token)?;
+        if self.storage.spec().is_t8() {
+            token.ensure_authorized_as(&[(fee_payer, AuthRole::sender())])?;
+        } else {
+            token.ensure_transfer_authorized(fee_payer, self.address)?;
+        }
```
0xrusowsky: let's add a TODO to remove this once T8 activates?

crates/evm/src/lib.rs
```diff
@@ -113,9 +120,7 @@ where
         precompiles.set_precompile_lookup(move |address: &alloy_primitives::Address| {
             if is_tip20_prefix(*address) {
                 Some(create_tip20_precompile(*address, &env, sequencer.clone()))
-            } else if *address == TIP_FEE_MANAGER_ADDRESS {
-                Some(TipFeeManager::create_precompile(&tempo_env))
-            } else if *address == STABLECOIN_DEX_ADDRESS {
+            } else if *address == TIP_FEE_MANAGER_ADDRESS || *address == STABLECOIN_DEX_ADDRESS {
```
0xrusowsky: let's drop `TIP_FEE_MANAGER_ADDRESS` form here as we already define it above

crates/evm/src/fee_manager.rs
```diff
@@ -0,0 +1,129 @@
...
+        ctx: ProtocolFeeContext<'_, DB>,
+        fee_payer: Address,
+        fee_token: Address,
+        max_amount: U256,
+        _beneficiary: Address,
+        _skip_liquidity_check: bool,
+    ) -> Result<Address> {
+        if !self.l1_reader.is_fee_token_enabled(fee_token) {
+            return Err(FeeManagerError::invalid_token().into());
+        }
```
0xrusowsky: i don’t think we should use `L1StorageReader` directly in the handler path because it reads node-local cache state and bypasses `L1OverlayDB`.

i believe token enablement should instead be derived from l1-anchored zone state by checking whether it is initialized with something like:
 ```rust
   ctx.enter(|| {
       if !TIP20Token::from_address(fee_token)?.is_initialized()? {
           return Err(FeeManagerError::invalid_token().into());
       }

       ZoneFeeManager::new().collect_fee_pre_tx(
           fee_payer,
           fee_token,
           max_amount,
       )
   })
 ```

crates/evm/src/fee_manager.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl<L1: L1StorageReader> ZoneProtocolFeeManager<L1> {
+    pub(crate) const fn new(l1_reader: L1) -> Self {
+        Self { l1_reader }
+    }
+
+    fn resolve_fee_token(&self, tx: &TempoTxEnv) -> Result<Address> {
+        let Some(token) = tx
+            .fee_token()
+            .or_else(|| self.l1_reader.default_fee_token())
```
0xrusowsky: sorry, but i think i wasn't clear enough in the prev cmnt. Unless we totally we the `l1_reader: L1` dependency for `ZoneProtocolFeeManager`, have the same problem as before here.

instead i guess we have to store `default_fee_token` in `ZoneFeeManager` state?

0xrusowsky (dismissed):
LGTM after all the fixes.

let's wait for @klkvr and/or cyclops review before merging

crates/evm/src/fee_manager.rs
```diff
@@ -0,0 +1,83 @@
...
+        fee_payer: Address,
+        fee_token: Address,
+        max_amount: U256,
+        _beneficiary: Address,
+        _skip_liquidity_check: bool,
+    ) -> Result<Address> {
+        ctx.enter(|| {
+            if !TIP20Token::from_address(fee_token)?.is_initialized()? {
+                return Err(FeeManagerError::invalid_token().into());
+            }
```
klkvr: this seems redunant because tempo's handler would validate fee token internally

crates/precompiles/src/zone_fee_manager.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// Returns the canonical default token used when a transaction omits `fee_token`.
+    pub fn default_fee_token(&self) -> Result<Address> {
+        self.default_fee_token.read()
+    }
+
+    /// Returns fees accrued to `beneficiary` in `token`.
+    pub fn collected_fees(&self, beneficiary: Address, token: Address) -> Result<U256> {
+        self.collected_fees[beneficiary][token].read()
+    }
```
klkvr: should we enforce any restrictions on caller here for privacy reasons?

crates/precompiles/src/tempo_state.rs
```diff
@@ -37,6 +37,11 @@ pub struct TempoState {
 pub const TEMPO_BLOCK_NUMBER_SLOT: alloy_primitives::U256 = slots::TEMPO_BLOCK_NUMBER;
 
 impl TempoState {
+    /// Returns the finalized Tempo block used for anchored L1 reads.
+    pub fn finalized_block_number(&self) -> tempo_precompiles::Result<u64> {
+        self.tempo_block_number.read()
+    }
```
klkvr: is this unrelated to PR?

crates/evm/src/fee_manager.rs
0xKitsune: removed in https://github.com/tempoxyz/zones/pull/752/changes/c4bc04db305a3880542832e16e6dfd76dcda76b4

crates/precompiles/src/tempo_state.rs
0xKitsune: removed in https://github.com/tempoxyz/zones/commit/c4bc04db305a3880542832e16e6dfd76dcda76b4

crates/precompiles/src/zone_fee_manager.rs
0xKitsune: Yeah we can enforce this, updated in https://github.com/tempoxyz/zones/pull/752/changes/64268736d85bbb37913bab2f02c5fb3193d0321b

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+    tip403_registry::AuthRole,
+    view,
+};
+use tempo_precompiles_macros::contract;
+use tempo_zone_contracts::IZoneFeeManager;
+
+/// Zone-owned fee balances at Tempo's canonical fee-manager address.
+#[contract(addr = TIP_FEE_MANAGER_ADDRESS)]
+pub struct ZoneFeeManager {
+    collected_fees: Mapping<Address, Mapping<Address, U256>>,
```
0xKitsune: This is a non-issue, there are no pre-existing fee manager fees on the zone.

crates/evm/src/fee_manager.rs
```diff
@@ -0,0 +1,83 @@
...
+        tx: &TempoTxEnv,
+        _fee_payer: Address,
+        spec: TempoHardfork,
+        actions: StorageActions,
+    ) -> Result<Address> {
+        if let Some(token) = tx.fee_token() {
+            return Ok(token);
+        }
+
+        journal
```
0xKitsune: Updated in https://github.com/tempoxyz/zones/pull/752/changes/0217870de2cd80587a82c04498f76a16f76e9dd8

crates/node/src/genesis.rs
```diff
@@ -20,6 +20,8 @@ const TEMPO_STATE_ADDRESS: Address = address!("0x1c00000000000000000000000000000
 const ZONE_INBOX_ADDRESS: Address = address!("0x1c00000000000000000000000000000000000001");
 /// ZoneConfig predeploy address.
 const ZONE_CONFIG_ADDRESS: Address = address!("0x1c00000000000000000000000000000000000003");
+/// ZoneFeeManager precompile address.
+const ZONE_FEE_MANAGER_ADDRESS: Address = address!("0xfeec000000000000000000000000000000000000");
```
legion2002: since this is not the same logic as the L1 fee manager think we should choose a different address for this to avoid confusion

crates/evm/src/fee_manager.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub(crate) const fn new() -> Self {
+        Self
+    }
+}
+
+impl<DB> ProtocolFeeManager<DB> for ZoneProtocolFeeManager
+where
+    DB: Database,
+{
+    fn get_fee_token(
```
legion2002, line 52: do we need this? we already have resolve_fee_token and doesn't seem like this is being used anywhere

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+        if !actual_spending.is_zero() {
+            self.collected_fees[beneficiary][fee_token].sinc(actual_spending)?;
+        }
+        Ok(actual_spending)
+    }
+
+    /// Pays a beneficiary's accrued balance.
+    ///
+    /// External dispatch restricts this action to the beneficiary.
+    pub fn distribute_fees(&mut self, beneficiary: Address, token: Address) -> Result<()> {
```
legion2002: will we ever have multiple beneficiaries, my understanding is that the fee always goes to a single entity. In a single sequencer model, it goes to  the sequencer key, in a multi sequencer model it goes to the admin.

So maybe we should remove beneficiary param from everywhere? (can also keep it for more flexibility, don't feel strongly about it)

crates/precompiles/src/zone_fee_manager.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// Escrows the maximum fee directly in the selected token.
+    pub fn collect_fee_pre_tx(
+        &mut self,
+        fee_payer: Address,
+        fee_token: Address,
+        max_amount: U256,
+    ) -> Result<Address> {
+        let mut token = TIP20Token::from_address(fee_token)?;
+        // TODO: Remove the pre-T8 authorization path once T8 activates.
+        if self.storage.spec().is_t8() {
```
legion2002: we can just assume T8 is activated in all the environments zones is working on, since it depends on T9

crates/node/src/eth_api.rs
```diff
@@ -0,0 +1,381 @@
...
+use reth_transaction_pool::{PoolTx, TransactionOrigin};
+use tempo_chainspec::hardfork::TempoHardfork;
+use tempo_node::rpc::{TempoEthApi, TempoEthApiBounds, TempoEthApiBuilder};
+use tempo_precompiles::storage::StorageActions;
+use tempo_primitives::TEMPO_GAS_PRICE_SCALING_FACTOR;
+use tempo_revm::TempoStateAccess;
+use tokio::sync::Mutex;
+use zone_evm::resolve_fee_token;
+
+fn fee_token_gas_allowance<DB>(
```
legion2002: Can we add this trait upstream in tempo, and just get rid of this boiler plate code?
```  
trait FeeTokenGasAllowance {
      fn caller_gas_allowance<DB>(
          &self,
          db: &mut DB,
          tx: &TempoTxEnv,
          spec: TempoHardfork,
      ) -> Result<u64, ProviderError>
      where
          DB: Database;
  }```

File: crates/evm/src/lib.rs [hunk 15]
```diff
@@ -104,6 +107,10 @@ where
         precompiles.apply_precompile(&ZONE_TIP20_FACTORY_ADDRESS, |_| {
             Some(ZoneTokenFactory::create(&env))
         });
+        let fee_env = env.clone();
+        precompiles.apply_precompile(&TIP_FEE_MANAGER_ADDRESS, move |_| {
```
0xKitsune: Addressed in https://github.com/tempoxyz/zones/pull/752/changes/09dcf57db0aea5b2367ce9e28ee5c2073628630e

crates/precompiles/src/zone_fee_manager/mod.rs
0xKitsune: I think we need this in the case where there are multiple sequencers. If one gets promoted to the leader, then the beneficiary could change. We could also have them all specify the same beneficiary but this seems reasonable to keep as is. Lmk what you think.

crates/node/src/eth_api.rs
0xKitsune: Yes lets add this in a followup

crates/precompiles/src/zone_fee_manager.rs
0xKitsune: Updated in [35f5ec5](https://github.com/tempoxyz/zones/pull/752/commits/35f5ec55605dbf60a0b8ee7a74c9617c003676e7)

crates/node/src/genesis.rs
0xKitsune: Updated in [35f5ec5](https://github.com/tempoxyz/zones/pull/752/commits/35f5ec55605dbf60a0b8ee7a74c9617c003676e7)

crates/evm/src/fee_manager.rs
0xKitsune: This is used by Tempo’s transaction handler via `ProtocolFeeManager::get_fee_token` (`tempo-revm/src/handler.rs`). If we remove the override, the trait default calls `journal.get_fee_token(...)`, which resolves omitted fee tokens through Tempo’s L1 `TipFeeManager` at `0xfeec…0000` instead of the ZoneFeeManager default at `0xfeec…0001`. I added a comment at the override to make that call path explicit.

crates/l1/src/state/cache.rs
0xrusowsky, line 1: let's drop all the cache changes, since `enabled_tokens` are not needed because we perform the check via the L2 bytecode check on the tip20

crates/l1/src/subscriber.rs
0xrusowsky, line 1: same here, let's drop the changes related to the cache

crates/evm/src/zone_evm/mod.rs
```diff
@@ -91,27 +88,6 @@ where
     }
```
0xrusowsky, line 91: i think this trait shouldn't be dropped

crates/l1/src/state/cache.rs
0xKitsune: Removed the enabled-token fields, API, reset behavior, and cache test in fd342306. Fee-token validity now relies on the initialized L2 TIP-20 state.

crates/l1/src/subscriber.rs
0xKitsune: Removed the TokenEnabled-driven subscriber cache updates in fd342306, along with startup priming and fixture cache seeding for the obsolete enabled-token cache.

crates/evm/src/zone_evm/mod.rs
0xKitsune: Yes, will update this.

crates/node/src/eth_api.rs
0xrusowsky, line 1: i think we can avoid this by using https://github.com/tempoxyz/tempo/pull/6928

crates/contracts/src/precompiles/zone_fee_manager.rs
0xrusowsky, line 1: this is fine, but noting that we could use the upstream interface directly, and simply revert with unsuported selector on the l1-specific selectors.

it would allow us to avoid drift if we ever do breaking api changes

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+            )?;
+        }
+
+        token.decrement_balance(fee_payer, max_amount)?;
+        token.increment_balance(self.address, max_amount)?;
+        Ok(fee_token)
+    }
+
+    /// Refunds unused gas and credits the net fee to the block beneficiary.
+    pub fn collect_fee_post_tx(
```
0xrusowsky, line 53: can we document why we do manual tip20 ledger manipulation and even emission rather than transfering?

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -0,0 +1,247 @@
...
+        if !reward_recipient.is_zero() {
+            let opted_in_supply = U256::from(token.get_opted_in_supply()?)
+                .checked_sub(max_amount)
+                .ok_or_else(TempoPrecompileError::under_overflow)?;
+            token.set_opted_in_supply(
+                opted_in_supply
+                    .try_into()
+                    .map_err(|_| TempoPrecompileError::under_overflow())?,
+            )?;
+        }
```
0xKitsune: We should be able to remove this completely?

crates/contracts/src/precompiles/zone_fee_manager.rs
0xKitsune: Updated to use the upstream interface.

crates/precompiles/src/zone_fee_manager/mod.rs
0xKitsune: This was updated to use _transfer rather than increment/decrement directly.

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+                &Recipient::direct(beneficiary),
+                actual_spending,
+            )?;
+
+            StorageCtx.emit_event(
+                fee_token,
+                TIP20Event::transfer(fee_payer, self.address, actual_spending).into_log_data(),
+            )?;
+        }
+
```
0xrusowsky, line 80: `fn _transfer()` already emits the event, so we can remove this
```suggestion
```

0xrusowsky: NOTE: we should modify tests to make sure i am not wrong

0xKitsune: Updated to simplify the post tx hook in [585b575](https://github.com/tempoxyz/zones/pull/752/commits/585b575f1173447446f26e8ec3ec66cb45cb18f7)

crates/precompiles/src/zone_fee_manager/dispatch.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use super::ZoneFeeManager;
+
+impl TempoPrecompile for ZoneFeeManager {
+    fn call(&mut self, calldata: &[u8], _sender: Address) -> PrecompileResult {
+        if let Some(error) = charge_input_cost(&mut self.storage, calldata) {
+            return error;
+        }
+
+        unknown_selector_result(calldata)
```
0xrusowsky, line 22: can we add a short comment explaining that the zone's fee manager precompile doesn't accept external calls?

0xrusowsky (commented):
LGTM now, let's fix CI and do a final cyclops run

crates/precompiles/src/zone_fee_manager/dispatch.rs
0xKitsune: Addressed in [3e1616e](https://github.com/tempoxyz/zones/pull/752/commits/3e1616e46be2dd9eba90be33e503bb35208ef28f)

mattsse (commented):
## Findings

### [P1] A supported non-USD initial token creates an unusable default fee configuration

[`CreateZone` copies `initial_token` directly into `default_fee_token`](https://github.com/tempoxyz/zones/blob/3e1616e46be2dd9eba90be33e503bb35208ef28f/xtask/src/create_zone.rs#L149-L156). However, the pinned upstream [`ZoneFactory::create_zone` validation only checks that the initial token is an initialized TIP-20](https://github.com/tempoxyz/tempo/blob/67dd20428738dee30081043ff1b3e55897d5e1b5/crates/precompiles/src/zone_factory/mod.rs#L180-L191), while the transaction handler [requires every non-free fee token to have `USD` currency](https://github.com/tempoxyz/tempo/blob/67dd20428738dee30081043ff1b3e55897d5e1b5/crates/revm/src/handler.rs#L982-L992).

This means zone creation can successfully accept an EUR or other non-USD initial token, persist it as the immutable default, and then reject every ordinary transaction that omits `fee_token` with `FeeTokenNotUsdCurrency`. Explicitly selecting that token fails for the same reason, so this also contradicts the PR contract that any enabled Zone token can pay fees.

**Suggested fix:** align creation and execution around one invariant. If Zones are intentionally USD-only, validate the initial/default token currency before creating the portal or genesis and reject unsupported configurations atomically. If any enabled token is intended, make the upstream currency validation injectable so `ZoneEvmConfig` can replace the USD-only check with Zone-specific enabled-token validation. In either case, add an integration test creating a Zone with a non-USD initial token and asserting the intended creation or fee-payment behavior. This needs to be enforced at the protocol boundary rather than only in the CLI, because portals can be created through other callers.

### [P2] The maximum fee is spendable during transaction execution instead of being escrowed

[`collect_fee_pre_tx` credits the full maximum fee directly to the beneficiary](https://github.com/tempoxyz/zones/blob/3e1616e46be2dd9eba90be33e503bb35208ef28f/crates/precompiles/src/zone_fee_manager/mod.rs#L35-L49). The beneficiary balance is therefore live while the transaction body executes. If the beneficiary is a contract, or has granted an allowance used by transaction code, that provisional balance can be moved before [`collect_fee_post_tx` tries to debit the refund from the beneficiary](https://github.com/tempoxyz/zones/blob/3e1616e46be2dd9eba90be33e503bb35208ef28f/crates/precompiles/src/zone_fee_manager/mod.rs#L52-L75). Spending more than the final fee makes the refund debit underflow and invalidates settlement; spending up to the final fee can redirect the beneficiary revenue.

**Suggested fix:** use the Zone fee-manager account as transaction-scoped escrow. In the pre hook, debit the payer and credit `self.address`, leaving the beneficiary unchanged. In the post hook, debit `self.address` for the refund and credit the payer, then debit the remaining actual fee and credit the beneficiary before emitting the net payer-to-beneficiary event. The escrow is isolated because [the Zone fee-manager precompile rejects all external selectors](https://github.com/tempoxyz/zones/blob/3e1616e46be2dd9eba90be33e503bb35208ef28f/crates/precompiles/src/zone_fee_manager/dispatch.rs#L11-L19).

Please add a regression test that checks: after the pre hook the manager owns `max_amount` and the beneficiary balance is unchanged; arbitrary beneficiary balance movement between hooks cannot affect the refund; and after the post hook the manager balance is zero, the payer received the refund, and the beneficiary received only `actual_spending`. This preserves the stated escrow invariant and removes the refund path's dependency on beneficiary behavior.

crates/node/src/eth_api.rs
0xKitsune: addressed in a9bd6137 — rpc now uses the upstream FeeTokenResolver path from tempo#6928, and the local eth_api shim is gone.

crates/evm/src/lib.rs
```diff
@@ -125,8 +136,6 @@ where
         precompiles.set_precompile_lookup(move |address: &alloy_primitives::Address| {
             if is_tip20_prefix(*address) {
                 Some(create_tip20_precompile(*address, &env, tip20_l1.clone()))
```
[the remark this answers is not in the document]
0xKitsune: addressed in 304c0732 — keeping upstream TIP-1060 behavior unchanged. the fix only reuses evm.non_creditable_slots() across every Zone and upstream precompile environment.

crates/precompiles/src/zone_fee_manager/dispatch.rs
```diff
@@ -0,0 +1,20 @@
...
+use super::ZoneFeeManager;
+
+impl TempoPrecompile for ZoneFeeManager {
+    fn call(&mut self, calldata: &[u8], _sender: Address) -> PrecompileResult {
+        if let Some(error) = charge_input_cost(&mut self.storage, calldata) {
+            return error;
+        }
+
+        // Zone fee settlement is protocol-only, this precompile does not accept external calls.
+        unknown_selector_result(calldata)
```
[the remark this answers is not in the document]
0xKitsune: addressed in 2488bffa — 0–3 byte calldata now takes missing_selector_result() and has regression coverage for each length.

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        max_amount: U256,
+        beneficiary: Address,
+    ) -> Result<Address> {
+        let mut token = TIP20Token::from_address(fee_token)?;
+        token.ensure_authorized_as(&[(fee_payer, AuthRole::sender())])?;
+        token.check_not_paused()?;
+        token.check_and_update_spending_limit(fee_payer, max_amount)?;
+
+        token.decrement_balance(fee_payer, max_amount)?;
+        token.increment_balance(beneficiary, max_amount)?;
```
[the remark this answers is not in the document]
0xKitsune: addressed in 6da1c256 — max fee now sits in ZoneFeeManager escrow, refunds come from escrow, and only actual spend reaches the authorized beneficiary after execution. zones do not use rewards, so there is no rewards path here.

zone-genesis/genesis.json
```diff
@@ -64,11 +64,11 @@
     "0xc0709e413c2f926223aa9aad128d5a068bcbfe21": {
       "balance": "0x3635c9adc5dea00000"
     },
-    "0xfeec000000000000000000000000000000000000": {
+    "0xfeec000000000000000000000000000000000001": {
```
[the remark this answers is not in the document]
0xKitsune: addressed in 0989a1cf — regenerated the root artifact with the current tool, so the manager marker and default fee token are initialized now.

0xKitsune:
went through matt + cyclops one by one. current state:

- non-usd fee tokens: addressed in c3976cb8 (depends on https://github.com/tempoxyz/tempo/pull/6938), tempo stays usd-only by default, zones accepts any initialized TIP-20 through the same execution/txpool/rpc/replay path.

- fee escrow: addressed in 6da1c256, max fee stays in ZoneFeeManager until post-execution, refund comes from escrow, and only actual spend reaches the authorized beneficiary. no rewards path on zones.

- storage credits: addressed in 304c0732,  zones keeps upstream TIP-1060 behavior unchanged and reuses the authoritative non-creditable-slot guard across every precompile environment.

- short calldata: addressed in 2488bffa,  0–3 byte inputs now deterministically empty-revert.

- root genesis: addressed in 0989a1cf, regenerated, with ZoneFeeManager + default token initialized.

- pre-T1C behavior: no change,  zones start at T9, so there is no supported pre-T1C history to replay.

- upgrade compatibility: no change, these interfaces and storage layouts are still intentionally breaking; migration support starts once the boundary is stable.

mattsse (dismissed):
Approved for the current operating model: protocol-enforced zero fees and a trusted first-party sequencer.

Informational follow-ups if those assumptions change:

- Zone fee settlement creates the beneficiary TIP-20 balance while TIP-1060 accounting is disabled. This is inactive when `actual_spending == 0` and is not an untrusted path with the current sequencer model. Before enabling nonzero fees or third-party sequencers, ensure an exempt balance creation cannot later mint an unbacked storage credit when cleared using another fee token.
- Pool ordering compares raw nominal tips across fee tokens. This is inactive when effective tips are enforced at zero. Before enabling nonzero priority fees or economically different fee tokens, use one priority denomination, normalize values, or isolate scheduling by token.

Neither item blocks the stated zero-fee, trusted-sequencer deployment.

0xrusowsky (approved):

0xrusowsky (dismissed):

0xrusowsky (approved):

PR #753, opened 2026-07-21
feat: `ZoneOutbox` precompile

klkvr:
Introduces native `ZoneOutbox` precompile


klkvr:
cyclops audit

crates/primitives/src/constants.rs
```diff
@@ -59,18 +59,13 @@ pub const ZONE_TOKEN_ADDRESS: Address = address!("0x20C0000000000000000000000000
 pub const PORTAL_SEQUENCER_SLOT: B256 = B256::ZERO;
 
 /// ZonePortal storage slot 1: `admin` (address).
-pub const PORTAL_ADMIN_SLOT: B256 = {
-    let mut bytes = [0u8; 32];
-    bytes[31] = 1;
-    B256::new(bytes)
-};
+pub const PORTAL_ADMIN_SLOT: B256 = B256::with_last_byte(1);
```
legion2002: We can export these slots from Tempo [PR](https://github.com/tempoxyz/tempo/pull/6874) which uses the #[contract] macro instead of hardcoding them here

crates/contracts/src/precompiles/outbox.rs
```diff
@@ -84,3 +106,34 @@ crate::sol! {
         function finalizeWithdrawalBatch(uint256 count, uint64 blockNumber, bytes[] calldata encryptedSenders) external returns (bytes32 withdrawalQueueHash);
     }
 }
+
+crate::sol! {
+    /// Legacy seven-argument withdrawal interface. The current overload adds `revealTo`.
+    #[derive(Debug)]
+    interface ILegacyZoneOutbox {
```
legion2002: why do we need this? should we just get rid of the legacy stuff?

crates/contracts/src/precompiles/zone_portal.rs
```diff
@@ -388,6 +391,11 @@ impl Withdrawal {
         }
     }
 
+    /// Hash this withdrawal as one link in a withdrawal queue.
+    pub fn hash_with_tail(&self, tail: B256) -> B256 {
```
legion2002, line 386: is this only used in `queue_hash` ? can be inlined if yes

crates/evm/src/executor.rs
```diff
@@ -112,7 +112,9 @@ where
         // transaction's resolved fee token, so the handler skips FeeAMM.
         self.override_validator_token();
 
-        let _tx_hash_guard = tx_context::set_current_tx_hash(*recovered.tx().tx_hash());
+        let fee_payer = tx_env.fee_payer().unwrap_or(tx_env.caller());
+        let _tx_context_guard =
+            tx_context::set_current_transaction(*recovered.tx().tx_hash(), fee_payer);
```
legion2002: nit
```suggestion
        let _tx_context_guard =
            tx_context::set_current_transaction(*recovered.tx().tx_hash(), tx_env.fee_payer().unwrap_or(tx_env.caller()));
```

crates/evm/src/lib.rs
```diff
@@ -95,6 +97,16 @@ where
             Some(TempoState::create(l1.clone(), &env))
         });
         precompiles.apply_precompile(&ZONE_TX_CONTEXT_ADDRESS, |_| Some(ZoneTxContext::create()));
+        let outbox_env = env.clone();
+        let outbox_l1 = l1.clone();
+        let outbox_portal = self.portal_address;
+        precompiles.apply_precompile(&ZONE_OUTBOX_ADDRESS, move |_| {
```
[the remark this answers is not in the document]
legion2002: invalid

crates/precompiles/src/outbox/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+        Ok(l1.read_l1_storage(portal_address, slot, anchor)?.into())
+    }
+
+    fn ensure_sequencer<P: L1StorageReader>(
+        &self,
+        l1: &L1State<P>,
+        portal_address: Address,
+        caller: Address,
+    ) -> ZoneResult<()> {
+        if caller == Address::ZERO {
```
legion2002, line 69: is there a reason we allow `caller == 0`? think the cyclops finding could be valid, so we should just allowe caller == `sequencer`?

crates/contracts/src/precompiles/zone_portal.rs
0xrusowsky: this will also be used by inbox

crates/precompiles/src/outbox/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+        l1: &L1State<P>,
+        portal_address: Address,
+        token: Address,
+    ) -> ZoneResult<()> {
+        let value = self.read_portal_slot(l1, portal_address, portal_token_config_slot(token))?;
+        if value.byte(0) == 0 {
+            return Err(ZonePortalError::token_not_enabled().into());
+        }
+        Ok(())
+    }
```
0xrusowsky: i think this fn is unsound and instead it should check whether the tip20 is initialized on the L2. 

`advanceTempo` makes L2 state canonical for execution, whereas direct L1 reads bypass the db overlay anchor and can introduce non-deterministism.

0xrusowsky: proposed fix https://github.com/tempoxyz/zones/pull/781

crates/precompiles/src/outbox/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+        if call.fallbackRecipient.is_zero() {
+            return Err(ZoneOutboxError::invalid_fallback_recipient().into());
+        }
+        if call.data.len() > MAX_CALLBACK_DATA_SIZE {
+            return Err(ZoneOutboxError::callback_data_too_large().into());
+        }
+
+        validate_gas_limit(call.gasLimit)?;
+
+        self.ensure_token_enabled(l1, portal_address, call.token)?;
```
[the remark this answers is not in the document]
klkvr: fixed by https://github.com/tempoxyz/zones/pull/781

crates/precompiles/src/outbox/mod.rs
klkvr: no reason i think, this is just how `ZoneOutbox` solidity impl worked

re cyclops finding i don't think on private RPC we allow to set `from` to arbitrary addresses, including `Address::ZERO` so it should be fine

tho it does look like we don't entirely disable the regular RPC which might be a separate problem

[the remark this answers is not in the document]
klkvr: if normal RPC is exposed we have a different issue because all privacy preserving methods can be bypassed through it

crates/contracts/src/precompiles/outbox.rs
klkvr: removed here 08d4e153

crates/primitives/src/constants.rs
klkvr: opened https://github.com/tempoxyz/tempo/pull/6923, can do in a follow up once merged

crates/precompiles/src/outbox/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    fn finalize_withdrawal_batch<P: L1StorageReader>(
+        &mut self,
+        l1: &L1State<P>,
+        portal_address: Address,
+        caller: Address,
+        call: IZoneOutbox::finalizeWithdrawalBatchCall,
+    ) -> ZoneResult<B256> {
+        self.ensure_sequencer(l1, portal_address, caller)?;
```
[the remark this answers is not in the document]
klkvr: already tracked as ZONE-73

crates/precompiles/src/outbox/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+    fn read_portal_slot<P: L1StorageReader>(
+        &self,
+        l1: &L1State<P>,
+        portal_address: Address,
+        slot: B256,
+    ) -> ZoneResult<U256> {
+        let anchor = StorageCtx::default().sload(TEMPO_STATE_ADDRESS, TEMPO_BLOCK_NUMBER_SLOT)?;
+        let anchor = u64::try_from(anchor).map_err(|_| {
+            TempoPrecompileError::Fatal("Tempo block number does not fit in u64".into())
+        })?;
```
0xrusowsky: nit: can we do use the precompile api instead?
```suggestion
        let anchor = TempoState::new().tempo_block_number.read()?;;
```

crates/precompiles/src/outbox/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+    fn ensure_sequencer<P: L1StorageReader>(
+        &self,
+        l1: &L1State<P>,
+        portal_address: Address,
+        caller: Address,
+    ) -> ZoneResult<()> {
+        if caller == Address::ZERO {
+            return Ok(());
+        }
+        let membership_slot = keccak256((caller, PORTAL_IS_SEQUENCER_SLOT).abi_encode());
```
0xrusowsky:
```suggestion
        let membership_slot = caller.mapping_slot(PORTAL_IS_SEQUENCER_SLOT.into());
```

0xrusowsky (commented):
as a general question, do u think it is worth limiting the precompile impl to "pure" evm execution + limit `L1StorageReader` to a wrapper type?

this way we would do dispatching on the wrapper, verify the caller is the sequencer when needed, and fwd regular "pure" execution to the precompile.

imo this would help us prevent unintended access to non-anchored L1 data, thus making it more difficult to introduce "data access" bugs, but maybe it adds too much complexity for little benefit

crates/precompiles/src/outbox/mod.rs
legion2002: okay should we just remove the 0 check then, just want to avoid any simulation or null values footguns in the future.

legion2002 (approved):
lgtm

0xrusowsky (approved):

PR #754, opened 2026-07-21
fix: support TIP-1091 single-sequencer settlement

legion2002:
## Summary

- align the Zones contracts, bindings, storage layout, generated genesis, and docs with the current TIP-1091 ZoneFactory and ZonePortal runtime
- create the required EIP-712 settlement certificate locally and submit it through the new quorum-aware submitBatch ABI
- pin the matching Tempo revision and update TIP-1092 fixtures for the current registry-first policy model

## Compatibility scope

This is the minimal bridge to the new on-chain interface: it supports an operational 1-of-1 sequencer set using the existing local sequencer key. It deliberately rejects thresholds other than 1.

Remote signature exchange, certificate aggregation, and multi-node quorum coordination remain in the full multi-sequencer PR. Keeping those out of this change lets the TIP-1091 contract/runtime compatibility land first without weakening the certificate check.

## Validation

- cargo nextest run --profile ci --no-fail-fast --locked (468 passed, 6 skipped)
- cargo check --workspace --all-targets --locked
- cargo test -p zone-sequencer --no-run --locked
- real-L1 deposit and signed batch settlement E2E
- ZonePortal storage layout matches the checked-in artifact
- Rust formatting and diff checks


legion2002:
Removed the legacy 7-argument `submitBatch` path and the `sequencerSetVersion == 0` mode switch. The nonce is still part of the settlement certificate for replay protection, but now starts at `0` for the initial valid configuration and increments only when the sequencer set or threshold changes. Every portal therefore uses the certificate-bearing `submitBatch` path from creation.

The matching Tempo factory initialization change is in tempoxyz/tempo#6914.

crates/sequencer/src/settlement.rs
```diff
@@ -430,6 +480,91 @@ impl BatchSubmitter {
...
+            .portal
+            .withdrawalBatchIndex()
+            .call()
+            .await?
+            .checked_add(1)
+            .ok_or_else(|| eyre::eyre!("portal withdrawal batch index overflow"))?;
+        let verifier = self.portal.verifier().call().await?;
+        let chain_id = self.l1_provider.get_chain_id().await?;
+
+        let domain_typehash = keccak256(
```
decofe: Alloy can generate this EIP-712 encoding and avoid keeping the Rust type string in sync manually. Define the signed type with `sol!`, then use `eip712_domain!` and `SolStruct::eip712_signing_hash`:

```rust
use alloy_sol_types::{SolStruct, eip712_domain, sol};

sol! {
    struct SettlementAttestation {
        uint32 zoneId;
        uint64 sequencerSetVersion;
        uint256 zoneHeight;
        uint256 withdrawalBatchIndex;
        address verifier;
        uint64 tempoBlockNumber;
        uint64 anchorBlockNumber;
        bytes32 anchorBlockHash;
        bytes32 blockTransitionHash;
        bytes32 depositQueueTransitionHash;
        bytes32 withdrawalQueueHash;
        bytes32 verifierConfigHash;
    }
}

let domain = eip712_domain! {
    name: "ZonePortal",
    version: "1",
    chain_id,
    verifying_contract: self.portal_address,
};
let digest = message.eip712_signing_hash(&domain);
```

`message` is the generated `SettlementAttestation` populated with the existing values/hashes. This replaces both manual type hashes, `domain_separator`, `struct_hash`, and the final `\x19\x01` concatenation.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -900,30 +1003,51 @@ contract ZonePortal is IZonePortal {
...
+            depositQueueTransition,
+            withdrawalQueueHash,
+            verifierConfig,
+            proof,
+            nextZoneHeight,
+            signatures
+        );
+    }
+
+    function _submitBatch(
```
adityapk00: Now that we don't have 2 submitBatch, we can remove this one and inline it into the submitBatch?

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -872,6 +963,18 @@ contract ZonePortal is IZonePortal {
         }
     }
```
0xKitsune, line 925: Do we need this?

legion2002:
Updated this PR to the new portal storage and pinned Tempo PR #6914. We removed the local `ZoneFactory.sol` and the standalone sequencer representative because the native TIP-1091 factory is now canonical and every active sequencer must have equal membership-based authority. The generated genesis and storage layout were regenerated against that exact Tempo revision.

crates/node/src/dev.rs
```diff
@@ -82,11 +82,9 @@ pub async fn provision_zone(config: ProvisionConfig) -> eyre::Result<Provisioned
             "ZoneFactory must use TIP-1091 address {ZONE_FACTORY_ADDRESS}, got {address}"
         );
     }
-    let factory_address = deploy_zone_factory(&l1_rpc_url, wallet).await?;
+    let factory_address = native_zone_factory(&l1_rpc_url, wallet).await?;
```
0xKitsune, line 85: How is this expected to work against a stock T9 dev chain? `createZone` is owner-only, but this sends it from `dev_key` (the default is `0xf39f...`) while the native factory is initialized with `0xaF57...` as owner. 

The integration fixture sets the owner to the dev signer and preinstalls all three runtimes, so it does not cover that setup. Unless Anvil or native dev mode has another bootstrap step, `tempo-zone dev` will revert with `NotOwner` before provisioning. Can we either bootstrap the factory here or validate and document the required setup?

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -210,29 +218,81 @@ contract ZonePortal is IZonePortal {
...
+        if (length == 0 || length > MAX_SEQUENCERS || newThreshold == 0 || newThreshold > length) {
+            revert InvalidSequencerSet();
+        }
+
+        for (uint256 i = 0; i < length; ++i) {
+            address signer = newSequencers[i];
+            if (signer == address(0)) revert InvalidSequencerSet();
+
+            if (rejectUnchanged) {
+                if (i > 0 && signer <= newSequencers[i - 1]) revert InvalidSequencerSet();
```
0xKitsune: TIP-1091 says sequencer ordering has no protocol meaning and valid sets must not be rejected based on order. This rejects every unsorted admin rotation, even though creation accepts the same set and certificate verification only checks membership.

 It also treats the same members in a different order as a new configuration. Can we use the same nonzero and uniqueness validation for rotations, and compare membership without relying on array order?

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: removed here - https://github.com/tempoxyz/zones/pull/754/changes/91a807f4d42878ca7c205c5cc0910acd801c1302
we pay the fee directly to the admin now

crates/sequencer/src/settlement.rs
legion2002: Done in 49c87faf. The Rust signer now defines `SettlementAttestation` with `sol!` and uses `eip712_domain!` plus `SolStruct::eip712_signing_hash`, removing the manually synchronized type hashes, domain separator, struct hash, and digest assembly.

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: Done in 49c87faf. Rotations now accept sequencers in any order, validate uniqueness independently of ordering, and compare unchanged configurations by membership rather than array position. I also updated the spec/interface wording and added coverage for unordered rotation plus reordered unchanged membership.

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: Done in 49c87faf. With the legacy overload gone, `submitBatch` now contains the implementation directly and the redundant `_submitBatch` wrapper has been removed.

crates/node/src/dev.rs
legion2002: we're going to ignore these tests for now and fix it in a separate PR with a key in the CI

legion2002 (commented):
:eye: **Cyclops Review**

This PR migrates zones to the native TIP-1091 factory + M-of-N sequencer-set / quorum-certificate settlement model (deletes the Solidity `ZoneFactory`, replaces the single `sequencer` slot with an `isSequencer` set + `sequencerThreshold`, adds EIP-712 `SettlementAttestation` verification, and removes on-chain genesis anchoring). The contract-side quorum machinery is solid, but the node side only implements a 1-of-1 subset of what the portal now accepts, plus a few genesis/bootstrap and factory-cutover concerns need human confirmation.

**Head-drift note:** findings were originally verified against `e693d411`; current head is `f052e29`. The following audited findings are now **stale/resolved** and were dropped:
- _Legacy sequencer slot still authorizes zone-side privileges_ — RESOLVED (slots removed; auth is now the `isSequencer` set).
- _Provisioning signs an owner-only factory call with the zone key_ — RESOLVED (`dev.rs` now asserts `factory_owner == dev_address`).
- _TIP-403 `migrateTransferPolicyIds` selector denylist halts payload build_ — STALE (no such file/selector at this head).

Inline comments below cover the surviving actionable findings.

<details>
<summary>Reviewer Callouts</summary>

- :zap: **Native ZoneFactory zone-ID allocation** (`crates/contracts/src/precompiles/zone_factory.rs`, native precompile not in diff): confirm the TIP-1091 native factory's `nextZoneId` counter cannot restart or collide with IDs already allocated by the legacy factory — consumers derive chain/RPC-auth domains from the zone ID, so reuse cross-authenticates zones.
- :zap: **Zero-anchor / canonical-genesis bootstrap** (`specs/ref-impls/src/tempo/ZonePortal.sol` `initialize`): init no longer sets `blockHash`/`genesisTempoBlockNumber`; `blockHash` defaults to 0. Confirm the first `submitBatch` bootstraps correctly from a zero anchor and that genesis tooling matches the new flow.
- :zap: **Active-active monitor resync** (`crates/sequencer/src/monitor.rs` finalized-batch processing, outside this diff): the audited PoC showed a backup monitor retrying a stale range forever after another sequencer advanced the portal first. With multiple active submitters now first-class, re-verify the monitor updates `last_submitted_zone_block` and re-reads portal state when it loses the race.

</details>

crates/sequencer/src/settlement.rs
```diff
@@ -430,6 +497,70 @@ impl BatchSubmitter {
...
+            anchor_block_hash,
+            block_transition,
+            deposit_transition,
+            verifier_config,
+        } = attestation;
+        let zone_id = self.portal.zoneId().call().await?;
+        let sequencer_set_version = self.portal.sequencerSetVersion().call().await?;
+        let sequencer_threshold = self.portal.sequencerThreshold().call().await?;
+        eyre::ensure!(
+            sequencer_threshold == 1,
```
legion2002, line 517: :warning: **[ISSUE]** Node cannot produce the M-of-N certificates the portal now accepts

The portal now supports `MAX_SEQUENCERS = 8` and any `sequencerThreshold` up to the set size, and `admin` can freely set it via `setSequencerSet`. But this path hard-fails with `sequencer_threshold == 1` and only ever submits a single signature (`vec![signature]`). If the admin configures a threshold >= 2 (fully valid on-chain), every `submit_batch` errors and settlement + withdrawals halt indefinitely — funds effectively freeze until threshold is reset to 1. Contract capability and node capability are out of sync.

**Recommended Fix:**
Implement multi-signer aggregation on the node (collect/submit N signatures), or constrain the portal/factory so a threshold the node cannot satisfy cannot be configured, and surface the mismatch as an operational alarm rather than a silent settlement stall.

crates/l1/src/subscriber.rs
```diff
@@ -256,24 +251,17 @@ impl L1Subscriber {
...
-            warn!(
-                "Portal genesisTempoBlockNumber is 0 — skipping backfill. \
-                 Set --l1.genesis-block-number to backfill from the correct block."
-            );
-            return Ok(None);
-        }
-
-        info!(genesis = on_chain, "Using portal's genesisTempoBlockNumber");
-        Ok(Some(on_chain + 1))
+        warn!(
```
legion2002, line 254: :warning: **[ISSUE]** Removing the on-chain genesis fallback can leave a fresh node unable to backfill

`resolve_start_block` no longer reads the portal's `genesisTempoBlockNumber` fallback. Without a local `tempoBlockNumber` checkpoint and without an explicit `--l1.genesis-block-number`, it just logs this warning and returns `Ok(None)`, skipping backfill entirely and starting live subscription at the L1 tip. A freshly provisioned zone that misses the CLI flag silently never backfills its deposit history.

**Recommended Fix:**
Fail fast (or require the flag) when neither a local checkpoint nor an explicit genesis block is available, or persist the creation block during provisioning so the fallback is deterministic.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -1001,4 +1062,75 @@ contract ZonePortal is IZonePortal {
...
+        bytes calldata verifierConfig,
+        bytes[] memory signatures
+    )
+        internal
+        view
+        returns (bool)
+    {
+        uint256 threshold = sequencerThreshold;
+        if (
+            nextZoneHeight <= zoneHeight || signatures.length < threshold
```
legion2002, line 1082: :shield: **[DEFENSE-IN-DEPTH]** `nextZoneHeight` has no upper-bound sanity check

`_verifySettlement` only enforces `nextZoneHeight > zoneHeight` (monotonic increase) and binds the height into the certificate, but it is not tied to actual block content and has no upper bound. A single certificate signed for an implausibly large height (operator bug or compromised quorum) sets `zoneHeight` to that value, after which every future submission fails `nextZoneHeight <= zoneHeight` and settlement is permanently wedged. The height is now quorum-signed and node-capped to u64, so the audited "unbounded" root cause is largely mitigated, but the wedging surface remains.

**Recommended Fix:**
Bound the per-batch height increase to a sane maximum (e.g. reject increments beyond a configurable delta) so one erroneous certificate cannot irrecoverably halt settlement.

crates/sequencer/src/settlement.rs
legion2002: this is fine, it is intentional to block threshold > 1 for now

crates/l1/src/state/provider.rs
```diff
@@ -243,10 +244,10 @@ impl L1StateProvider {
         self.get_storage(address, slot, block_number)
     }
 
-    /// Read the active sequencer address from the configured portal at the latest known L1 height.
-    pub fn get_latest_sequencer(&self) -> Result<Address> {
-        let value = self.get_latest_storage(self.portal_address, PORTAL_SEQUENCER_SLOT)?;
-        Ok(Address::from_slice(&value.as_slice()[12..]))
+    /// Read active sequencer membership from the configured portal at the latest known L1 height.
+    pub fn is_active_sequencer(&self, account: Address) -> Result<bool> {
+        let slot = keccak256((account, PORTAL_IS_SEQUENCER_SLOT).abi_encode());
```
0xKitsune, line 249: Lets plan to open a follow on to derive the slots from the sol! bindings.

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: think this is a won't fix

crates/node/src/rpc.rs
```diff
@@ -307,16 +307,25 @@ impl<Api: EthApiTypes + 'static> ZoneRpc<Api> {
             .map_err(internal)
     }
 
-    async fn zone_sequencer(&self) -> Result<Address, JsonRpcError> {
+    async fn zone_sequencers(&self) -> Result<Vec<Address>, JsonRpcError> {
         if self.config.zone_portal.is_zero() {
-            return Ok(Address::ZERO);
+            return Ok(Vec::new());
         }
```
0xKitsune: When would this be zero? Seems like there should always be a zone portal configured?

crates/node/src/rpc.rs
```diff
@@ -326,7 +335,14 @@ impl<Api: EthApiTypes + 'static> ZoneRpc<Api> {
     ) -> Result<(), JsonRpcError> {
         let caller = auth.caller;
         zone_rpc::policy::enforce_authorized(request, auth, async {
-            Ok(self.zone_sequencer().await? == caller)
+            if self.config.zone_portal.is_zero() {
+                return Ok(false);
+            }
```
0xKitsune: Same thing here when would this be zero?

crates/l1/src/subscriber.rs
legion2002: I think the warn here is fine, I don't see a reason to hard error, as this case only happens as the last fallback if a locally persisted tempoBlockNumber is not present _and_ the flag is not present.

crates/node/src/rpc.rs
```diff
@@ -454,7 +470,7 @@ where
     }
 
     fn coinbase(&self) -> BoxFut<'_> {
-        Box::pin(async move { to_raw(&self.zone_sequencer().await?) })
+        Box::pin(async move { to_raw(&Address::ZERO) })
```
0xKitsune: Why do we fix this to 0?

crates/node/src/node.rs
```diff
@@ -941,18 +942,20 @@ where
...
+        .with_max_tx_input_bytes(ctx.config().txpool.max_tx_input_bytes)
+        .with_local_transactions_config(pool_config.local_transactions_config.clone())
+        .set_tx_fee_cap(ctx.config().rpc.rpc_tx_fee_cap)
+        .with_max_tx_gas_limit(ctx.config().txpool.max_tx_gas_limit)
+        .set_block_gas_limit(ctx.chain_spec().genesis().gas_limit)
+        .disable_balance_check()
+        .with_minimum_priority_fee(ctx.config().txpool.minimum_priority_fee)
+        .with_custom_tx_type(TempoTxType::AA as u8)
+        .no_eip4844()
+        .build::<TempoPooledTransaction, _>(blob_store.clone());
```
0xrusowsky: this is a regression, you should preserve usage of `evm_config`

crates/node/src/node.rs
0xrusowsky, line 1: you should revert all changes in this file

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -151,19 +149,19 @@ mod tests {
     };
 
     #[derive(Clone, Copy)]
-    struct MockSequencer {
-        address: Option<Address>,
+    struct MockSequencerSet {
+        member: Option<Address>,
     }
```
0xKitsune, line 154: Do we need this change?

0xKitsune (dismissed):

crates/node/src/node.rs
legion2002: yup done, needed to update tempo to the latest head of your PR

crates/node/src/rpc.rs
legion2002: Addressed in 7b64523f.

crates/node/src/rpc.rs
legion2002: Addressed in 543049ab.

crates/node/src/rpc.rs
legion2002: Addressed in 543049ab.

0xrusowsky (approved):

PR #755, opened 2026-07-21
fix(l1): restore enabled tokens in state cache

decofe:
## Summary
- seed enabled tokens into `L1StateCacheInner` at the persisted L1 checkpoint
- update the cache from confirmed `TokenEnabled` events
- reset event-derived tokens to the checkpoint baseline on reorgs
- add regression coverage for tracking and reset behavior

## Validation
- `git diff --check`
- `cargo test -p zone-l1 state::cache::tests` *(blocked: sandbox Git credentials cannot fetch pinned `alloy-evm` revision)*

Prompted by: @0xrusowsky

### DIFF
--- a/crates/l1/src/state/cache.rs
+++ b/crates/l1/src/state/cache.rs
@@ -64,6 +64,10 @@ impl L1StateCache {
 #[derive(Debug, Default)]
 pub struct L1StateCacheInner {
     tracked_contracts: HashSet<Address>,
+    /// Tokens enabled on the ZonePortal at the current canonical L1 anchor.
+    enabled_tokens: HashSet<Address>,
+    /// Enabled tokens discovered at the persisted zone checkpoint.
+    initial_enabled_tokens: HashSet<Address>,
     /// Per-slot value history: `(address, slot) → { block_number → value }`.
     /// The `BTreeMap` enables efficient range lookups for "latest value at or before block N".
     slots: HashMap<(Address, B256), BTreeMap<u64, B256>>,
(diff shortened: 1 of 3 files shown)


0xKitsune:
cyclops audit fast

0xKitsune (dismissed):

crates/node/src/node.rs
```diff
@@ -452,6 +454,22 @@ where
...
+            .await
+        {
+            Ok(tokens) => {
+                self.l1_config
+                    .l1_state_cache
+                    .write()
+                    .initialize_enabled_tokens(tokens.iter().copied());
+                info!(target: "reth::cli", count = tokens.len(), tempo_block_number, "Initialized enabled-token L1 cache");
+            }
+            Err(err) => {
```
0xKitsune: Should we fail or retry here instead of continuing? On restart the subscriber resumes at `tempo_block_number + 1`, so if this historical read fails, tokens enabled before the checkpoint are never replayed and the cache stays incomplete until another restart. We probably need to special case the pre-portal genesis bootstrap, but otherwise I think this should be required.

0xrusowsky: addressed in [`04e4d4c` (this PR)](https://github.com/tempoxyz/zones/pull/755/commits/04e4d4c14c9980a588b67f357822a2caa7137ad7)

0xKitsune (dismissed):

crates/l1/src/state/cache.rs
```diff
@@ -156,10 +160,31 @@ impl L1StateCacheInner {
...
+    /// Returns the tokens enabled at the current canonical L1 anchor.
+    pub fn enabled_tokens(&self) -> &HashSet<Address> {
+        &self.enabled_tokens
+    }
+
     /// Clears chain-derived data while retaining the tracked-contract set and initial floor.
     pub fn clear(&mut self) {
         self.slots.clear();
         self.invalidations.clear();
+        self.enabled_tokens = self.initial_enabled_tokens.clone();
```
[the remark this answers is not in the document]
0xalpharush: https://github.com/tempoxyz/zones/pull/764

PR #760, opened 2026-07-21
feat: require advanceTempo first in zone blocks

decofe:
## Summary

- require the first transaction in every zone block submitted through `newPayload` to be the `advanceTempo` system transaction
- reject duplicate `advanceTempo` transactions and blocks that omit it during `convert_payload_to_block`
- keep the block executor stateless so payload-builder view simulations are not subject to whole-block layout checks
- add an end-to-end `newPayload` regression test for a block with `advanceTempo` removed
- keep the change scoped to the required block-layout invariant, excluding the unrelated policy changes in #745

This enforces the one-to-one block mapping documented in #759 at the payload-validation boundary.

## Validation

- `cargo +nightly fmt -- crates/evm/src/executor.rs crates/payload/src/attrs.rs`
- `cargo check -p zone-payload`
- `cargo test -p zone-node --test it test_payload_without_advance_tempo_is_rejected -- --nocapture`
- `git diff --check`

Prompted by: @0xalpharush

### DIFF
--- a/crates/evm/src/executor.rs
+++ b/crates/evm/src/executor.rs
@@ -91,7 +91,17 @@ where
             && tx_env.tempo_tx_env.is_none()
             && tx_env.kind() == TxKind::Call(ZONE_INBOX_ADDRESS)
             && tx_env.data.get(..4) == Some(ZoneInbox::advanceTempoCall::SELECTOR.as_slice());
-        validate_advance_tempo_position(is_advance_tempo, self.inner.receipts.len())?;
+        if self.inner.receipts.is_empty() {
+            if !is_advance_tempo {
+                return Err(validation_error(
+                    "advanceTempo must be the first transaction in every zone block",
+                ));
+            }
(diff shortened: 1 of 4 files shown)


crates/evm/src/executor.rs
```diff
@@ -108,6 +118,23 @@ where
             tempo_tx_env.expiring_nonce_idx = None;
         }
 
+        let is_advance_tempo = tx_env.is_system_tx
```
adityapk00: I wonder if we should put this check in attrs.rs inside convert_payload_to_block(). Putting it here is risky, because the payload builder uses the executor for non-committing view
  simulations too, which might not be 'valid' blocks

adityapk00 (changes requested):

crates/evm/src/executor.rs
0xalpharush: much nicer! fixed

jenpaff:
superseded by #963

PR #764, opened 2026-07-21
fix: invalidate enabled token cache entries

decofe:
## Summary

- invalidate the enabled TIP-20 address before publishing L1 receipt coverage
- prevent pre-creation policy values from inheriting past `TokenEnabled`
- add regression coverage for poisoned policy `0` followed by initialized policy `1`

## Validation

- `git diff --check`
- focused test blocked before compilation by a 401 fetching private `alloy-evm`

Prompted by: @0xalpharush


0xalpharush:
cyclops audit

PR #765, opened 2026-07-21
docs: bind encrypted deposits to sender in HKDF

decofe:
## Summary
- define the canonical 104-byte encrypted-deposit HKDF info encoding
- bind key derivation to the portal-recorded sender to prevent cross-sender replay
- clarify that ciphertext and calldata sizes do not increase

## Validation
- `git diff --check`

Prompted by: dankrad


Welith:
This PR binds the ciphertext to the immediate L1 depositor. That works for direct deposits, but with the shared swap-and-deposit router, every user has the same depositor identity: the router. Another router user can copy the ciphertext, fund a new routed deposit, and choose themselves as the refund recipient. When the replay fails, the refund path reveals the original private Zone recipient and memo.

dankrad:
> When the replay fails, the refund path reveals the original private Zone recipient and memo.

I don't see how the refund exposes this information?

I don't think there is currently a path in which replaying deposits reveals information, but we still want to bind encrypted deposits. Idk how to best fix this, should we add tx.origin maybe?

jenpaff:
superseded by https://github.com/tempoxyz/zones/pull/1028

PR #768, opened 2026-07-22
fix(evm): exclude simulated outbox views from fee override

Rjected:
## Summary

Avoids leaking the executor's fee-manager override into payload-builder simulations of `ZoneOutbox.getPendingWithdrawals()`.

## Root cause

The simulation uses `CommitChanges::No`, but the override writes before that decision. The builder therefore included a `validatorTokens[beneficiary]` change that fresh block execution could not reproduce, yielding a different state root.

## Validation

- `cargo test -p zone-evm`


Rjected:
superseded by https://github.com/tempoxyz/zones/pull/771

PR #771, opened 2026-07-22
refactor(payload): call outbox view through EVM

mattsse:
Simplify pending withdrawal reads by calling `ZoneOutbox.getPendingWithdrawals` directly through the payload EVM.

This removes the fake system transaction and builder transaction hooks, inlines the one-off view call, and preserves explicit handling for EVM reverts and halts.


Rjected (approved):
lgtm, system call does not override the token, which is only done in tx stuff

PR #775, opened 2026-07-22
test(node): reproduce stale payload transaction context

0xalpharush:
## Summary

- add two expected-red node e2e regressions for stale `ctx.tx` state in payload construction
- cover both the builder-only outbox view introduced by #771 and a rejected transaction-pool candidate
- run full leader nodes and assert follower import so the fixed cases must survive independent canonical execution

No production behavior changes are included.

## Why #771 is incomplete

[#771](https://github.com/tempoxyz/zones/pull/771) moved the builder-only `getPendingWithdrawals` call to `Evm::transact_system_call`. That system call still installs its transaction environment in `ctx.tx`. The subsequent `finalizeWithdrawalBatch` is executed through `builder.execute_transaction(tx)` and **is committed**, but its validator-token override can resolve the builder-only view as its predecessor. Canonical block execution never sees that view.

The block-8 regression executes `advanceTempo`, an alternate-token transaction-pool transfer, the builder-only outbox view, and the committed finalizer. The body contains exactly the three committed transactions; the view is not encoded. The resulting payload is rejected because the builder and canonical execution disagree on the state root.

## Rejected candidate continuation

The block-4 regression admits Alice's explicit alternate-fee-token transfer and Bob's pathUSD transaction against the same parent state. The next L1 anchor invalidates Alice's inherited token-policy slot while the fixture's RPC fallback is deliberately unavailable. Alice therefore takes the production transient-L1-storage continuation in `execute_pool_transactions`; Bob is still included. Alice's rejected transaction environment remains in `ctx.tx`, so Bob's validator-token override differs between the builder and canonical execution.

The built body contains only `advanceTempo + Bob` (`tx_count=2`). Alice does not consume a nonce.

## No FeeAMM dependency

Neither regression creates a FeeAMM pool or liquidity. Both use zero-fee Zone transactions, which the Zone pool already admits, so the tests exercise the stale-context bug without a mechanism that Zones will disable.

This matches the direction of [Tempo #6917](https://github.com/tempoxyz/tempo/pull/6917), which adds a transaction-pool option for nodes such as Zones to bypass FeeAMM liquidity validation when an alternate fee manager is used.

## E2E boundary

These are self-contained node e2e tests, not executor unit tests:

- transactions enter through HTTP RPC and the production transaction pool
- the leader runs the production payload builder and `ZoneEngine::newPayload`
- synthetic L1 blocks enter through the existing deposit-queue fixture
- a separate follower imports the complete leader block over authenticated P2P and independently executes its encoded body once the leader accepts it

The expected-red base branch currently fails at the leader's `newPayload`; after the stale-context fix, the follower assertion provides a second independent replay boundary.

## Expected failures on `main`

```text
RUST_LOG=zone_l1::subscriber=off,zone::engine=error \
  cargo test -p zone-node --test it payload_state_root -- --nocapture

test_builder_only_outbox_view_preserves_payload_state_root
  Built zone payload number=8 ... tx_count=3
  Invalid payload for block 8
  eth_blockNumber remains 7

test_rejected_pool_candidate_preserves_payload_state_root
  Built zone payload number=4 ... tx_count=2
  Invalid payload for block 4
  eth_blockNumber remains 3
```

Both failures are intentional and show that the builder-produced state root is rejected by canonical execution.

## Validation

- `git diff --check`
- `cargo test -p zone-node --test it test_p2p_follower_tracks_leader_balance -- --nocapture` — 1 passed
- `cargo test -p zone-node --test it test_withdrawal_batch_finalization -- --nocapture` — 1 passed
- both new regressions — expected-red at `ZoneEngine::newPayload`, as shown above

cc @Rjected

### DIFF
--- a/crates/node/tests/it/e2e.rs
+++ b/crates/node/tests/it/e2e.rs
@@ -10,24 +10,68 @@ use std::{net::TcpListener, time::Duration};
 use alloy::primitives::{Address, B256, Bytes, TxKind, U256, address};
 use alloy_consensus::Transaction;
 use alloy_eips::NumHash;
-use alloy_provider::{DynProvider, Provider};
+use alloy_network::ReceiptResponse;
+use alloy_provider::{DynProvider, Provider, ProviderBuilder};
 use alloy_rpc_types_eth::TransactionRequest;
+use alloy_signer_local::{MnemonicBuilder, coins_bip39::English};
 use alloy_sol_types::SolCall;
+use tempo_alloy::{
+    TempoNetwork,
+    rpc::{TempoCallBuilderExt, TempoTransactionRequest},
+};
 use tempo_chainspec::spec::TEMPO_T0_BASE_FEE;
+use tempo_contracts::precompiles::ITIP20;
 use tempo_precompiles::PATH_USD_ADDRESS;
 use tempo_zone_contracts::{
     TEMPO_STATE_ADDRESS, TempoState, Withdrawal, ZONE_INBOX_ADDRESS, ZONE_OUTBOX_ADDRESS,
     ZoneInbox, ZoneOutbox,
 };
-use zone_l1::ChainTempoStateExt;
+use zone_l1::{ChainTempoStateExt, EnabledToken};
 
 use crate::utils::{
-    DEFAULT_POLL, DEFAULT_TIMEOUT, L1Fixture, WITHDRAWAL_TX_GAS, ZoneTestNode, approve_outbox,
-    leader_p2p_config, local_dev_zone_account, poll_until, seed_fixture_for_zone,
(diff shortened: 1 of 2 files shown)


0xalpharush:
@Rjected — these expected-red regressions reproduce both stale-`ctx.tx` paths. The outbox-view case remains red with #768’s selector guard applied because the view still replaces `ctx.tx` before `finalizeWithdrawalBatch`.

Rjected:
we use
```
    builder
        .execute_transaction(tx)
```

for `finalizeWithdrawalBatch`, which commits, so i think the test doesn't match, since it calls `finalizeWithdrawalBatch` and does not commit

PR #779, opened 2026-07-22
feat(precompiles): `ZoneInbox`

0xrusowsky:
introduces a native `ZoneInbox` precompile.

depends on https://github.com/tempoxyz/zones/pull/753


0xrusowsky:
cyclops audit

crates/primitives/src/constants.rs
```diff
@@ -68,6 +68,12 @@ pub const PORTAL_IS_SEQUENCER_SLOT: B256 = {
     B256::new(bytes)
 };
 
+/// ZonePortal storage slot 3: `currentDepositQueueHash` (bytes32).
+pub const PORTAL_CURRENT_DEPOSIT_QUEUE_HASH_SLOT: B256 = B256::with_last_byte(3);
```
legion2002, line 68: can use the exported constants from tempo

crates/contracts/src/precompiles/zone_inbox.rs
```diff
@@ -32,11 +35,10 @@ crate::sol! {
             Encrypted,
         }
 
-        /// A queued deposit (regular or encrypted) passed to `advanceTempo`.
+        /// A canonical deposit (regular or encrypted) passed to `advanceTempo`.
         struct QueuedDeposit {
             DepositType depositType;
             bytes depositData;
-            bool rejected;
```
legion2002, line 39: from conversation: we're going to add this back in either in this PR or a followup

crates/l1/src/block.rs
```diff
@@ -18,7 +18,8 @@ impl L1BlockDeposits {
     ///
     /// Decrypts encrypted deposits and ABI-encodes into the types the `advanceTempo` call expects.
     /// Mint-recipient policy is enforced by upstream TIP-20 after the L1 state is anchored.
-    /// The resulting [`PreparedL1Block`] is ready to be passed via payload attributes to the builder.
+    /// The resulting [`PreparedL1Block`] is ready to be passed via payload attributes to the
+    /// builder.
     pub async fn prepare(
         self,
         sequencer_key: &k256::SecretKey,
```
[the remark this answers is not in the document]
0xKitsune: This looks valid. `prepare` needs to select the private key by `keyIndex` and retain old keys for the grace period.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+            DepositType::Encrypted => {
+                EncryptedDeposit::abi_decode(&queued.depositData).map(Self::Encrypted)
+            }
+            _ => return Err(ZonePrecompileError::MalformedCalldata),
+        }
+        .map_err(|_| ZonePrecompileError::MalformedCalldata)
+    }
+}
+
+fn decode_deposits(deposits: Vec<QueuedDeposit>) -> ZoneResult<Vec<DecodedQueuedDeposit>> {
```
legion2002, line 353: nit: do we need this helper? seems like only used in 1 place

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+        decryption.sharedSecretYParity,
+        &key_x.0,
+        key_y_parity,
+        &decryption.cpProof.s.0,
+        &decryption.cpProof.c.0,
+    ) {
+        return Ok(None);
+    }
+
+    let info = hkdf_info(
```
legion2002, line 377: trusting this cryptography is correct in the ref impl

crates/precompiles/src/chaum_pedersen/dispatch.rs
```diff
@@ -27,12 +24,12 @@ impl TempoPrecompile for ChaumPedersenVerify {
                     verifyProof(call) => {
                         debug!(target: "zone::precompile", "ChaumPedersenVerify: verifyProof");
 
-                        let mut storage = StorageCtx::default();
-                        if storage.deduct_gas(CP_VERIFY_GAS).is_err() {
-                            return Ok(storage.halt_output(PrecompileHalt::OutOfGas));
+                        let storage = StorageCtx::default();
+                        if let Err(error) = Self::charge_gas() {
```
0xKitsune: nit: can we make this `verify_proof_gas()` or similar?

crates/precompiles/src/chaum_pedersen/mod.rs
```diff
@@ -76,6 +76,62 @@ impl ChaumPedersenVerify {
             |data, caller| Self.call(data, caller),
         )
     }
+
+    /// Charge the gas cost for Chaum-Pedersen proof verification.
+    pub fn charge_gas() -> tempo_precompiles::Result<()> {
```
0xKitsune: nit: can we make this `verify_chaum_pederson_gas()`?

crates/precompiles/src/outbox/mod.rs
```diff
@@ -229,14 +231,27 @@ impl ZoneOutbox {
         Ok(recipient)
     }
 
+    #[cfg(test)]
+    pub(crate) fn seed_fallback_recipient(
```
legion2002, line 284: unused helper?

crates/primitives/src/constants.rs
```diff
@@ -68,6 +68,12 @@ pub const PORTAL_IS_SEQUENCER_SLOT: B256 = {
...
+
+/// ZonePortal storage slot 5: `_encryptionKeys` dynamic array.
+pub const PORTAL_ENCRYPTION_KEYS_SLOT: B256 = B256::with_last_byte(5);
```
0xKitsune, line 71: Can we derive these from the Solidity storage layout instead of duplicating the slot numbers? Can also do this as a follow on.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+    /// Dispatch an Inbox ABI call using execution-local L1 state.
+    pub(crate) fn call_with_l1_state<P>(
+        &mut self,
+        l1: &L1State<P>,
+        portal_address: Address,
+        calldata: &[u8],
+        msg_sender: Address,
+    ) -> PrecompileResult
+    where
+        P: L1StorageReader,
```
0xKitsune: nit: can we move this to a `dispatch.rs` mod?

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,472 @@
...
+
+    /// Create the direct-call-only native Inbox precompile.
+    pub fn create<P>(l1: L1State<P>, env: &crate::ZonePrecompileEnv) -> DynPrecompile
+    where
+        P: L1StorageReader,
+    {
+        crate::execution::create_precompile(
+            "ZoneInbox",
+            env,
+            crate::execution::NoCallRules,
```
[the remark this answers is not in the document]
0xKitsune: This looks valid. We should make `advanceTempo` system-only and enforce that it can only run once as the first tx in the block.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+                        if self.storage.is_static() {
+                            Ok(self.storage.revert_output(Bytes::new()))
+                        } else {
+                            self.advance_tempo(l1, portal_address, msg_sender, call)
+                                .encode_precompile_result(0, 0, |()| Bytes::new())
+                        }
+                    },
+                }
+            },
+        )
```
0xKitsune: Can we make the builder assert that this system tx succeeded? `execute_transaction` returns `Ok` for a revert, so right now we can freeze a payload and pop the L1 block without advancing Tempo.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,471 @@
...
+        Ok(())
+    }
+
+    fn process_deposit(
+        &mut self,
+        outbox: &mut ZoneOutbox,
+        current_hash: B256,
+        deposit: Deposit,
+    ) -> ZoneResult<()> {
+        if deposit.tempoRefundRecipient.is_zero() {
```
0xKitsune, line 197: This should be checked by the portal at deposit time and we should not need to handle this case.

legion2002, line 197: we don't need to handle this case, because the portal should never allow this.
We should probably have an assertion that this is never 0. This will allow us to remove `process_withdrawal_bounce_back`

https://github.com/tempoxyz/zones/blob/308bc1b67c31c0872c63c8843986a92b90919bf0/specs/ref-impls/src/tempo/ZonePortal.sol#L659-L664

crates/precompiles/src/inbox/mod.rs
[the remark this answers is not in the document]
0xrusowsky: hardened in [`c63db08` (this PR)](https://github.com/tempoxyz/zones/pull/779/commits/c63db0845b7a2540231322df321d226455b5fb34)

crates/evm/src/lib.rs
```diff
@@ -97,6 +99,11 @@ where
             Some(TempoState::create(l1.clone(), &env))
         });
         precompiles.apply_precompile(&ZONE_TX_CONTEXT_ADDRESS, |_| Some(ZoneTxContext::create()));
+        let inbox_env = env.clone();
+        let inbox_l1 = l1.clone();
+        precompiles.apply_precompile(&ZONE_INBOX_ADDRESS, move |_| {
```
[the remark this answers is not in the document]
0xrusowsky: we will restore rejected

crates/l1/src/block.rs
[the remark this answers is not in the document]
0xrusowsky: since this seems to be an outstanding issue not related to the contract > precompile migration, let's do it in a follow-up tomorrow?

crates/precompiles/src/inbox/mod.rs
0xrusowsky: it helps defining the types in a non-verbose manner

crates/precompiles/src/outbox/mod.rs
0xrusowsky: it is used at ~L750

crates/primitives/src/constants.rs
0xrusowsky: addressed in [`09775777`](https://github.com/tempoxyz/zones/commit/09775777)

crates/primitives/src/constants.rs
0xrusowsky: addressed in [`09775777`](https://github.com/tempoxyz/zones/commit/09775777)

crates/precompiles/src/chaum_pedersen/dispatch.rs
0xrusowsky: addressed in [`09775777`](https://github.com/tempoxyz/zones/commit/09775777)

crates/precompiles/src/chaum_pedersen/mod.rs
0xrusowsky: addressed in [`09775777`](https://github.com/tempoxyz/zones/commit/09775777)

crates/precompiles/src/inbox/mod.rs
0xrusowsky: addressed in [`09775777`](https://github.com/tempoxyz/zones/commit/09775777)

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,407 @@
...
+        Ok(())
+    }
+
+    fn process_deposit(
+        &mut self,
+        outbox: &mut ZoneOutbox,
+        current_hash: B256,
+        deposit: Deposit,
+    ) -> ZoneResult<()> {
+        if self.try_mint(deposit.token, deposit.to, deposit.amount)? {
```
0xKitsune, line 201: Withdrawal bounce-backs arrive as regular deposits with `tempoRefundRecipient == address(0)` and the fallback nonce encoded in `to`. This mints to the nonce address instead of consuming the saved recipient from `ZoneOutbox`, stranding the funds. We need to restore the bounce-back branch here and park the refund if the mint fails.

crates/precompiles/src/inbox/mod.rs
0xrusowsky: i think this should stay because despite the public `deposit` and `depositEncrypted` entry points do reject a zero `tempoRefundRecipient`, `_enqueueBounceBack` creates a regular deposit with `tempoRefundRecipient = address(0)` after a withdrawal or callback fails

0xrusowsky: i think this should stay because despite the public `deposit` and `depositEncrypted` entry points do reject a zero `tempoRefundRecipient`, `_enqueueBounceBack` creates a regular deposit with `tempoRefundRecipient = address(0)` after a withdrawal or callback fails

crates/precompiles/src/inbox/mod.rs
0xrusowsky: addressed in [`c63db08` (this PR)](https://github.com/tempoxyz/zones/pull/779/commits/c63db0845b7a2540231322df321d226455b5fb34). If tx fails we return `BlockValidationError`

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,472 @@
...
+        // Step 3: Bind the canonical Tempo queue head into the execution witness.
+        //
+        // `current_hash` may be an ancestor of this value when the sequencer processes only a
+        // bounded prefix of pending deposits. The batch proof validates that hashing the
+        // unprocessed suffix from `current_hash` reaches `tempo_current_hash`; requiring equality
+        // here would incorrectly forbid partial processing.
+        // A zero portal denotes the explicit no-L1 mode used by local development and offline
+        // execution. There is no canonical queue to bind in that mode.
+        if !portal.is_zero() {
+            let _tempo_current_hash = l1.read_l1_storage(
```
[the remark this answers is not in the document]
0xrusowsky: this will be done once we have a prover

crates/precompiles/src/outbox/mod.rs
```diff
@@ -85,7 +84,10 @@ impl ZoneOutbox {
         to: Address,
         gas_limit: u64,
     ) -> ZoneResult<()> {
-        let token_config = self.read_portal_slot(l1, portal_token_config_slot(token))?;
+        let token_config = self.read_portal_slot(
+            l1,
+            token.mapping_slot(PORTAL_TOKEN_CONFIGS_SLOT.into()).into(),
+        )?;
```
klkvr, line 90: i guess the API here could just be `l1.read_portal_slot(slot)` at this point

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+
+impl CallRules for InboxCallRules {
+    fn admit(&self, data: &[u8], caller: Address) -> CallCheck {
+        if !caller.is_zero() && data.starts_with(&ADVANCE_TEMPO_SELECTOR) {
+            return CallCheck::Revert(IZoneInbox::OnlySequencer {}.abi_encode().into());
+        }
+
+        CallCheck::Continue
+    }
+}
```
klkvr: can we move this into `fn advance_tempo`? i don't really like us splitting the logic between `CallRules` and actual precompile impl. this pattern makes sense for tip20 because zones don't own the implementation but here we can unify

crates/evm/src/executor.rs
```diff
@@ -77,6 +84,26 @@ where
...
+                return Err(BlockValidationError::msg(
+                    "advanceTempo must only execute once per zone block",
+                )
+                .into());
+            }
+            _ => {
+                // Non-system calls to this selector take the `(true, false)` path after the opening
+                // transaction and execute normally; the Inbox rejects their caller with a revert.
+            }
+        }
```
0xKitsune: Can we add some tests for this logic?

crates/evm/src/executor.rs
```diff
@@ -96,6 +131,12 @@ where
     fn finish(
         self,
     ) -> Result<(Self::Evm, BlockExecutionResult<Self::Receipt>), BlockExecutionError> {
+        if !self.has_advanced_tempo {
+            return Err(BlockValidationError::msg(
+                "zone block is missing its advanceTempo system transaction",
+            )
+            .into());
```
0xKitsune, line 140: Can we add some tests for this?

crates/evm/src/lib.rs
```diff
@@ -502,6 +510,124 @@ mod tests {
...
+        };
+        use tempo_precompiles::{
+            TIP403_REGISTRY_ADDRESS, storage::StorageKey, tip403_registry::tip403_registry_slots,
+        };
+        use tempo_zone_contracts::IZoneInbox;
+        use zone_precompiles::{tempo_state::TEMPO_BLOCK_NUMBER_SLOT, test_utils::MockL1Reader};
+        use zone_primitives::constants::{
+            PORTAL_CURRENT_DEPOSIT_QUEUE_HASH_SLOT, PORTAL_IS_SEQUENCER_SLOT, TEMPO_STATE_ADDRESS,
+            ZONE_INBOX_ADDRESS,
+        };
```
0xKitsune: nit: can we move imports to the top of the `tests` block?

crates/precompiles/src/chaum_pedersen/mod.rs
```diff
@@ -74,6 +74,62 @@ impl ChaumPedersenVerify {
...
+    pub fn verify_chaum_pedersen_gas() -> tempo_precompiles::Result<()> {
+        StorageCtx::default().deduct_gas(CP_VERIFY_GAS)
+    }
+
+    /// Verify a Chaum-Pedersen DLOG equality proof on secp256k1.
+    ///
+    /// Proves knowledge of scalar `x` such that `pubSeq = x*G` AND `sharedSecret = x*ephemeralPub`.
+    pub(crate) fn verify(
+        ephemeral_pub_x: &[u8; 32],
+        ephemeral_pub_y_parity: u8,
```
0xKitsune, line 88: nit: do we need this as an associated function?

crates/precompiles/src/inbox/mod.rs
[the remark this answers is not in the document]
klkvr: out of context here but why don't just compare? we don't do partial processing yet right so it will always match?

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+        );
+        let recipient = outbox.consume_fallback_recipient(ZONE_INBOX_ADDRESS, fallback_nonce)?;
+        if self.try_mint(deposit.token, recipient, deposit.amount)? {
+            self.emit_event(deposit.withdrawal_bounce_back_processed_event(recipient))?;
+        } else {
+            let previous = self.refunds[deposit.token][recipient].read()?;
+            let Some(refund) = previous.checked_add(deposit.amount) else {
+                return Err(TempoPrecompileError::under_overflow().into());
+            };
+            self.refunds[deposit.token][recipient].write(refund)?;
```
klkvr: could use `.sinc` here

crates/evm/src/executor.rs
```diff
@@ -86,7 +113,15 @@ where
...
 
         self.evm_mut().clear_l1_overlay_state();
-        result
+        let result = result?;
+        if is_advance_tempo && !result.result.result.is_success() {
+            return Err(BlockValidationError::msg(
+                "block-opening advanceTempo transaction must succeed",
+            )
+            .into());
+        }
```
klkvr: `TempoEvm` already enforces that system transactions never revert https://github.com/tempoxyz/tempo/blob/fbfe81abfee5c2d78404ca608f3ef7044b858d49/crates/evm/src/evm.rs#L290-L294

klkvr (dismissed):

crates/precompiles/src/inbox/mod.rs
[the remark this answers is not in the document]
0xrusowsky: that's what the solidity impl was doing, but i can enforce the exact match

crates/precompiles/src/inbox/mod.rs
0xrusowsky: [`a4e4e75` (this PR)](https://github.com/tempoxyz/zones/pull/779/commits/a4e4e753dd03466e97668fa653b72469855e3e97)

crates/precompiles/src/inbox/mod.rs
0xrusowsky: [`a4e4e75` (this PR)](https://github.com/tempoxyz/zones/pull/779/commits/a4e4e753dd03466e97668fa653b72469855e3e97)

crates/evm/src/executor.rs
0xrusowsky: [`c3bd6f5` (this PR)](https://github.com/tempoxyz/zones/pull/779/commits/c3bd6f528d4b832de21bd0fb4eefb4596f3762de)

crates/precompiles/src/inbox/mod.rs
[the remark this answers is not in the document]
0xrusowsky: [`1831c2a` (this PR)](https://github.com/tempoxyz/zones/pull/779/commits/1831c2a145cec67604c397f8b0905dc33949ae76)

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+
+/// Zone-side bridge Inbox state and deposit-processing logic.
+#[contract(addr = ZONE_INBOX_ADDRESS)]
+pub struct ZoneInbox {
+    /// Hash-chain head after the last processed L1 deposit.
+    processed_deposit_queue_hash: B256,
+    /// Monotonic number of deposits consumed from the L1 queue.
+    processed_deposit_number: u64,
+    /// Withdrawal bounce-back mints that failed and can be claimed later.
+    refunds: Mapping<Address, Mapping<Address, u128>>,
```
0xKitsune: nit: wdyt about naming this `withdrawal_bounce_backs` to be explicit? IIUC this is the only way funds get here.

crates/evm/src/executor.rs
```diff
@@ -77,6 +84,27 @@ where
...
+                )
+                .into());
+            }
+            (true, true) => {
+                return Err(BlockValidationError::msg(
+                    "advanceTempo must only execute once per zone block",
+                )
+                .into());
+            }
+            (false, true) => self.has_advanced_tempo = true,
```
klkvr: not sure if it matters but it actually seems wrong to flip the bool here because even though transaction was executed it might not get committed

in tempo we advance executor's sections in `commit_transaction` only

klkvr (dismissed):

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+        for queued in deposits {
+            current_hash = queued.hash_with_tail(current_hash)?;
+
+            match queued {
+                DecodedQueuedDeposit::Regular(deposit) => {
+                    self.process_deposit(&mut outbox, current_hash, deposit)
+                }
+                DecodedQueuedDeposit::Encrypted(deposit) => {
+                    let Some(decryption) = decryptions.next() else {
+                        return Err(ZoneInboxError::missing_decryption_data().into());
```
0xKitsune, line 116: Should this bounce back if decryption fails rather than erroring and exiting `advance_tempo()`?

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+                        decryption,
+                        key,
+                    )
+                }
+            }?;
+        }
+
+        if decryptions.next().is_some() {
+            return Err(ZoneInboxError::extra_decryption_data().into());
+        }
```
0xKitsune, line 134: Same thing here, should this bounce back rather than error?

crates/precompiles/src/outbox/mod.rs
```diff
@@ -274,6 +280,20 @@ impl ZoneOutbox {
         Ok(recipient)
     }
 
+    #[cfg(test)]
```
0xKitsune, line 283: nit: why isnt this in a test mod?

crates/precompiles/src/outbox/mod.rs
```diff
@@ -274,6 +280,20 @@ impl ZoneOutbox {
...
+    pub(crate) fn seed_fallback_recipient(
+        &mut self,
+        nonce: u64,
+        recipient: Address,
+    ) -> TempoResult<()> {
+        self.fallback_recipients[nonce].write(recipient)
+    }
+
+    #[cfg(test)]
```
0xKitsune, line 292: nit: same thing here should this be in a test mod?

0xKitsune (dismissed):

crates/evm/src/executor.rs
0xKitsune: Can we add an executor-level regression test proving a reverted `advanceTempo` cannot satisfy the block-opening requirement? It should also cover missing and duplicate calls since this is consensus validation.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+                .try_into()
+                .expect("address suffix is eight bytes"),
+        );
+        let recipient = outbox.consume_fallback_recipient(ZONE_INBOX_ADDRESS, fallback_nonce)?;
+        if self.try_mint(deposit.token, recipient, deposit.amount)? {
+            self.emit_event(deposit.withdrawal_bounce_back_processed_event(recipient))?;
+        } else {
+            let previous = self.refunds[deposit.token][recipient]
+                .slot()
+                .sinc(U256::from(deposit.amount))?;
```
0xKitsune: Can we add regression coverage for two failed withdrawal bounce-backs to the same token and recipient so we know the refund accumulates correctly? It should also cover overflow reverting atomically and a failed `claimRefund` preserving the stored refund.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+            }
+        }
+
+        // Step 4: Update state
+        self.processed_deposit_queue_hash.write(current_hash)?;
+        let previous_number = self.processed_deposit_number.read()?;
+        let processed_number = previous_number
+            .checked_add(deposit_count)
+            .ok_or_else(TempoPrecompileError::under_overflow)?;
+        self.processed_deposit_number.write(processed_number)?;
```
0xKitsune, line 162: Can we add a test proving failed regular and encrypted deposits still advance the processed queue hash and deposit number exactly once while enqueueing exactly one bounce-back?

crates/l1/src/block.rs
```diff
@@ -193,7 +192,7 @@ pub struct PreparedL1Block {
     /// ABI-encoded queued deposits (regular + encrypted).
     #[serde(skip)]
     pub queued_deposits: Vec<abi::QueuedDeposit>,
-    /// Decryption data for encrypted deposits accepted for on-chain verification, in order.
+    /// Decryption data for every encrypted deposit submitted for on-chain verification, in order.
```
0xKitsune, line 163: Can we add one mixed-deposit test here proving every encrypted deposit gets exactly one correctly ordered decryption entry, including the invalid-ciphertext fallback path?

crates/precompiles/src/inbox/mod.rs
legion2002: why does `_enqueueBounceBack` send `tempoRefundRecipient = address(0)` when we know it is def going to fail?

klkvr (dismissed):

0xKitsune (approved):

PR #787, opened 2026-07-22
fix(node): enforce zone call policy in txpool

mattsse:
## Summary

- enforce the Zone call-surface policy as an additional stateless txpool validation hook
- expose a shared `CalldataFilter` plus `ZonePoolBuilder::with_calldata_filter` and `with_calldata_filter_fn_opt`, so calldata policy can be injected or passed around instead of being embedded in transaction validation
- configure the production filter to allow only fully ABI-decodable `transferFrom`, `transferWithMemo`, `transferFromWithMemo`, and `approve` calls to TIP-20-prefixed addresses, including every call in an AA batch
- reject Zone system-operation calldata from user pool admission while preserving the contract-deployer creation exemption, enabled-token balance admission, EIP-7702 rejection, and disabled FeeAMM checks from `main`; deployer status does not bypass call filtering

## Why

The call-surface restriction is local mempool policy: rejecting unsupported user calls before propagation or block building gives callers an immediate deterministic error without turning the restricted call surface into execution or block-import consensus.

The policy was previously coupled directly to the transaction validator. Separating the generic transaction validation from a configurable calldata filter lets callers reuse the pool builder with a different policy, share an `Arc`-backed filter, or omit call-level filtering while retaining the base Zone admission checks.

Imported blocks retain the current execution semantics and payload checks. This does not make TIP-20 selectors invalid during EVM execution and does not add new payload-conversion rejection paths.


mattsse:
cyclops audit fast

legion2002:
cyclops audit fast

PR #788, opened 2026-07-22
feat: n-of-m sequencer signing for settlement attestations

adityapk00:
- support n-of-m signatures
- The leader will build block, send out settlement attestation requests to the followers
- Followers do full reexecution , checks and return the signed attesations back to the leader
- Leader collects these, submits them during `submitBatch`


crates/node/src/cli.rs
```diff
@@ -176,8 +176,13 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
         if should_sequence_blocks {
             let sequencer_signer = sequencer_signer
                 .expect("sequencer signer is parsed whenever sequencing is enabled");
+            let l1_transaction_signer = p2p_config
+                .as_ref()
+                .filter(|config| config.role() == Role::Leader)
```
joshieDo, line 186:
```suggestion
                .filter(|config| config.role().is_leader())
```
would be nice

crates/node/src/node.rs
```diff
@@ -527,11 +554,23 @@ where
     fn launch_p2p(
         config: P2pConfig,
         network_id: P2pNetworkId,
+        attestation_domain: AttestationDomain,
+        attestation_store: Option<AttestationStore>,
```
joshieDo: maybe these could be one type?

crates/sequencer/src/attestation.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub fn decode(encoded: &[u8]) -> eyre::Result<Self> {
+        Self::abi_decode(encoded)
+            .map_err(|err| eyre::eyre!("invalid settlement signature encoding: {err}"))
+    }
+
+    pub fn recover_signer(&self, domain: AttestationDomain) -> eyre::Result<Address> {
+        let signature = Signature::try_from(self.signature.as_ref())
+            .map_err(|err| eyre::eyre!("invalid settlement signature: {err}"))?;
+        signature
+            .recover_address_from_prehash(&domain.settlement_digest(&self.attestation))
```
joshieDo: shouldnt accept high-s , since ZonePortal won't (TIP-1020)

crates/node/src/settlement_attestation.rs
```diff
@@ -0,0 +1,391 @@
...
+        tempo_header.hash_slow() == tempo_block_hash,
+        "zone batch's Tempo block hash does not match finalized L1"
+    );
+
+    let mut parent_hash = tempo_block_hash;
+    for block_number in (tempo_block_number + 1)..=anchor_block_number {
+        let header = context
+            .l1_provider
+            .get_header_by_number(block_number.into())
+            .await?
```
joshieDo: how big can this range be? bot said 7k , if so thats too much to be fetching one by one?

crates/node/src/node.rs
adityapk00: fixed

crates/sequencer/src/attestation.rs
adityapk00: fixed

crates/node/src/settlement_attestation.rs
adityapk00: fixed. The followers shouldn't be fetching this anyway, that's really the job of the prover/verifier, but limited this to only validate only the canonical endpoint hashes

crates/node/src/settlement_attestation.rs
```diff
@@ -0,0 +1,382 @@
...
+struct BlockCommitments {
+    tempo_block_hash: B256,
+    tempo_block_number: u64,
+    processed_deposit_hash: B256,
+    processed_deposit_number: u64,
+    withdrawal: Option<(B256, u64)>,
+}
+
+/// Extract commitments produced by the deterministic system transactions in a zone block.
+fn block_commitments<P>(provider: &P, number: u64) -> eyre::Result<BlockCommitments>
```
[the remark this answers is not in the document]
adityapk00: This is wont fix to prevent complexity bloat here, but this (and other invariants) is ultimately the job of the prover

crates/p2p/src/manifest.rs
```diff
@@ -135,6 +136,9 @@ impl ZoneManifest {
     /// Parses and validates a TOML manifest.
     pub fn parse(input: &str) -> Result<Self, ManifestError> {
         let raw: RawManifest = toml::from_str(input).map_err(ManifestError::Toml)?;
+        if raw.sequencer_set_version == 0 {
```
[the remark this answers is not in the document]
adityapk00: yes, this is intentional. we'll enable this in a subsequent PR

crates/node/src/replication.rs
```diff
@@ -387,9 +445,36 @@ async fn run_leader_backfill_server<P>(
...
                     }
+                    P2pEvent::SettlementSignatureReceived { follower, signature } => {
+                        let result = async {
+                            let signed = SignedSettlementAttestation::decode(&signature)?;
+                            let signer = signed.recover_signer(attestation.domain)?;
+                            let expected_signer = attestation.addresses.get(&follower).copied()
+                                .ok_or_else(|| eyre::eyre!("unknown follower identity"))?;
+                            eyre::ensure!(signer == expected_signer, "settlement signer does not match authenticated peer");
+                            let height: u64 = signed.attestation.zoneHeight.try_into()
+                                .map_err(|_| eyre::eyre!("settlement height does not fit in u64"))?;
```
hamdiallam: nit: You can use `wrap_err`

crates/node/src/replication.rs
```diff
@@ -387,9 +445,36 @@ async fn run_leader_backfill_server<P>(
...
+                            let expected = build_settlement_attestation(
+                                &provider,
+                                height,
+                                &attestation,
+                                Some((signed.attestation.anchorBlockNumber, signed.attestation.anchorBlockHash)),
+                            ).await?.ok_or_else(|| eyre::eyre!("signed block is not a batch boundary"))?;
```
hamdiallam: Same here re `ok_or_eyre`.

crates/p2p/src/runtime.rs
```diff
@@ -384,6 +426,16 @@ fn run(
         oracle.track(0, peers).await;
         let (block_sender, block_receiver) =
             commonware.register(BLOCK_CHANNEL, network::block_quota(), BLOCK_BACKLOG);
+        let (settlement_proposal_sender, settlement_proposal_receiver) = commonware.register(
```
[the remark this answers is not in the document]
adityapk00: I dont think we need per-channel limits. The sequencers are all trusted, and if a sequencer key(s) are compromised and a third party ends up controlling them, they can stall the chain anyway.

adityapk00:
Cyclops review:

> Leader settlement-signature handler runs L1 RPC before signature verification, inline 

Fixed. The leader handler now performs the following operations in order: 1. Decode + get signer 2. Check  signer matches 3.  L1 settlement-state validation 4. Store the signature.

> High-s follower signatures poison otherwise valid settlement quorums

Fixed

> Followers can sign mid-block BatchFinalized boundaries 

wont fix. Validating the structure of the block (like the system txns are in the right place) will ultimately the job of the prover (to limit complexity of the follower)

> Follower settlement-proposal validation runs an unbounded synchronous L1 ancestry walk

Fixed. The followers shouldn't be fetching this anyway, that's really the job of the prover/verifier, but limited this to only validate only the canonical endpoint hashes

> Follower attestations omit mandatory zone system-transaction order validation

Same as above, wontfix, should be provers job. 

> Ambiguous successful submitBatch can delete the quorum certificate

Fixed. Handles submitBatch failures better

>  A latest-value Reth stream can make the collector permanently skip the first unsettled boundary

Fixed, it now treats the stream as a signal, fetching blocks from reth in order

crates/node/src/settlement_attestation.rs
```diff
@@ -0,0 +1,382 @@
...
+                break;
+            }
+            Ok(false) => {}
+            Err(err) => {
+                debug!(target: "zone::p2p", %err, number, "Skipped non-current settlement boundary during recovery")
+            }
+        }
+    }
+
+    let mut persisted = provider.persisted_block_stream();
```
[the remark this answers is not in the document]
adityapk00: fixed

crates/p2p/src/runtime.rs
hamdiallam, line 690: This loop introduces head-of-the-line blocking across all these channels. Worth splitting and isolating state among them?

Block Receiver -> Replication Actor
Signature receivers -> Settlement Actor
Backfill Receiver -> Backfill Actor

adityapk00: this is a good idea. I have another PR following this one, i'll fold this change into that PRs refactor.

crates/node/src/replication.rs

crates/node/src/replication.rs

adityapk00:
Fixed the perf issue: 

a batch boundary persisted while the prev batch was still awaiting L1 confirmation could not extend the
portal tip, causing its settlement proposal to be rejected  until retry 5s later. 

Fixed by  settlement collector now receives an event-driven notification whenever the submitted L1 height advances ( retry timer as a fallback)

mattsse (approved):
LGTM

PR #789, opened 2026-07-22
fix(precompiles): preserve TIP-403 transfer policy binding

Zygimantass:
## Summary

Preserve the L1-owned TIP-403 transfer-policy binding during zone token initialization.

## Why

ZoneTokenFactory delegates initialization to the upstream TIP-20 implementation, which writes a default policy binding into the L1-mirrored TIP-403 registry. That transition is rejected as ZoneDbError::L1Write and stalls the zone while enabling its initial token.

The precompile now snapshots and restores the complete registry word, including the policy ID and isSet bit, around token initialization. Zones has no deployed legacy state, so this behavior is unconditional rather than hardfork-gated.

## Validation

- cargo test -p zone-precompiles (73 passed)
- cargo clippy -p zone-precompiles --all-targets -- -D warnings
- rustup run nightly cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/crates/precompiles/src/tip20_factory/mod.rs
+++ b/crates/precompiles/src/tip20_factory/mod.rs
@@ -20,8 +20,10 @@ pub use IZoneTokenFactory::IZoneTokenFactoryErrors as ZoneTokenFactoryError;
 use alloy_evm::precompiles::DynPrecompile;
 use alloy_primitives::Address;
 use tempo_precompiles::{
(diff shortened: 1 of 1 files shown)


legion2002:
cyclops audit fast

0xrusowsky (commented):
lgtm, pending cyclops' green light

0xrusowsky (approved):

0xrusowsky:
NOTE: after we merge this we can do a follow-up to clean the TIP20 slot deps in the `L1OverlayDB` and `L1StateCache`

PR #790, opened 2026-07-22
fix: perform sequencer membership checks at the same block

klkvr:
fixes CHAIN-1020

Changes TIP20's `is_sequencer` check to read state at the tempo anchor block instead of reading latest state causing potentially non-deterministic execution


klkvr:
cyclops audit

crates/evm/src/lib.rs
```diff
@@ -70,15 +70,19 @@ type TempoCtx<DB> = <TempoEvmFactory as EvmFactory>::Context<DB>;
 #[derive(Debug, Clone)]
 pub struct ZoneEvmFactory<L1 = L1StateProvider> {
     l1_reader: L1,
+    portal_address: Address,
 }
```
0xrusowsky, line 72: should we do the same for `L1State` and also make it aware of the portal address, so that we don't have to pass it around?

crates/precompiles/src/ztip20/mod.rs
```diff
@@ -186,6 +184,8 @@ mod tests {
         bob: Address,
         spender: Address,
         sequencer: Address,
+        l1: L1State<MockL1Reader>,
+        l1_reader: MockL1Reader,
```
0xrusowsky, line 186: nit: don't u already have it in `L1State`

klkvr: `L1State` doesn't expose it directly and we need direct access to seed state

crates/evm/src/lib.rs
klkvr: added here 30ba4ad

0xrusowsky (dismissed):

0xrusowsky (approved):

PR #793, opened 2026-07-22
chore(precompiles): chaum-pedersen simplifications

Rjected:
Some simplifications for the chaum-pedersen precompile:
 * Use `FieldBytes::from_slice`
 * Use a hasher instead of concat + hash
 * Use a loop for hashing
 * Document that we also accept 0x05 compact encoding


0xKitsune:
cyclops audit fast

0xKitsune (approved):

PR #794, opened 2026-07-22
refactor(evm): use L1 reader portal address

0xrusowsky:
## Summary
- make `L1StorageReader::portal_address()` the authoritative ZonePortal configuration
- derive the Outbox portal from its shared `L1State` instead of accepting a second address
- remove duplicate portal arguments from `ZoneEvmFactory` and `ZoneEvmConfig`
- update the production provider and test reader implementations

This prevents Inbox/Outbox and L1-backed execution from being configured against different portal accounts.

## Validation
- `cargo test -p zone-precompiles outbox::tests --lib`
- `cargo test -p zone-evm --lib`
- `git diff --check`


0xrusowsky:
cyclops audit

0xKitsune (approved):

PR #797, opened 2026-07-22
feat: enforce zone-backed withdrawals

0xalpharush:
## Summary

- add a native `WithdrawalTracker` precompile that records `zoneBalance(user, token)` and `zoneTotalSupply(token)` independently of TIP-20 storage
- credit successful Inbox mints and debit `amount + fee` before Outbox burns, reverting withdrawals that are not backed by the requesting user's Zone balance
- gate `submitBatch` on Portal custody covering the post-batch Zone supply plus all withdrawal amounts, fees, and protocol refunds in the batch
- emit a structured warning with the token, Portal balance, required backing, and shortfall when that gate blocks a batch
- wire the precompile into genesis, the Solidity reference implementation, specifications, invariants, and integration/unit tests

## Why

The previous ExEx assertion observed bridge accounting after execution but did not enforce the user-backed withdrawal invariant in consensus state. Moving the ledger into a precompile makes the check part of ordinary EVM execution and gives the sequencer an independent supply value for verifying that processing a withdrawal batch cannot prevent remaining users from exiting.

## Impact

Withdrawal eligibility now follows successful bridge credits rather than TIP-20 balances alone. Tracker writes revert with the surrounding transaction or block, and an undercollateralized Portal stops batch submission before any other L1 submission calls and emits an actionable warning.

## Validation

- `cargo test -p zone-precompiles -p zone-sequencer -p zone-evm`
- `cargo test -p zone-node --test advance_tempo`
- `forge build --root specs/ref-impls --no-lint`
- full `forge test` from `specs/ref-impls` (550 tests)
- `forge fmt --root specs/ref-impls --check`
- `git diff --check`


crates/sequencer/src/settlement.rs
```diff
@@ -380,6 +392,8 @@ impl BatchSubmitter {
         withdrawal_queue_hash = %batch.withdrawal_queue_hash,
     ))]
     pub async fn submit_batch(&self, batch: &BatchData) -> Result<ZonePortal::BatchSubmitted> {
+        self.check_portal_backing(&batch.required_portal_backing)
```
0xalpharush, line 395: TODO followers also check before signing

0xalpharush:
cyclops audit

decofe:
Closing per author cleanup. The subsequent design split this work across #1207 for Portal solvency verification and #1217 for bounded, authenticated bridge processing; this withdrawal-tracker precompile design itself was not merged.

PR #805, opened 2026-07-23
fix: enforce refund read privacy in ZoneInbox

mattsse:
Moves refund-read authorization into `ZoneInbox` so internal calls cannot bypass private RPC policy.

- replace the public refund mapping getter with an owner/active-sequencer guarded getter
- remove the topology-dependent `eth_call` refund parser
- add direct and Multicall3 regression coverage
- regenerate the bundled zone genesis

This closes the Multicall3 bypass where an authenticated caller could forward `refunds(token, victim)` and recover another account's pending refund balance.


mattsse:
The failing **ZonePortal layout unchanged** check is expected for this PR. Adding `Unauthorized()` to `IZoneInbox` shifts Solidity's generated AST IDs, so the regenerated `ZonePortal.json` differs only in `astId` values and AST-derived type identifiers (for example, `WithdrawalQueue` 4902 → 4904). Storage labels, slots, offsets, encodings, and byte widths are unchanged—there is no actual `ZonePortal` storage layout change. Forge build and tests pass, so this failure is safe to ignore.

0xrusowsky (dismissed):

0xrusowsky:
changes LGTM, only remains fixing CI.

proposed fix in [chore: update ZonePortal storage layout artifact](https://github.com/tempoxyz/zones/pull/849) (https://github.com/tempoxyz/zones/pull/849) @mattsse

0xrusowsky:
noticed that it wasn't updating the new precompile.
needs https://github.com/tempoxyz/zones/pull/852 to impl the solidity changes too

0xrusowsky (approved):

PR #806, opened 2026-07-23
[WIP] feat(l1): version cache by verified storage roots

legion2002:
## Summary

- authenticate the ZonePortal and TIP-403 registry storage roots against each finalized header state root
- replace semantic event-based cache invalidation with address-level barriers derived from storage-root changes
- allow cross-height reuse only for continuously root-tracked accounts; all other witness reads remain exact-height-only
- retain receipt fetching and verification for portal deposit ingestion

This is stacked on #791 and is intended as a storage-root-based alternative to #803.

## Performance / WIP

This adds two concurrent `eth_getProof` requests per finalized L1 block (one for ZonePortal and one for TIP-403). Before taking this out of draft, we should benchmark provider overhead and confirm historical proof availability across the intended backfill window.

Using the global state root as the cache key would invalidate every cached account on every L1 state change. This instead uses the state root only to authenticate each account proof, then versions reuse by the individual account storage root.

## Validation

- `cargo test -p zone-l1 --locked`
- `cargo clippy -p zone-l1 --all-targets --locked -- -D warnings`
- `cargo check -p zone-node --tests --locked`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13458,6 +13458,7 @@ dependencies = [
  "reth-provider",
  "reth-storage-api",
  "reth-tasks",
+ "schnellru",
  "serde",
  "serde_json",
  "tempo-alloy",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13446,6 +13446,7 @@ dependencies = [
  "alloy-rpc-types-eth",
  "alloy-sol-types",
  "alloy-transport",
+ "alloy-trie",
  "const-hex",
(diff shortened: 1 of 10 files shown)


rkrasiuk:
> we should benchmark provider overhead and confirm historical proof availability across the intended backfill window.

L1 node should be running with `--rpc.eth-proof-window <n>`

jenpaff:
Closing as superseded by #855 and #921

PR #820, opened 2026-07-23
fix(sequencer): avoid Tempo fee history requests

mattsse:
Configure sequencer-owned Tempo L1 writes with the known T1 fee cap and zero priority fee.

The T1 fee is above both the T0 fixed base fee and the T7 dynamic base-fee cap, so transactions remain valid while Alloy no longer needs an `eth_feeHistory` request before batch submission, withdrawal processing, or encryption-key registration.


adityapk00 (approved):
lgtm

PR #822, opened 2026-07-23
feat: build immutable settlement batch plans

0xalpharush:
## Summary

- add a single `ZonePortal.batchSubmissionState(address)` getter containing the Portal-owned inputs needed to plan and validate a batch
- cache the L1 chain ID and EIP-712 domain separator, and validate every pinned Portal snapshot against them
- build one immutable `BatchPlan` containing transitions, pinned Portal state, anchor data, witness commitment, attestation digest, proof, and signatures
- submit the exact frozen plan across retries; explicitly reconcile, invalidate, or rebuild when Portal state or the canonical anchor changes
- pin follower settlement proposals to the same Portal block number and hash used by the leader

## Why

The previous submission path progressively fetched mutable Portal and L1 values while constructing and validating an attestation. Besides adding RPC calls, that allowed time-of-check/time-of-use drift between signing, proof construction, and submission.

This change gives each attempt series one coherent snapshot and makes retry behavior explicit.

## Impact

The sequencer now depends on the new Portal getter, so environments must deploy the updated Portal implementation before running this version.

## Validation

- `cargo test -p zone-sequencer --lib` — 59 passed
- `cargo test -p zone-node --lib` — 33 passed
- `cargo check -p zone-sequencer -p zone-node`
- `forge test --match-contract ZonePortalTest` — 151 passed
- `cargo fmt --all`
- `forge fmt --check`
- `git diff --check`


jenpaff:
superseded by https://github.com/tempoxyz/zones/pull/1000

PR #837, opened 2026-07-24
feat(xtask): multi-sequencer set support in create-zone

Zygimantass:
## Motivation

Standing up a multi-sequencer zone (e.g. 3 nodes, 2-of-3 attestation quorum) currently isn't possible through `create-zone`: it hard-codes `sequencers: [sequencer], threshold: 1`.

There is also a version mismatch that would break attestations even with a multi set installed at creation: the native factory initializes portals at `sequencerSetVersion = 0`, but the P2P manifest [rejects version 0](https://github.com/tempoxyz/zones/blob/main/crates/p2p/src/manifest.rs) and followers refuse attestation requests whose version doesn't match the manifest.

## Solution

- `--sequencer` is now repeatable (first address is the leader) and `--threshold` is added (default 1). Existing single-sequencer invocations are unchanged, including the version-0 legacy behavior.
- For multi-sequencer sets, `create-zone` creates the zone with a 1-of-1 leader set, then installs the full set via `setSequencerSet`, which bumps the on-chain version to 1 — matching a manifest with `sequencer_set_version = 1`. Fails fast when `--private-key` is not the admin key, since `setSequencerSet` is `onlyAdmin`.
- Client-side validation mirrors the portal: max 8 sequencers, unique, non-zero, `1 <= threshold <= len`.
- Adds `setSequencerSet` + `SequencerSetUpdated` to the `ZonePortal` bindings.
- `zone.json` gains `sequencers`, `sequencerThreshold`, `sequencerSetVersion` (keeps `sequencer` = leader for downstream tooling).

Fixing genesis to land at version 1 instead would require changes to the embedded portal runtime + native init in tempoxyz/tempo and a hardfork rollout; the two-step flow here is self-contained.

## Deployment note

`setSequencerSet` only exists in portal runtimes from tempo `83982c7` (#6952) onward — L1 devnets must run an image at least that new for the multi-sequencer path.

## PR Checklist

- [x] Compiles, clippy and fmt clean
- [ ] Untested against a live chain (xtask has no test harness)

### DIFF
--- a/crates/contracts/src/precompiles/zone_portal.rs
+++ b/crates/contracts/src/precompiles/zone_portal.rs
@@ -157,6 +157,7 @@ crate::sol! {
 
         event RoleUpdated(address indexed account, Role prev, Role next);
         event EnforcementModesUpdated(bool accessMode, bool gatewayMode);
+        event SequencerSetUpdated(uint64 indexed nonce, uint8 threshold, address[] sequencers);
 
         // -- Errors --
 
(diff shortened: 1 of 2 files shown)


adityapk00 (approved):
lgtm

rkrasiuk (approved):

PR #840, opened 2026-07-24
feat(xtask): support separate encryption transaction signer

Zygimantass:
## Summary

- allow `tempo-xtask set-encryption-key` to use a distinct L1 transaction signer
- keep the encryption key responsible for the proof of possession
- preserve existing single-key behavior when `TRANSACTION_PRIVATE_KEY` is unset

## Why

Multi-sequencer zones share one block-encryption key, but ZonePortal authorization requires the transaction sender to be an active settlement sequencer. Zone regenesis therefore needs the leader's `secp256k1-key` to submit the transaction while the shared `sequencer-key` signs the encryption-key proof.

## Test plan

- `cargo fmt --check -p tempo-xtask`
- `cargo test -p tempo-xtask set_encryption_key`

Prompted by: @0xrusowsky


Zygimantass:
Closing because the affected Zone has one sequencer. The existing `set-encryption-key` command already supports the correct single-key flow; the permanent repair is scoped to tempoxyz/workflows#812.

0xrusowsky:
cyclops audit fast

0xrusowsky (approved):

PR #844, opened 2026-07-24
refactor(multi-seq): Allow Leaders/followers to be promoted/demoted

adityapk00:
Refactor the multi sequencer node and runtime code so they can watch for leadership changes (leader -> follower or follower -> leader) and switch the services that leaders and followers run dynamically. This is not triggered from anywhere yet, but the (future) leadership election / recovery paths will use this.


crates/node/src/node.rs
```diff
@@ -584,8 +621,11 @@ where
         pool: N::Pool,
         engine: ConsensusEngineHandle<ZonePayloadTypes>,
         l1_block_tracker: L1BlockTracker,
+        deposit_queue: DepositQueue,
     ) -> eyre::Result<()> {
-        let role = config.role();
+        let local_ed25519_public_key = config.ed25519_public_key();
+        let leadership = config.leadership();
+        let role = leadership.borrow().role_of(&local_ed25519_public_key);
```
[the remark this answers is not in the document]
adityapk00: Yes, this PR intentionally leaves this out since we won't do leader/follower rotation here. How we do this depends on how the rotation, we'll fix in subsequent PR

crates/node/src/node.rs
```diff
@@ -734,51 +776,88 @@ where
...
-                self.l1_config.clone(),
-                ctx.node.provider().clone(),
-                self.deposit_queue.clone(),
-                ctx.node.task_executor().clone(),
-            );
-            info!(target: "reth::cli", "L1 subscriber started with deposit enqueueing");
-        }
+    /// Spawn shared L1 observation and keep the deposit queue warm on every node.
+    fn spawn_l1_subscriber(&mut self, ctx: &AddOnsContext<'_, N>) {
+        L1Subscriber::spawn(
```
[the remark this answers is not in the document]
adityapk00: fixed

crates/node/src/replication.rs
```diff
@@ -836,7 +849,12 @@ where
                 "Peer block import is waiting for local L1 observation of its anchor"
             ),
         }
-    }
+    };
+
+    // The subscriber normally enqueues immediately after recording this observation. Enqueueing
+    // here as well closes that small scheduling window and makes follower import self-contained;
+    // the queue treats the subscriber's later enqueue as a duplicate.
+    deposit_queue.enqueue_sealed(l1_header, observed);
```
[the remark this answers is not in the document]
adityapk00: this is ultimately the provers job (to enforce that the blocks has one finalizeWithdrawal at the end) I don't think we should add this to the followers checks

crates/node/src/node.rs
```diff
@@ -877,20 +957,57 @@ where
...
+                        return;
+                    }
+                    res = &mut monitor_handle.0 => {
+                        tracing::error!(target: "reth::cli", ?res, "Zone monitor task exited");
+                        return;
+                    }
+                    result = leadership.wait_for(false) => {
+                        result.expect("leadership watch closed while sequencer tasks were active");
+                        withdrawal_handle.0.abort();
+                        monitor_handle.0.abort();
```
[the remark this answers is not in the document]
adityapk00: this PR doesn't actually do the follower/leader rotation, we'll deal with this once we have the rotation mechanisms in a subsequent PR (Because shape of the fix will depend on how we do the rotation)

crates/node/src/engine.rs
```diff
@@ -126,6 +131,11 @@ impl ZoneEngine {
 
         loop {
             tokio::select! {
+                biased;
+                () = stop.cancelled() => {
```
mattsse, line 168: Cancellation is only polled by this outer `select`. Once `advance_all_available` starts, the engine keeps starting blocks until the queue is empty, so demotion can drain a follower's entire warm queue before stopping. Please pass the token into the drain loop and check it before each new `advance`—not inside `advance`, so the in-flight block still completes. The cancellation test should cancel after one block has started with a backlog remaining and assert that the backlog was not drained.

crates/node/src/node.rs
```diff
@@ -584,8 +776,11 @@ where
         pool: N::Pool,
         engine: ConsensusEngineHandle<ZonePayloadTypes>,
         l1_block_tracker: L1BlockTracker,
+        deposit_queue: DepositQueue,
     ) -> eyre::Result<()> {
-        let role = config.role();
+        let local_ed25519_public_key = config.ed25519_public_key();
+        let leadership = config.leadership();
+        let role = leadership.role_of(&local_ed25519_public_key);
```
mattsse, line 612: This snapshots the bootstrap role for the lifetime of the node-side P2P workers. The runtime filters follow leadership, but after promotion there is still no block broadcaster, forwarded-transaction importer, or settlement collector; after demotion the existing leader sync loop ignores the new leader's blocks. Please supervise the complete role-specific worker bundle by leadership generation, stopping and awaiting the old bundle before starting the new one. If rotation is intentionally deferred, I would keep `publish` private until that bundle can switch safely.

crates/l1/src/subscriber.rs
```diff
@@ -630,20 +628,29 @@ impl L1Subscriber {
...
-                // Leaders do not gate imports on this tracker. Retain only its
-                // monotonic cursor and let the deposit queue own pending data.
+            // Publish derived L1 state only after the header has been admitted
+            // to the contiguous deposit queue.
+            self.apply_enabled_token_events(&events);
+            self.update_l1_state_anchor(block_number, &invalidated);
+            if appended {
+                self.subscriber_metrics.blocks_enqueued.increment(1);
+            }
+            if !self.config.retain_observations {
```
mattsse, line 704: `retain_observations = false` only prunes the tracker; every finalized block is still retained in `DepositQueue`. A node with neither a sequencer nor a manifest has no `confirm`/`confirm_through` caller, so the queue grows indefinitely. Please restore a sink-less observer path (or make subscriber mode explicit) and add a regression showing that caches continue advancing without queue growth.

crates/node/src/node.rs
```diff
@@ -80,6 +81,189 @@ use zone_sequencer::{
...
+            local_ed25519_public_key,
+            receiver: leadership.subscribe(),
+            _leadership: leadership,
+        }
+    }
+
+    fn is_leader(&self) -> bool {
+        self.receiver
+            .borrow()
+            .role_of(&self.local_ed25519_public_key)
```
mattsse: This ignores `start_block`, so publishing a future-dated record promotes the node immediately—the new test does this with `start_block = 120`. Please either delay publication until the activation boundary and enforce that contract in `publish`, or retain active and pending records and derive the role from the relevant next block. `publish` should also reject non-increasing epochs.

mattsse (changes requested):
The direction looks good, but I think we should close the observer queue regression and make the leadership handoff contract safe before merging. For the transition plumbing, either switch the complete role-specific service graph and honor the activation boundary here, or keep leadership publication private until the follow-up implements those semantics.

crates/node/src/engine.rs
adityapk00: Fixed. Cancellation is now checked before each new advance in the drain loop, while an in-flight block still completes. + test

crates/node/src/node.rs
adityapk00: removed the partial leadership supervisors and restored static bootstrap-role service configuration here. Moved it to the leadership rotating PR

crates/l1/src/subscriber.rs
adityapk00: Fixed. Restored an explicit sink-less observer mode for nodes without a sequencer or P2P manifest. +test

crates/node/src/node.rs
adityapk00: fixed, Leadership is now an immutable bootstrap record and the publication API has been mvoed to the other PR (plus the activation-boundary handling, pending records, and epoch validation )

adityapk00:
> I think we should close the observer queue regression and make the leadership handoff contract safe before merging.

Fixed here.

> For the transition plumbing, ... or keep leadership publication private until the follow-up implements those semantics.

Moved all this to the leader rotation PR

mattsse (approved):
LGTM

PR #838, opened 2026-07-24
docs(spec): anchor bootstrap batch to Tempo

legion2002:
## Summary

- require the bootstrap batch to contain at least two blocks: the canonical genesis block followed by at least one non-genesis block
- apply the normal `advanceTempo`, L1 anchoring, and end-of-batch finalization rules to the non-genesis portion of the bootstrap batch
- remove the unanchored zero-Tempo-input bootstrap exception and update the corresponding critical invariants

## Why

The canonical genesis block contains no transactions, so a genesis-only batch performs no Tempo L1 read. Requiring a non-genesis block in the same batch ensures the first accepted proof imports a finalized Tempo header, proves the portal existed at that L1 state through its non-zero sequencer slot, and binds the portal deposit queue read to the imported state root.

This also aligns the spec with the reference portal: its direct-anchor path rejects a zero EIP-2935 block hash, and the final non-genesis block can execute the mandatory `finalizeWithdrawalBatch` call that the transactionless genesis block cannot.

## Validation

- `git diff --check`
- verified the removed genesis-only and zero-anchor exceptions no longer appear in `specs/spec.md` or `invariants.md`

PR #845, opened 2026-07-24
fix(portal): anchor bootstrap batch to Tempo

legion2002:
## Summary

- submit the bootstrap settlement only after the first non-genesis batch boundary
- anchor the first batch to a real Tempo block and apply normal withdrawal-batch accounting
- build the first settlement transition from the portal's zero tip through canonical genesis and the first non-genesis blocks
- remove the zero-anchor/genesis-only bootstrap mode from the portal and sequencer

## Why

This implements the behavior specified in #838. A genesis-only settlement performs no Tempo read, so it cannot authenticate that the portal existed at the imported L1 state. Including at least one non-genesis block makes the first batch import Tempo, prove the portal state, and execute the normal end-of-batch finalization.

## Stack

Stacked on #725 because the implementation changes its zero-genesis settlement path. Review the diff against `agent/tip1091-zero-genesis`.

## Validation

- `rustup run nightly cargo test -p zone-sequencer` (60 passed)
- `rustup run nightly cargo test -p tempo-zone-contracts` (7 passed)
- `rustup run nightly cargo test -p zone-node --lib` (34 passed)
- `rustup run nightly cargo test -p zone-node --test it --no-run`
- `rustup run nightly cargo clippy -p zone-sequencer -p zone-node -p tempo-zone-contracts --all-targets -- -D warnings`
- `FOUNDRY_HARDFORK=cancun forge build --root specs/ref-impls --no-lint`
- `forge inspect ZonePortal storage-layout --json | jq .`
- `git diff --check`


legion2002:
superseded by current main plus PR #725.

PR #848, opened 2026-07-26
feat(rpc): expose public zone namespace

mattsse:
## Summary

- define the public `zone_` namespace with a reusable `jsonrpsee` server trait in `zone-rpc`
- expose `zone_getZoneInfo` and `zone_getEncryptionKey` on standard RPC transports only when the operator-selected module set includes `eth`
- back the public namespace and authenticated private server with the same `ZoneRpcApi` implementation
- route node-internal Zone reads and sequencer monitoring through a dedicated in-process full `eth` provider instead of a host RPC transport
- bound public RPC startup waits and return a temporary initialization error if setup stalls
- log detailed provider failures server-side while returning sanitized public JSON-RPC errors
- keep `zone_getAuthorizationTokenInfo` exclusive to the authenticated private RPC
- align the Zone RPC specification and cover transport scoping, startup, error sanitization, and public/private metadata behavior

## Why

Zone metadata is currently implemented only behind the authenticated privacy server even though `zone_getZoneInfo` and `zone_getEncryptionKey` are authentication-independent. Registering them through the normal reth add-ons flow makes them available to standard RPC clients while preserving a single implementation for the L1 and Zone lookups.

The dedicated in-process provider keeps settlement-critical reads independent of HTTP, WebSocket, IPC paths and public namespace configuration. Public exposure now respects the operator’s selected `eth` modules, and internal provider details cannot leak through public error responses.

This is an alternative to #842 that centralizes the namespace in `zone-rpc` and covers non-HTTP node configurations without relying on externally configured transports.


mattsse:
cyclops audit fast

PR #851, opened 2026-07-26
fix(l1): prefetch deposit state before payload builds

mattsse:
## Motivation

`advanceTempo` executes deposits sequentially. Although the Portal events reveal most of the L1 state that execution will need, cache misses were left to synchronous payload execution. A finalized L1 block containing deposits for many new recipients could therefore force a serial sequence of L1 RPC reads on the payload-builder critical path.

## What changed

- Build an event-derived prefetch plan while regular and encrypted deposits are decoded.
- Populate the existing exact-block L1 cache before sending the FCU that starts payload construction.
- Resolve TIP-403 dependencies in four bounded, deduplicated waves: Portal/token/receive roots, policy bases and receive memberships, simple memberships or compound records, then compound mint subpolicies.
- Reuse the operator-configured `l1.fetch-concurrency` limit for storage requests.
- Avoid speculative work for failed encrypted decryptions and recipients rejected before policy execution.
- Preserve partial token-policy prefetching for virtual recipients and withdrawal bounce-backs while documenting the Zone-local state that prevents recipient-specific prefetching.

## Impact

This moves predictable L1 latency ahead of payload construction without changing policy enforcement or Engine API payloads. Execution consumes the same shared, block-versioned cache at the same finalized child anchor. Prefetch failures occur before FCU, leaving the queued L1 block available for retry.


crates/node/src/engine.rs
```diff
@@ -177,13 +191,25 @@ impl ZoneEngine {
...
+            self.chain_spec.is_t6_active_at_timestamp(timestamp),
+            self.chain_spec.is_t9_active_at_timestamp(timestamp),
+            self.l1_fetch_concurrency,
+        );
+        let (prepared, prefetch) = l1_block
+            .prepare_for_build(&self.sequencer_key, self.portal_address)
+            .await?;
+        prefetch
+            .prefetch(&self.l1_state_provider, prefetch_config)
+            .await?;
```
0xrusowsky, line 256: should prefetching success be "optional" and have regular execution (synchronous) behavior as a fallback?

0xrusowsky: should we do the same optimization for non-leader sequencers in a follow-up PR? i guess otherwise they can fall behind?

crates/l1/src/state/prefetch.rs
```diff
@@ -0,0 +1,667 @@
...
+            .await
+            .wrap_err_with(|| {
+                format!(
+                    "failed to prefetch deposit L1 state at block {}",
+                    self.block_number
+                )
+            })
+    }
+
+    fn root_slots(&self, config: DepositPrefetchConfig) -> Result<HashSet<StorageSlot>> {
```
0xrusowsky, line 220: this seems fine for now, but down the line it would be nice to upstream the semantic coupling to upstream tempo. I think ideally the precompile itself can provide an api for slot updates (will think about it and make a proposal in a follow-up PR)

0xrusowsky (commented):
LGTM, just some minor comments

would be nice to get @klkvr's review too

crates/node/src/engine.rs
```diff
@@ -177,13 +191,25 @@ impl ZoneEngine {
...
+    /// Mint-recipient policy remains enforced by upstream TIP-20 execution. Prefetching only moves
+    /// event-derived RPC reads ahead of the payload-builder deadline and populates the same cache
+    /// that execution uses.
     async fn prepare_l1_block(&self, l1_block: L1BlockDeposits) -> eyre::Result<PreparedL1Block> {
-        l1_block
-            .prepare(&self.sequencer_key, self.portal_address)
-            .await
+        let timestamp = l1_block.header.timestamp();
+        let prefetch_config = DepositPrefetchConfig::new(
+            self.chain_spec.is_t6_active_at_timestamp(timestamp),
```
0xrusowsky, line 247: this will always be true, so we can drop it from the config

PR #854, opened 2026-07-27
refactor: simplify deposits queue

klkvr:
Removes reorg handling logic from queue which is not needed now that `L1Subscriber` is only processing finalized blocks


rkrasiuk (dismissed):

0xrusowsky (commented):
is it a potential footgun that the subscriber records the block before enqueueing it?

> If the queue rejects block N+1, the cache and tracker will already have advanced while the queue remains at N. Since the reconnect cursor considers the block tracker, the subscriber could resume from N+2, leaving the queue behind and unable to catch up.

should the subscriber validate queue continuity before publishing the cache/tracker updates instead?

0xrusowsky (approved):

PR #855, opened 2026-07-27
feat(l1): use storage-root changes for cache invalidation

0xrusowsky:
supersedes #806 

## Motivation

the L1 state cache currently relies on contract events to invalidate cached storage. This is error-prone if an L1-anchored contract updates state without emitting an event, or if Zones does not recognize a newly added event.

## Solution

this PR fetches and verifies account proofs for the zone portal and the TIP-403 registry against the header state root of every finalized L1 block. It tracks each account’s latest storage root and invalidates its cached slots whenever that root changes.

> [!NOTE]
> - This approach adds 2 concurrent `eth_getProof` requests (1 per account) per finalized L1 block. Needs benchmarking but hasn't been done yet.
> - It will require L1 nodes to be run with `--rpc.eth-proof-window <n>`


crates/l1/src/subscriber.rs
```diff
@@ -920,15 +884,79 @@ impl L1Subscriber {
...
+async fn fetch_and_verify_storage_roots<P: Provider<TempoNetwork>>(
+    provider: &P,
+    block: NumHash,
+    state_root: B256,
+    accounts: &[Address],
+) -> eyre::Result<impl Iterator<Item = (Address, Option<B256>)> + use<P>> {
+    let tasks = accounts.iter().map(|&address| async move {
+        match provider
+            .get_proof(address, Vec::new())
+            .block_id(BlockId::hash(block.hash))
```
rkrasiuk, line 912: what's the furthest in L1 history can this target block be? we need to make an informed decision regarding setting the `max-proof-window` on L1 nodes that would enable retrieving historical proofs

crates/l1/src/subscriber.rs
```diff
@@ -920,15 +884,79 @@ impl L1Subscriber {
...
+                    "L1 account proof unavailable"
+                );
+                Ok((address, None))
+            }
+        }
+    });
+    Ok(futures::future::try_join_all(tasks).await?.into_iter())
+}
+
+fn verify_account_storage_root(
```
rkrasiuk, line 936: you can do `reth_trie_common::AccountProof::from_eip1186_proof(proof.clone()).verify(state_root)` instead of this manual verification

crates/l1/src/subscriber.rs
```diff
@@ -920,15 +884,79 @@ impl L1Subscriber {
...
-            .invalidate_and_set_anchor(number, invalidated_accounts.iter().copied());
+            .set_anchor_with_storage_roots(number, storage_roots);
     }
 }
 
+/// Fetch and authenticate tracked account storage roots against a finalized header.
+///
+/// RPC failures yield a missing root, which makes the cache break inheritance for that account.
+/// A proof that is returned but fails authentication is fatal.
+async fn fetch_and_verify_storage_roots<P: Provider<TempoNetwork>>(
```
rkrasiuk, line 901: this might get slow depending on the target `block`, let's add a metric tracking the latency here

rkrasiuk (changes requested):

crates/l1/src/subscriber.rs
0xrusowsky: addressed in [`0b49629` (this PR)](https://github.com/tempoxyz/zones/pull/855/commits/0b4962963c43be8e517d9ed8f2647ea094f9b85d)

crates/l1/src/subscriber.rs
0xrusowsky: addressed in [`0b49629` (this PR)](https://github.com/tempoxyz/zones/pull/855/commits/0b4962963c43be8e517d9ed8f2647ea094f9b85d)

rkrasiuk: nit: this can be inlined now

rkrasiuk (approved):

0xrusowsky:
blocked by https://github.com/tempoxyz/zones/pull/981

PR #857, opened 2026-07-27
fix(evm): scope account-indexed getters

mattsse:
## Summary

- restrict AccountKeychain account getters, NonceManager.getNonce, and TIP20.nonces to the relevant account or current sequencer
- keep authorization deterministic from the consensus-committed Zone block beneficiary
- cover authenticated direct and Multicall3 reads, keychain token authentication, malformed calldata, and zero-beneficiary behavior

## Context

[ZONE-97](https://linear.app/tempoxyz/issue/ZONE-97/authenticated-rpc-callers-can-read-other-users-private-account) remained partially applicable after #805: that change protected ZoneInbox refunds, while AccountKeychain, NonceManager, and TIP20 nonce getters still exposed private metadata to authenticated private-RPC callers.

The root cause was that these getters inherited upstream read behavior without Zone execution-level caller policy. RPC-only filtering would not be sufficient because authenticated callers can forward calls through Multicall3 or arbitrary contracts. This change applies authorization at the precompile/runtime boundary, where the effective EVM caller is known.

Permit2 is intentionally excluded: it is installed in the bundled genesis but is not an expected Zone integration, so this PR does not add runtime policy for unused Permit2 state.


mattsse:
cyclops audit fast

mattsse:
cyclops audit fast

mattsse:
cyclops audit fast

legion2002:
cyclops audit fast

crates/precompiles/src/account_privacy.rs
```diff
@@ -0,0 +1,30 @@
...
+pub(crate) struct AccountPrivacy {
+    current_sequencer: Address,
+}
+
+impl AccountPrivacy {
+    pub(crate) const fn new(current_sequencer: Address) -> Self {
+        Self { current_sequencer }
+    }
+
+    pub(crate) fn authorize(&self, caller: Address, accounts: &[Address]) -> CallCheck {
```
legion2002, line 21: nit: (feel free to ignore) can we have some docs comments here about the behaviour of authorized, or call it something clearer like `is_authorized`

legion2002 (dismissed):
lgtm, might need it for precompiles like the TIP 20 escrow when we enable them.
we should probably standardize this state access restriction pattern somewhere for full zones, because I am assuming all DeFi contracts will face the same issue

PR #860, opened 2026-07-28
feat: Add setLeader RPC/ZonePortal methods to allow leader rotation

adityapk00:
- ZonePortal now keeps track of who the leader is (along with the full list of nodes)
- Add a new setLeader() ZonePortal method that allows the operator to manually set the leader
- Calling the RPC on any/all nodes causes them to rotate the leadership, with the current leader becoming a follower and the target node being promoted to a leader.


specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -317,6 +342,29 @@ contract ZonePortal is IZonePortal {
...
+        // One distinct leader per Tempo block keeps exactly one authorized producer for the
+        // corresponding zone block.
+        if (leaderActivationTempoBlock == uint64(block.number)) {
+            revert LeaderAlreadyUpdatedThisBlock();
+        }
+
+        address previous = leader;
+        leader = newLeader;
+        leaderEpoch += 1;
+        leaderActivationTempoBlock = uint64(block.number);
```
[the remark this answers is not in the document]
adityapk00: This is out of scope for this PR. Recovery from a crashed leader (a follower promotes itself without co-operation from the leader) needs a slightly different flow, and it will be added in a subsequent PR

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -317,6 +342,29 @@ contract ZonePortal is IZonePortal {
         return _sequencers[index];
     }
 
+    /// @inheritdoc IZonePortal
+    function setLeader(address newLeader, uint64 expectedEpoch) external onlySequencer {
```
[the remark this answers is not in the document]
adityapk00: Yes, this is intended. The RPC is inside out infra, so no public access, callable only by the operator

crates/node/src/replication.rs
```diff
@@ -598,13 +665,18 @@ async fn run_follower_block_sync<P>(
...
                     }
-                    P2pEvent::BlockReceived { block, .. }
-                    | P2pEvent::BackfillBlockReceived { block, .. } => {
+                    P2pEvent::BlockReceived { .. }
+                    | P2pEvent::BackfillBlockReceived { .. } => {
+                        let (block, live_sender) = match event {
+                            P2pEvent::BlockReceived { leader_ed25519_public_key, block } => {
+                                (block, Some(leader_ed25519_public_key))
+                            }
+                            P2pEvent::BackfillBlockReceived { block, .. } => (block, None),
```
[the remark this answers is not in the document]
adityapk00: This is intentional to simplify the backfill, since backfilled blocks can be served by anyone. In any case, the prover/verifier will ultimately be responsible for correctness

crates/node/src/role.rs
```diff
@@ -0,0 +1,918 @@
...
+            reasons.push(format!(
+                "peer {peer} advertises tip {} above the local head {local_head}",
+                tip.zone_height
+            ));
+            continue;
+        }
+        match provider.sealed_header(tip.zone_height) {
+            Ok(Some(header)) if header.hash() == tip.zone_hash => {}
+            Ok(Some(header)) => {
+                return Readiness::Conflicted(format!(
```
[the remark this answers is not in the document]
adityapk00: All peers are verified by default (i.e., we will run all nodes, so no external actors can join). A sequencer key compromise is the only way a peer can publish a wrong PeerTip, but if the key is compromised, they can stall the chain anyway.

crates/rpc/src/types.rs
```diff
@@ -225,6 +352,10 @@ pub fn classify_method(method: &str) -> Option<MethodTier> {
         | "web3_sha3"
         | "zone_getAuthorizationTokenInfo"
         | "zone_getZoneInfo"
+        | "zone_getSequencerInfo"
+        // Authenticated operator mutation: the handler additionally requires the caller to
+        // be the portal admin. Deliberately outside the wildcard-rejected admin_* namespace.
+        | "zone_setLeader"
```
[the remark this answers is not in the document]
adityapk00: This is designed intentionally for the operator to be able to call this. Note that this is not exposed to the internet, so the rpc is only callable by the admin

crates/p2p/src/runtime.rs
```diff
@@ -835,10 +885,8 @@ async fn run_receivers(
             // Got backfill response (for an existing request)
             result = backfill_responses.recv() => {
                 let (peer, bytes) = result.map_err(|err| eyre::eyre!("backfill response receive failed: {err}"))?;
-                let (role, leader) =
-                    local_role_and_leader(&leadership, &local_ed25519_public_key);
-                let eligible = match role { Role::Leader => peer != leader, Role::Follower => peer == leader };
-                if !eligible {
+                // Backfill responses are accepted from any node
```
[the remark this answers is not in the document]
adityapk00: All peers are trusted, we'll run them all. A compromised peer (key compromise) can stall the zone anyway, so we won't fix it here for now

crates/node/src/replication.rs
```diff
@@ -829,6 +908,29 @@ where
         .tempo_num_hash()?;
     validate_l1_checkpoint_transition(&l1_header, local.number, local.hash, block_number)?;
     let anchor = l1_header.num_hash();
+
+    // Anchor-aware fence for live blocks: the  sender must
+    // be the scheduled leader of the block's anchor. This stops an honest stale
+    // leader's broadcast from splitting followers.
+    // Backfilled blocks carry no producer claim and are judged by parent/anchor/execution/conflict
+    // validation only.
+    if let Some(sender) = &peer_block.live_sender {
```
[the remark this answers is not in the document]

crates/node/src/role.rs
```diff
@@ -0,0 +1,918 @@
...
+            }
+            continue;
+        }
+        let Some(sink) = sinks.sink_for(&event) else {
+            metrics::counter!("zone_leadership_events_dropped_between_generations_total")
+                .increment(1);
+            debug!(target: "zone::role", "Dropping P2P event with no active generation consumer");
+            continue;
+        };
+        if sink.send(event).await.is_err() {
```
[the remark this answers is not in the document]
adityapk00: fixed

crates/node/src/role.rs
```diff
@@ -0,0 +1,1000 @@
...
+        {
+            let next_anchor = match context
+                .provider
+                .latest()
+                .map_err(eyre::Report::from)
+                .and_then(|state| state.tempo_block_number().map_err(eyre::Report::from))
+            {
+                Ok(checkpoint) => checkpoint.saturating_add(1),
+                Err(_) => 0,
+            };
```
legion2002: `desired_role` already did this exact read a few lines up (:295-306) and discarded the result, so we're paying two MDBX txns + two storage reads per control-loop tick.

The part that worries me more is that the two copies disagree on failure: `desired_role` fences on error, this one falls back to `0`. A transient provider hiccup here silently evaluates the promotion barrier against anchor 0 rather than the real next anchor.

Suggest having `desired_role` return `(DesiredRole, u64 /* next_anchor */)` — one read, one error policy, and the fallback disappears.

crates/p2p/src/manifest.rs
```diff
@@ -102,6 +73,263 @@ impl LeadershipState {
...
+                .transitions
+                .range(..active)
+                .next_back()
+                .map_or(active, |(&predecessor, _)| predecessor);
+            state
+                .transitions
+                .retain(|&activation, _| activation >= keep_from);
+        }
+        drop(state);
+        self.changed.send_replace(());
```
legion2002: This fires unconditionally, and both writers call `record_applied_anchor` on every applied block (`engine.rs:410` on production, `replication.rs:970` on import). So the role controller's `schedule.subscribe()` arm wakes once per block, and every wake does a state read via `desired_role` — during catch-up that's the block-apply rate, not 2/sec.

Suggest firing only when the pruning above actually moved the record governing the next anchor.

Worth pairing with the other half of this: `ROLE_RECHECK_INTERVAL` (`role.rs:51`) exists only because `PeerTipRegistry` has no change signal — even though `LeadershipSchedule` already has exactly the `watch` shape the registry needs. Give the registry the same treatment and the 500ms tick can go entirely. Right now it's a floor on handoff latency, which is the thing this PR is trying to make fast.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -317,6 +342,29 @@ contract ZonePortal is IZonePortal {
...
+        // corresponding zone block.
+        if (leaderActivationTempoBlock == uint64(block.number)) {
+            revert LeaderAlreadyUpdatedThisBlock();
+        }
+
+        address previous = leader;
+        leader = newLeader;
+        leaderEpoch += 1;
+        leaderActivationTempoBlock = uint64(block.number);
+        emit LeaderUpdated(previous, newLeader, leaderEpoch, uint64(block.number));
```
legion2002: These four lines are the same transition `initialize` performs at :221-224 — the only delta is `= 1` vs `+= 1`, which coincide because `leaderEpoch` starts at 0. That's two write paths to one three-slot invariant, and `crates/l1/src/event.rs` decodes `LeaderUpdated`, so if a field is ever added to a transition and only one path gets updated, the Rust bootstrap read-back breaks quietly.

Suggest a private `_setLeader(address newLeader)` alongside the existing `_replaceSequencerSet` / `_setRole` helpers. `initialize` then gets `previous == address(0)` and `epoch == 1` for free, and `setLeader` keeps only its four guards.

Related, at :292-304 — `retainsLeader` is initialised to `leader == address(0)`, so the flag carries two meanings at once ("vacuously fine during initialize" and "found in the new set"), and the loop SLOADs `leader` once per candidate. A single check after the `isSequencer` rewrite states the rule directly and drops both:

```solidity
if (leader != address(0) && !isSequencer[leader]) revert ActiveLeaderRemoved();
```

legion2002 (commented):
A few code-quality notes from a simplify pass (no correctness overlap with the Cyclops review). Three inline comments below — the two Rust ones are about the control loop, the Solidity one is about the leadership write path.

crates/node/src/role.rs
adityapk00: fixed

crates/p2p/src/manifest.rs
adityapk00: fixed

specs/ref-impls/src/tempo/ZonePortal.sol
adityapk00: fixed

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -202,6 +215,10 @@ contract ZonePortal is IZonePortal {
         emit EnforcementModesUpdated(accessEnforced, gatewayEnforced);
 
         _replaceSequencerSet(initialSequencers, _threshold, false);
+        // The first sequencer bootstraps leadership so a fresh zone has a producer without a
+        // separate setLeader call. The creation block is replayed by every zone node because
+        // zone genesis anchors before createZone.
+        _setLeader(initialSequencers[0]);
```
legion2002, line 221: noting we'd also need updates to the 1091 factory, since it can't call initialize directly

adityapk00: Draft PR for tempo to be merged after this one: https://github.com/tempoxyz/tempo/pull/6987

crates/node/src/rpc.rs
```diff
@@ -45,19 +47,207 @@ use tokio::{
...
+    portal_address: Address,
+    sequencer: Arc<std::sync::OnceLock<SequencerRpcContext>>,
+    provider: P,
+) -> Result<RpcModule<()>, jsonrpsee::core::RegisterMethodError>
+where
+    P: BlockNumReader + StateProviderFactory + Clone + Send + Sync + 'static,
+{
+    let mut module = RpcModule::new(());
+    let set_leader_sequencer = sequencer.clone();
+    module.register_async_method("zone_setLeader", move |params, _, _| {
```
0xKitsune, line 125: `zone_setLeader` is registered on the unauthenticated public HTTP module and signs the L1 transaction with the node’s sequencer key. Since follower RPC may be publicly exposed, any caller that can reach it can rotate leadership among registered members.

crates/node/src/replication.rs
```diff
@@ -587,8 +666,13 @@ async fn run_follower_block_sync<P>(
...
+                            // Return the signed settlement attestation to the peer that
+                            // proposed it. During a scheduled handoff that is the outgoing
+                            // leader, not the most recently observed one.
+                            commands.send(P2pCommand::SendSettlementSignature {
+                                leader: leader.clone(),
+                                signature: signed.encode(),
+                            })
                                 .await
                                 .wrap_err("P2P command channel closed")?;
                             Ok::<_, eyre::Report>(height)
```
0xKitsune, line 678: Backfill requests go to every peer, but this branch discards the responder identity and sets `live_sender` to `None`. The first response for a height bypasses scheduled leader validation, so a faulty follower can race a different execution valid block and leave the recovering node on a conflicting canonical chain.

crates/node/src/node.rs
```diff
@@ -586,6 +740,114 @@ where
...
+            snapshot_anchor,
+            "Portal is not deployed at the local Tempo anchor; leadership stays fenced until \
+             the creation block replays"
+        );
+        return Ok(());
+    }
+
+    let portal = ZonePortal::new(portal_address, l1_provider);
+    let leader = portal.leader().block(block_id).call().await?;
+    eyre::ensure!(
```
0xKitsune, line 817: An upgraded portal keeps `leader` at zero until the first `setLeader` transaction. This check exits the node before its RPC starts, so `zone_setLeader` cannot initialize the portal through the node.

crates/node/src/role.rs
```diff
@@ -0,0 +1,1007 @@
...
+                .l1_transaction_signer
+                .clone()
+                .unwrap_or_else(|| sequencer.config.sequencer_signer.clone());
+            let zone_provider = context.provider.clone();
+            let sequencer_token = token.clone();
+            tasks.spawn(async move {
+                let handle = spawn_zone_sequencer(
+                    sequencer_config,
+                    signer,
+                    zone_provider,
```
0xKitsune, line 908: `spawn_zone_sequencer` starts two long running tasks and waits for both with `tokio::join!`. If one panics, the join keeps waiting for the other task, so the role controller never observes the panic or restarts the failed task.

crates/node/src/role.rs
```diff
@@ -0,0 +1,1007 @@
...
+                .best_block_number()
+                .map_err(eyre::Report::from)
+                .and_then(|number| {
+                    context
+                        .provider
+                        .sealed_header(number)?
+                        .ok_or_else(|| eyre::eyre!("no latest block header"))
+                }) {
+                Ok(last_header) => {
+                    let engine = build_engine(context, sequencer, last_header);
```
0xKitsune: If this head read fails, engine creation is skipped but the rest of the leader generation still starts and the node is reported as `Leader`. Later checks see that a leader generation already exists, so a temporary read failure leaves the node unable to produce blocks.

crates/l1/src/subscriber.rs
```diff
@@ -677,19 +691,39 @@ impl L1Subscriber {
 
         while let Some((header, receipts)) = fetched.try_next().await? {
             let block_number = header.number();
-            let (events, invalidated) = self.extract_events(block_number, &receipts);
+            // Decoding fails closed: a decode failure of a recognized portal log aborts this
+            // block before anything is enqueued or any cache advances. Ingestion halts here (fenced)
+            // instead of continuing under a partial event view.
+            let (events, invalidated) =
```
0xKitsune, line 754: `push_log` and `apply_leader_transition` failures enter the generic subscriber reconnect loop after receipts have been verified. The same finalized block is retried forever, blocking all later deposits and portal events.

crates/node/src/rpc.rs
adityapk00: This was specifically requested by infra, so that it works with the operator. I'll send you the discussion on DM

crates/node/src/replication.rs
adityapk00: We decided not to do these validations during the p2p layer because this can only happen under the sequencer key compromise (i.e., a hacked peer), but a hacked peer can stall the chain anyway.

crates/node/src/node.rs
adityapk00: > so zone_setLeader cannot initialize the portal through the node.

Yes, this is intentional. When a zone is created, we want the admin to set the leader first (while creating the zone) so that the very first block is built by the appointed leader (We don't want the operator to call the zoneportal to start a zone)

crates/node/src/replication.rs
adityapk00: For eg, if a crashed leader is rotated out, then a new follower node can never backfill that block because the leader that built the block is no longer around to authenticate that block. That's why backfilled blocks dont carry signatures, and we work under that trust assumumption

crates/node/src/role.rs
adityapk00: fixed

crates/node/src/role.rs
adityapk00: fixed. it will fail the startup and retry in case of temporary network failres

crates/l1/src/subscriber.rs
adityapk00: this is intentional but perhaps misleadingly implemented. a failure to apply the leadership transition should error the node, since we. don't want a partial view of L1, but the restartable errors should be retried. fixed.

mattsse (approved):

PR #865, opened 2026-07-28
feat(multi-seq): recover a restarting leader before block production

legion2002:
## What

A leader that crashed or lost its disk now catches up to the canonical chain before it starts producing blocks. Previously only followers were backfilled.

## How

Leaders get a startup recovery phase. During it the leader behaves like a follower: it probes peers for the blocks it is missing and imports them through the same validating path.

Block production, block broadcast, and settlement collection wait on a `watch` barrier and re-read the head once recovery finishes, so none of them acts on the stale head the process came up with. Node launch itself does **not** wait — recovery imports through the engine API, which reth only serves after `launch_add_ons` returns, so awaiting it during launch would deadlock.

Recovery ends when the local head covers every tip a peer advertised:

- No peer answers within a 10s grace period → start anyway, with a warning. A silent network is indistinguishable from a brand new zone.
- Once any peer answers, its tip is authoritative — the leader keeps retrying rather than forking the zone.

Follower replication and leader recovery now share one block-absorption path instead of two copies of the buffer/drain/refresh logic.

## Testing

`cargo test -p zone-node --lib`, clippy, fmt. `e2e::test_p2p_follower_tracks_leader_balance` exercises leader startup against a real follower, plus new unit tests for the catch-up decision table and recovery completion.

Not covered: a true "leader restarts with a wiped datadir" e2e — the node harness builds a fresh in-memory node per launch and has no relaunch-with-state path.

### DIFF
--- a/crates/node/src/node.rs
+++ b/crates/node/src/node.rs
@@ -5,7 +5,10 @@
 
 use crate::{
     ZoneEngine,
-    replication::{AttestationContext, broadcast_persisted_blocks, run_block_sync},
+    replication::{
+        AttestationContext, LeaderRecovery, after_leader_recovery, broadcast_persisted_blocks,
+        run_follower_block_sync, run_leader_block_sync,
+    },
     rpc::{ZoneRpc, ZoneRpcApi, rpc_connection_config, start_private_rpc},
     settlement_attestation::collect_leader_settlements,
     tx_forwarding::{forward_new_transactions, insert_forwarded_transactions, route_p2p_events},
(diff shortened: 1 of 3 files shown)


legion2002:
cyclops audit

legion2002:
cyclops audit

crates/node/src/node.rs
```diff
@@ -788,32 +821,50 @@ where
...
+        let chain_spec = provider.chain_spec();
+        let engine_handle = ctx.beacon_engine_handle.clone();
+        let payload_builder = ctx.node.payload_builder_handle().clone();
+        let deposit_queue = self.deposit_queue.clone();
+        let block_tracker = self.l1_config.block_tracker.clone();
+        let portal_address = self.portal_address;
+
+        ctx.node.task_executor().spawn_critical_task(
+            "zone-engine",
+            after_leader_recovery(recovery, move || async move {
```
klkvr, line 846: it's a bit annoying that p2p recovery leaks into here

i'm wondering if we could change how things are spawned so that we just await the recovery before spawning engine

klkvr:
superseded by https://github.com/tempoxyz/zones/pull/909, crashing leader should now result in forced recovery

PR #866, opened 2026-07-28
feat(multi-seq): add an rpc-only follower role

legion2002:
## What

Adds a third multi-sequencer role: an **rpc-only follower** that replicates and serves the zone's chain without joining the on-chain settlement quorum, so the leader and quorum followers can stay off the internet.

## How

A manifest node marked `rpc_only = true` resolves to `Role::RpcFollower`.

| | leader | follower | rpc-follower |
|---|---|---|---|
| produces blocks | ✅ | — | — |
| imports replicated blocks | — | ✅ | ✅ |
| serves backfill to peers | ✅ | ✅ | ✅ |
| forwards transactions to the leader | — | ✅ | ✅ |
| receives settlement proposals | — | ✅ | — |
| signs settlement attestations | ✅ | ✅ | — |
| registered with `ZonePortal` | ✅ | ✅ | — |

Standbys are excluded from `block_attestation_addresses`, so the on-chain threshold is unaffected by how many you run. The manifest minimum is now three *quorum* nodes rather than three nodes, and the leader cannot be `rpc_only`. `rpc_only` defaults to false, so existing manifests are unchanged.

A standby backfills from the quorum followers too, not just the leader — it exists to keep serving reads when the leader is unavailable. It will not backfill from another standby.

## Deployment note

An rpc-only node still needs its own Ed25519 and secp256k1 keys and a manifest entry, but its secp256k1 address **must not** be registered with `ZonePortal` — that would add a signer the zone never collects a signature from.

## Testing

`cargo test -p zone-p2p --lib`, `cargo test -p zone-node --lib`, clippy, fmt. New `rpc_only_follower_replicates_but_never_settles` spins up a real four-node mesh and asserts the standby receives live blocks, is never sent a settlement proposal, cannot get a signature to the leader, and can get a forwarded transaction there. Plus unit tests for role derivation, manifest validation, and peer-set routing across all three roles.


legion2002:
cyclops audit

legion2002:
cyclops audit

grandizzy:
cyclops audit

crates/p2p/src/manifest.rs
```diff
@@ -356,14 +408,28 @@ impl ZoneManifest {
 
     /// Static leadership record used for the lifetime of the process.
     pub fn bootstrap_leadership(&self) -> LeadershipState {
-        LeadershipState::new(0, self.leader_ed25519_public_key.clone(), 0)
+        LeadershipState::new(
+            0,
+            self.leader_ed25519_public_key.clone(),
+            0,
+            self.nodes
```
klkvr: unrelated but it seems that `epoch` and `start_block` are always zeros so do we actually need those fields?

crates/p2p/src/runtime.rs
```diff
@@ -757,12 +758,57 @@ async fn run_commands(
...
+        .iter()
+        .filter(|peer| record.role_of(peer) == Role::Follower)
+        .cloned()
+        .collect()
+}
+
+/// Whether `peer` may serve block backfill to a local node holding `role`.
+///
+/// One predicate drives both sending requests and admitting responses, so the two cannot drift.
+fn serves_backfill_for(record: &LeadershipState, role: Role, peer: &PublicKey) -> bool {
```
klkvr: those free functions are a bit hard to follow, i'm wondering if we could have more context kept on `LeadershipState` that'd allow us to do something like `record.quorum_followers()`/`record.replicas()` etc

crates/p2p/src/runtime.rs
```diff
@@ -787,9 +833,9 @@ async fn run_receivers(
             // Got a block
             result = blocks.recv() => {
                 let (peer, bytes) = result.map_err(|err| eyre::eyre!("block channel receive failed: {err}"))?;
-                let (role, leader) =
-                    local_role_and_leader(&leadership, &local_ed25519_public_key);
-                if role == Role::Leader || peer != leader {
+                let (role, record) =
+                    local_leadership(&leadership, &local_ed25519_public_key);
```
klkvr: why not do this once at the beginning of the loop instead of doing it in every branch?

adityapk00:
I think we should not require an rpc-only node to have a secp256k1 key. it's not going to sign anything with this key, so we should not even need it (in the manifest). If such a node ever gets promoted to a leader or follower, the key should be generated then and inserted into the zoneportal.

crates/node/src/cli.rs
```diff
@@ -452,8 +454,7 @@ fn prepend_log_filter(filter: &mut String, directives: &str) {
 
 fn sequencer_enabled(cli_flag: bool, manifest_role: Option<Role>) -> bool {
     match manifest_role {
-        Some(Role::Leader) => true,
-        Some(Role::Follower) => false,
+        Some(role) => role == Role::Leader,
```
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/9ef193bbd2b0356f1fd8951d7bf9c291096924cd

crates/p2p/src/manifest.rs
```diff
@@ -356,14 +408,28 @@ impl ZoneManifest {
...
     }
 
     /// All nodes in the static peer set.
     pub fn nodes(&self) -> &[ManifestNode] {
         &self.nodes
     }
 
+    /// Nodes registered with `ZonePortal` for the on-chain settlement quorum.
+    pub fn quorum_nodes(&self) -> impl Iterator<Item = &ManifestNode> {
+        self.nodes.iter().filter(|node| !node.rpc_only)
```
legion2002: addressed in https://github.com/tempoxyz/zones/commit/dd13b57ec8798c6cbc4af59774e54cd49505616a

crates/p2p/src/runtime.rs
```diff
@@ -757,12 +758,57 @@ async fn run_commands(
...
+    match role {
+        // A recovering leader catches up from the quorum followers. Standbys hold the same chain,
+        // but they are the internet-facing members — the point of the role is to keep chain data
+        // for the leader and the quorum coming from nodes that are not exposed.
+        Role::Leader => peer != &record.leader && record.role_of(peer).in_quorum(),
+        // A follower's only authoritative source is the leader.
+        Role::Follower => peer == &record.leader,
+        // An RPC standby exists for when the leader is unavailable, so it also catches up from
+        // the quorum followers — they hold the same chain and are not exposed to the internet.
+        Role::RpcFollower => record.role_of(peer).in_quorum(),
```
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/73c58dbac3589d026d9b1dfdafa89a7584553ae6

crates/p2p/src/manifest.rs
```diff
@@ -402,6 +468,9 @@ struct RawManifestNode {
     ed25519_public_key: String,
     secp256k1_address: String,
     address: String,
+    /// Serve RPC as a hot standby without joining the on-chain settlement quorum.
+    #[serde(default)]
+    rpc_only: bool,
```
legion2002, line 847: Not taking this for now — reverted in https://github.com/tempoxyz/zones/commit/749e26c8

Nothing is live, so there is no fleet rolling a manifest ahead of its binaries. Strict parsing is worth more meanwhile: with the startup sequencer-set-version check gone, a mistyped `sequencer_set_version` would silently default to 1 and surface only at the first batch boundary.

crates/p2p/src/manifest.rs
legion2002: Addressed after rebase — https://github.com/tempoxyz/zones/commit/0a8a32070d71249f42a635e4bd66bdf3d81f8fab

crates/p2p/src/runtime.rs
legion2002: Addressed after rebase — https://github.com/tempoxyz/zones/commit/0a8a32070d71249f42a635e4bd66bdf3d81f8fab

crates/p2p/src/runtime.rs
legion2002: Addressed after rebase — https://github.com/tempoxyz/zones/commit/0a8a32070d71249f42a635e4bd66bdf3d81f8fab

crates/p2p/src/runtime.rs
```diff
@@ -605,20 +634,45 @@ async fn run_commands(
...
+            // of leaving it open on every request.
+            let (backfill_peers, leader_only) = {
+                let job = backfill_job.lock().await;
+                let leader = leadership.next_anchor_record().map(|record| record.leader);
+                match leader {
+                    Some(leader)
+                        if backfill_peers.contains(&leader)
+                            && !job.is_unresponsive(&leader, now) =>
+                    {
+                        (vec![leader], true)
```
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/zones/commit/87550bb4b67739433b5775074f46e37936ea6630

crates/p2p/src/network.rs
```diff
@@ -97,6 +97,10 @@ pub(crate) fn instantiate(
     network_id: P2pNetworkId,
 ) -> eyre::Result<(Network, Oracle, Map<PublicKey, Address>)> {
     let namespace = namespace(manifest.zone_id(), network_id);
+    // Logged, not bound into the namespace: a mismatch between nodes stalls settlement loudly at
+    // the next batch boundary, and making it a handshake failure would turn every membership edit
+    // into a coordinated fleet restart. Compare this across nodes to diagnose one.
+    tracing::info!(target: "zone::p2p", membership_digest = %manifest.membership_digest(), "Zone P2P membership");
```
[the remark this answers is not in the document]
legion2002: Not doing this — it was bound into the namespace in f57c4fb and deliberately unbound in dd13b57.

The failure you describe is real but it is liveness, not integrity: `ZonePortal` checks `isSequencer[signer]` when verifying a batch, so a manifest disagreement cannot get an unregistered signature counted on-chain. It can only stall settlement, which surfaces at the next batch boundary.

Binding the digest makes that stall a total partition instead: nodes on either side of any membership edit cannot connect at all, so every edit becomes a coordinated fleet restart. Trading a loud stall for an outage plus permanent operational rigidity isn't worth it at this stage.

The README lines you cite were stale — they described the bound version after it had been unbound. Fixed in https://github.com/tempoxyz/zones/commit/bbcb89d0, so the contradiction is gone and the digest is now documented as diagnostic only.

legion2002:
cyclops audit

adityapk00 (approved):
lgtm

crates/node/src/settlement_attestation.rs
```diff
@@ -23,6 +23,79 @@ use zone_p2p::P2pCommand;
...
+///
+/// A quorum node the portal has not registered can never settle, and an unreachable threshold
+/// stalls settlement — both otherwise surface as a mysterious stall at the next batch boundary.
+///
+/// Extra registered signers only warn: deregistering is a cleanup task, and failing on it would
+/// make every membership change a window in which no node can start.
+///
+/// Skipped when no portal is deployed yet, matching `seed_leadership_schedule`: a zone whose
+/// creation block has not replayed has nothing to reconcile against.
+pub(crate) async fn validate_registered_sequencer_set(
```
klkvr, line 36: this seems not really relevant for this PR?

crates/p2p/src/manifest.rs
```diff
@@ -600,11 +712,82 @@ impl ZoneManifest {
...
+            .filter_map(|node| Some((node, node.secp256k1_address?)))
+    }
+
+    /// Digest of the settlement-relevant membership, logged at startup.
+    ///
+    /// Covers each member's Ed25519 identity, quorum standing, and the address its signatures must
+    /// recover to — everything two nodes must agree on to derive the same roles. Compare it across
+    /// nodes to diagnose a manifest mismatch. Addresses are excluded so relocating a node does not
+    /// change it. Sorted, so file order does not matter.
+    pub fn membership_digest(&self) -> B256 {
```
klkvr, line 752: this as well?

crates/p2p/src/runtime.rs
```diff
@@ -649,6 +648,71 @@ async fn run_commands(
                 .cloned()
                 .collect::<Vec<_>>()
         };
+        if let P2pCommand::RequestBackfill { start } = command {
```
klkvr, line 651: imo this could be done in a separate PR, if we want to rewrite logic for how backfill is done, it doesn't have to be done here

klkvr (commented):
overall looking good but the pr looks quite bloated without a strong reason imo

PR #871, opened 2026-07-28
ci: sync Zone runtimes to Tempo

0xrusowsky:
Compiles the canonical `ZonePortal` and `ZoneMessenger` after each push to `main` and compares their executable bytecode with Tempo. On drift, it opens or refreshes the corresponding Tempo PR and fails the Zones post-merge check with a link to that PR, keeping the mismatch visible until Tempo is aligned and the check is rerun.

Before T9 is scheduled, drift updates the existing runtime constants in a normal Tempo sync PR. Once `t9Time` appears in Presto, it preserves the frozen constants, adds `_TN` candidates with hardfork TODOs, and opens a draft `[DRAFT] feat(hardfork): stage updated Zone runtimes` PR.

Verified the Portal sync path end-to-end with an intentionally stale runtime in tempoxyz/tempo#6980, including cross-repo authentication, bytecode correction, branch push, and PR refresh. Also tested the scheduled-T9 path locally and verified both current Portal and Messenger executables reproduce Tempo exactly; workflow run: https://github.com/tempoxyz/zones/actions/runs/30353025076


legion2002:
I don't think we can just do this directly, the bytecode needs to be gated behind a new hardfork after T9, and the old bytecode needs to be retained.

0xrusowsky:
> I don't think we can just do this directly, the bytecode needs to be gated behind a new hardfork after T9, and the old bytecode needs to be retained.

it will now open the PR as draft and add the new bytecode commented out in a TODO, so that we can easily do it

legion2002 (approved):
think we need some way to notify the person who made the zones repo, so they don't forget to update the tempo repo/PR.

But let's test out the workflow with this merged, and update with feedback later

PR #875, opened 2026-07-28
refactor: simplify leader rotation control paths

legion2002:
Stacked on #860 — **base branch is `aditya/zero_downtime`, not `main`.** Rebased onto `1e92ec19`.

Quality-only pass over the leader-rotation code (reuse, simplification, efficiency, altitude). Deliberately no overlap with the consensus/liveness/authorization review already on #860.

Net: **+261 / −268** across 11 files.

> **Rebase note.** `1e92ec19` ("simplify role polling, advace, portal") independently implemented three of the findings this PR originally carried — the `desired_role` anchor plumbing, the notify-on-real-change fix in `record_applied_anchor`, and the Solidity `_setLeader` / `retainsLeader` cleanup. In each case I dropped my version and kept upstream's, which is better on all three counts: `next_anchor` lives inside the `DesiredRole::Leader` variant (no `Option` + `expect`), `record_applied_anchor` compares the whole governing record rather than just its activation, and `_setLeader` caches `block.number` in a local. `1e92ec19` also went further than I did and removed `ROLE_RECHECK_INTERVAL` outright by adding a `watch` to `PeerTipRegistry` — a change I had explicitly deferred as needing its own review. What remains below is only what upstream did not already do.

### `role.rs`
- Delete `GenerationKind`. It was `DesiredRole` with the epoch removed, and it cost five hand-written matches inside one 50-line stretch — role name, epoch (twice, identically), metric label, and gauge value. Replaced with `name()` / `gauge()` / `epoch()` / `same_variant()`. This got *more* valuable after `1e92ec19`, which had to add `{ epoch, .. }` patterns in three places to work around the redundant enum.
- Derive `ready_for_promotion` from `promotion_reasons.is_empty()` instead of a separate field every write site had to keep consistent with the reasons.
- Replace the inner `mod zone_transaction_pool_alias` with a plain `use` — `tempo-transaction-pool` is already a direct dependency and `tx_forwarding.rs` imports the same path directly.

### `manifest.rs`
- `latest_observed_epoch` is derived rather than maintained. `publish` enforces `epoch == last.epoch + 1` and pruning never drops the last entry, so the field could only ever equal the last retained transition's epoch.
- Delete `role_for` — no production callers, only its own unit test.
- Add the rationale comments to upstream's `record_applied_anchor` (why a plain advance must stay silent).

### `runtime.rs`
- Hoist the recipient set out of the command loop. It used to be `followers()`, whose predicate depended on the mutable leader; it is now "all peers except self", and both inputs are fixed for the task's lifetime.
- Move the `RequestBackfill` arm back into its `match`. The earlier hoist left `unreachable!("backfill requests are routed before role derivation")` in a dispatcher that no longer does role derivation, plus a character-identical copy of the peer filter.

### `engine.rs`
- `permit` returns `Option<EngineExit>`. `PermitDecision::{Produce, Demoted, Fenced}` mapped 1:1 onto `EngineExit`, and the `u64` in the returned tuple was always `block.header.number()` — which the caller already has. Deletes the enum, its `lib.rs` re-export, and the translation match.

### `node.rs`
- Extract `ensure_zone_chain_id` (inlined three times, same hand-written message) and `build_sequencer_config` (a 10-field struct literal duplicated between the leader generation and legacy single-sequencer startup — a field added to one silently changed leader-mode behaviour only).
- Batch the three finalized portal reads (`leader`, `leaderEpoch`, `leaderActivationTempoBlock`) into one `try_join!`. Same block, no data dependency, and `set_leader` already used this pattern for two of them.

### `replication.rs`
- Fold the two dead `BackfillRequested` arms into the neighbouring no-op arms. `route_events_to_generations` diverts that variant before any generation sink sees it, so both `debug!` calls were unreachable. Kept as explicit arms rather than deleted so the matches stay exhaustive and the compiler still catches a new variant.

### sequencer
- Replace three copies of the cancel-or-sleep `tokio::select!` (`monitor.rs` ×2, `withdrawals.rs` ×1) with `CancellationToken::run_until_cancelled`, and name the retry interval instead of repeating `Duration::from_secs(5)`.
- Fix a doc comment split across the `#[instrument]` attribute on `WithdrawalProcessor::run`.

### rpc / Solidity
- Drop assertions on `admin_setLeader`, which exists nowhere in the repo. The `admin_*` prefix rule is already covered with a real method (`admin_peers`) in an existing test.
- Document `_setLeader` as the single write path, noting that `crates/l1` decodes `LeaderUpdated` (the reason both callers must go through it).

## Verification

- `forge test` — **166 passed, 0 failed** (`ZonePortal.t.sol`, `ZoneConfig.t.sol`)
- `cargo test -p zone-p2p -p zone-node -p zone-rpc -p zone-sequencer --lib` — **170 passed, 0 failed**
- `cargo clippy` on those four crates, `--all-targets` — clean

Integration tests (`handoff_e2e` and friends) have **not** been run locally — they spin up full reth nodes and are left to CI.

## Not done

Listed so the gaps are explicit rather than implied.

**Not started:**
- **Stringly-typed RPC surface** — `LocalSequencerInfo.role: String`, `mode: "multi"/"single"` (exactly `local.is_some()`; four fields encoding one bit, and a client can observe `mode:"multi"` with `local:null`), `SetLeaderResponse.status` (exactly `tx_hash.is_some()`), `Vec<String>` readiness reasons as an API contract. The largest remaining finding. Worth doing while the API is unreleased, as its own PR.
- `crates/rpc` owns eight response types it never references, while the methods are served from the node's HTTP module.
- The promotion quorum reads `context.attestation.addresses.keys()` rather than `manifest.nodes()` — same set today, silently coupled.
- `OnceLock<SequencerRpcContext>` is never observed empty (`set()` precedes `launch_add_ons_with`), yet emptiness doubles as "single-sequencer mode".
- `publish()`'s clamp-first-activation-to-0 and the three-condition duplicate-epoch special case it then requires.
- `--sequencer.role` asserts against the manifest's `leader_ed25519_public_key`, which production code no longer consults (`P2pConfig::new` builds an `uninitialized()` schedule). `crates/p2p/README.md:25` still describes it as a live bootstrap.
- Efficiency items: backfill tip advertisement recomputed on empty pages; `EventSinks::sink_for` taking a mutex + `Sender` clone per P2P event; `PeerTipRegistry::snapshot` allocations and duplicate header lookups per tick.
- `seed_leadership_schedule` duplicates the portal-deployment probe and the leader-resolution tail that `ScheduleLeadershipSink` already has.

**Partial:**
- Dead API: `role_for` and the phantom test are gone; `is_initialized` / `seeded` (test-only) and the `LeadershipState` accessors remain.
- Startup: the `try_join!` landed; the eagerly-connected `relayer` provider — used only by `zone_setLeader` — still blocks startup on every node.

**Judged not worth it:** the nested `match` + `unreachable!` at `replication.rs:679` (removing it properly means extracting a ~55-line body with many captures); test-helper dedup (`_sequencerSetWithLeader` ×4, the producer-boundary loop ×3 in `handoff_e2e.rs`, the manifest-TOML builder ×5 in `runtime.rs`, `make_receipt_with_logs`); putting the anchor on the `BroadcastBlock` wire frame to retire `is_scheduled_leader`; inlining `sequencer_enabled`, where a named and tested predicate reads better than the bare `||`.

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -5,6 +5,7 @@ use std::{net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 use alloy_primitives::Address;
 use alloy_signer_local::PrivateKeySigner;
 use clap::{Args, CommandFactory, FromArgMatches};
+use reth_chainspec::EthChainSpec;
 use reth_consensus::noop::NoopConsensus;
 use reth_ethereum::cli::Cli;
 use reth_tracing::tracing::info;
@@ -106,6 +107,7 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
 
         validate_l1_rpc_url(&args.l1_rpc_url)?;
         validate_portal_address(args.portal_address)?;
+        args.validate_zone_id(builder.config().chain.genesis().config.chain_id)?;
 
         let p2p_config = args
             .sequencer_manifest
@@ -416,7 +418,7 @@ pub struct ZoneArgs {
     )]
     pub l1_retry_connection_interval_ms: u64,
 
-    /// Zone ID for the private RPC auth token validation.
+    /// Zone ID used for chain identity and private RPC authentication.
     #[arg(long = "zone.id", env = "ZONE_ID", default_value_t = 0)]
     pub zone_id: u32,
 
@@ -444,6 +446,26 @@ pub struct ZoneArgs {
         conflicts_with = "sequencer_manifest"
     )]
     pub enable_sequencer: bool,
+}
+
+impl ZoneArgs {
+    /// Assert the genesis chain ID is the one `zone.id` requires.
+    ///
+    /// `zone_id == 0` means "unset", which imposes no constraint.
+    fn validate_zone_id(&self, chain_id: u64) -> eyre::Result<()> {
+        if self.zone_id == 0 {
+            return Ok(());
+        }
+        let expected = zone_primitives::constants::zone_chain_id(self.zone_id);
+        eyre::ensure!(
+            chain_id == expected,
+            "chain ID mismatch: zone.id={} requires chain_id={}, but genesis has {}",
+            self.zone_id,
+            expected,
+            chain_id,
+        );
+        Ok(())
+    }
 }
 
 fn prepend_log_filter(filter: &mut String, directives: &str) {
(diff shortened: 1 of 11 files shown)


adityapk00 (dismissed):
lgtm

adityapk00 (approved):
Lgtm

adityapk00 (approved):
Lgtm

PR #881, opened 2026-07-29
feat(l1): prefetch deposit state before payload build

0xrusowsky:
Add builder-local L1 cache prewarming for Zone payload construction. Before canonical `advanceTempo`, the builder runs bounded, ordered one-deposit `advanceTempo` simulations to discover and warm the shared L1 cache through the real Inbox, policy, encrypted-deposit, and token-enable paths.

This enables parallel L1-read discovery without duplicating protocol logic or changing canonical execution: `advanceTempo` remains the sole committed system transaction and never waits for prewarming.


0xrusowsky:
cyclops audit fast

PR #882, opened 2026-07-29
fix(precompiles): restrict TIP-403 registry callers

decofe:
## Summary

- reject TIP-403 registry calls from EOAs and arbitrary contracts before L1-backed dispatch
- preserve access for fixed Tempo system precompiles and the existing read-only selector policy
- add regression coverage proving an untrusted caller receives `Unauthorized` before registry execution

Fixes the reachable call path reported in [SIGP-322](https://linear.app/tempoxyz/issue/SIGP-322).

## Validation

- `git diff --check`
- `cargo test -p zone-precompiles tip403_proxy --lib` *(blocked: fetching pinned `alloy-evm` returned HTTP 401 in the sandbox)*

Prompted by: @jenpaff

### DIFF
--- a/crates/precompiles/src/tip403_proxy/mod.rs
+++ b/crates/precompiles/src/tip403_proxy/mod.rs
@@ -8,7 +8,8 @@
 use crate::execution::{CallCheck, CallRules};
 use alloy_primitives::Address;
 use alloy_sol_types::{SolCall, SolError};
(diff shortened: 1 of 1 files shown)


decofe:
Closing because this fix belongs in tempoxyz/tempo; I am moving the change there.

PR #883, opened 2026-07-29
fix(precompiles): forbid external calls to TIP403

0xrusowsky:
ref: `SIGP-322`


0xrusowsky:
cyclops audit fast

0xrusowsky:
> ### 🚨 [SECURITY] TIP-20 `eth_call` still exposes arbitrary TIP-403 L1 reads
> **Severity:** Medium **File:** `crates/precompiles/src/ztip20/mod.rs:99`
> 
> The new TIP-403 guard rejects only calls whose EVM target is `TIP403_REGISTRY_ADDRESS`. Authenticated users can still simulate TIP-20 `transfer`/memo-transfer calls with fresh recipients; `TIP20Rules::admit` falls through to upstream TIP-20, which reads TIP-403 receive-policy slots. On zones, those reads go through the anchored L1 overlay and the default synchronous L1 provider retries cache misses indefinitely, so repeated free simulations can consume L1 RPC and blocking EVM/RPC worker resources.

~~this is a separate issue. Sequencers should not expose public simulation methods.~~
ended up fixing here https://github.com/tempoxyz/zones/pull/891

0xKitsune (dismissed):

0xKitsune:
Regarding https://github.com/tempoxyz/zones/pull/883#pullrequestreview-4807852250

Looks like we will need address the fact that `eth_call`/`eth_estimateGas` is publically callable. We can also do this in a follow up.

crates/node/src/node.rs
```diff
@@ -89,6 +89,21 @@ use zone_sequencer::{
...
+/// Simulations can execute TIP-20 policy checks backed by finalized L1 state. Keeping them on the
+/// authenticated private RPC prevents arbitrary public callers from forcing sequencer L1 reads,
+/// while non-sequencer RPC nodes can continue serving simulations.
+fn configure_public_simulation_methods(modules: &mut TransportRpcModules, is_sequencer: bool) {
+    if is_sequencer {
+        modules.remove_http_methods(PUBLIC_SIMULATION_METHODS);
+        modules.remove_ws_methods(PUBLIC_SIMULATION_METHODS);
+        modules.remove_ipc_methods(PUBLIC_SIMULATION_METHODS);
+    }
+}
```
0xKitsune: minor nit: IMO its a bit cleaner to just make this `fn disable_simulation_methods()` and only call this in the node setup if the sequencer is enabled, rather than passing in the sequencer here. Very minor, feel free to disregard if preferred.

crates/node/src/node.rs
```diff
@@ -89,6 +89,21 @@ use zone_sequencer::{
...
 };
 
+const PUBLIC_SIMULATION_METHODS: [&str; 2] = ["eth_call", "eth_estimateGas"];
+
+/// Remove simulation methods from a sequencer's unauthenticated RPC transports.
+///
+/// Simulations can execute TIP-20 policy checks backed by finalized L1 state. Keeping them on the
+/// authenticated private RPC prevents arbitrary public callers from forcing sequencer L1 reads,
+/// while non-sequencer RPC nodes can continue serving simulations.
+fn configure_public_simulation_methods(modules: &mut TransportRpcModules, is_sequencer: bool) {
```
0xKitsune: Should we add an e2e test asserting that these are disabled when the node is in sequencing mode?

0xKitsune (dismissed):

crates/node/src/node.rs
0xrusowsky: addressed in [`96c1dce` (this PR)](https://github.com/tempoxyz/zones/pull/883/commits/96c1dce6b6d35ef2a909636bd475fe959fcb19bb)

0xKitsune (dismissed):

0xKitsune (approved):

PR #886, opened 2026-07-29
Remove unsafe L1 genesis block-number override

klkvr:
Removes the `--l1.genesis-block-number` / `L1_GENESIS_BLOCK_NUMBER` bootstrap override and its config wiring.

The L1 subscriber now derives its starting height exclusively from the checkpoint persisted in Zone state. An unanchored genesis is rejected instead of trusting an operator-supplied block number or silently starting at the RPC finalized tip.

Breaking impact:
- remove `--l1.genesis-block-number` from launch commands
- `L1_GENESIS_BLOCK_NUMBER` is no longer recognized
- regenerate Zone genesis files that do not embed an L1 checkpoint

Validation:
- `cargo test -p zone-l1 --lib`
- `cargo test -p zone-node --lib`
- `cargo clippy -p zone-l1 --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -287,7 +287,6 @@ zone-up name reset="false" profile="dev" args="":
                       --chain "$GENESIS_JSON" \
                       --l1.rpc-url "${L1_RPC_URL:?Set L1_RPC_URL env var (wss://...)}" \
                       --l1.portal-address "$PORTAL" \
-                      --l1.genesis-block-number "$ANCHOR_BLOCK" \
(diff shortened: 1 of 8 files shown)


rkrasiuk (approved):

rkrasiuk (approved):

rkrasiuk (approved):

rkrasiuk (dismissed):

rkrasiuk (approved):
anotha one

PR #888, opened 2026-07-29
feat(rpc): expose zone metadata on standard transports

mattsse:
Alternative to #848.

Exposes `zone_getZoneInfo` and `zone_getEncryptionKey` on the standard RPC transports with a small provider-backed `ZoneApi`. The public API stays independent of the authenticated `ZoneRpcApi`, while shared portal helpers keep both surfaces aligned.

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -23,7 +23,7 @@ use alloy_rpc_types_eth::{
 use alloy_sol_types::SolCall;
 use eyre::WrapErr;
 use futures::StreamExt;
-use jsonrpsee::{RpcModule, types::ErrorObjectOwned};
+use jsonrpsee::{RpcModule, core::RpcResult, proc_macros::rpc, types::ErrorObjectOwned};
 use reth_provider::CanonStateSubscriptions;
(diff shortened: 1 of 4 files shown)


crates/node/src/rpc.rs
```diff
@@ -146,6 +215,96 @@ fn public_rpc_error(error: JsonRpcError) -> ErrorObjectOwned {
...
+///
+/// The caller supplies the local Zone's processed Tempo block number; the
+/// remaining dynamic fields are read directly from the ZonePortal on Tempo L1.
+async fn zone_info(
+    zone_id: u32,
+    chain_id: u64,
+    portal_address: Address,
+    tempo_block_number: u64,
+    l1_provider: &DynProvider<TempoNetwork>,
+) -> Result<ZoneInfoResponse, JsonRpcError> {
```
0xrusowsky, line 262: isn't it a little bit weird that we pass tempo block num (anchor) and read from the L1 anyways? i guess it doesn't matter much as the config is "static" in practical terms

0xrusowsky (approved):

PR #890, opened 2026-07-29
fix(precompiles): don't leak user balance on reverts

0xrusowsky:
ref: `ZONE-104`

this PR redacts the user balance from a tip20 transfer from `InsufficientBalance(from, to, balance)` error

### DIFF
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -28,14 +28,7 @@
 extern crate alloc;
 
 pub mod error;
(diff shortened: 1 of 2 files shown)


mattsse (approved):

legion2002 (approved):

legion2002:
Noting that this does not solve the binary search issue, also mentioned in 104. An attacker can just binary search across reverts to find the true balance.

Think we'll need some blocker on who can call `transferFrom`, but that's a more invasive change so we can merge this PR for now

PR #892, opened 2026-07-29
fix(rpc): make private fee quotes deterministic

mattsse:
## Summary

- return the fixed T1 gas price from eth_gasPrice
- return zero from eth_maxPriorityFeePerGas
- cover both sequencer and ordinary authenticated callers and document the policy

## Why

The private RPC delegated fee quotes to the Reth gas oracle, allowing recent private transaction tips to influence responses even though eth_feeHistory redacts the same activity. Zones do not accept priority fees, so these endpoints should return caller-independent values instead of sampling private blocks.

Fixes ZONE-195

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -37,7 +37,7 @@ use tempo_alloy::{
     TempoNetwork,
     rpc::{TempoCallBuilderExt as _, TempoHeaderResponse, TempoTransactionRequest},
 };
-use tempo_chainspec::spec::TEMPO_T0_BASE_FEE;
+use tempo_chainspec::spec::{TEMPO_T0_BASE_FEE, TEMPO_T1_BASE_FEE};
 use tempo_contracts::precompiles::{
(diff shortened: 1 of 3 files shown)


crates/node/src/rpc.rs
```diff
@@ -670,19 +670,11 @@ where
     }
 
     fn gas_price(&self) -> BoxFut<'_> {
-        Box::pin(async move {
-            let price = EthFees::gas_price(&self.eth.api).await.map_err(internal)?;
-            to_raw(&price)
-        })
+        Box::pin(async move { to_raw(&U256::from(TEMPO_T0_BASE_FEE)) })
```
0xKitsune: Should this be `TEMPO_T1_BASE_FEE`?

mattsse: Yes — updated `eth_gasPrice` and its regression coverage/spec to use `TEMPO_T1_BASE_FEE` in de879258. The zero priority-fee quote remains unchanged.

Rjected (approved):

PR #896, opened 2026-07-29
fix(node): retry startup settlement recovery errors

0xKitsune:
This PR retains the first settlement boundary that errors during leader startup so it is retried before later boundaries.

Resolves ZONE-208.


crates/node/src/settlement_attestation.rs
```diff
@@ -294,19 +294,8 @@ pub(crate) async fn collect_leader_settlements<P>(
...
-                pending_boundary = Some(number);
-                break;
-            }
-            Ok(false) => {}
-            Err(err) => {
-                debug!(target: "zone::p2p", %err, number, "Skipped non-current settlement boundary during recovery")
-            }
-        }
-    }
+    let mut pending_boundary =
```
adityapk00, line 370: the initial scan now retains any candidate that errors (not just a settlement boundary). if a normal block transiently errors, then returns Ok(false) on retry, line 354 below clears pending_boundary without scanning the remaining persisted range. and because last_scanned was already set to head, a real boundary later in that range is permanently skipped, even after new blocks arrive.

for eg.  block 50 errors transiently, block 100 is the next boundary. Retry of block 50 returns false; blocks 51–100 are never examined.

i think the retry should resume scanning from number + 1 after Ok(false) (or maybe retain a separate scan cursor).

adityapk00 (changes requested):

crates/node/src/settlement_attestation.rs
0xKitsune: Updated in [824a1f1](https://github.com/tempoxyz/zones/pull/896/commits/824a1f196bbe6c6ffc8c07ed50ef5df08a1c3e5e)

crates/node/src/settlement_attestation.rs
```diff
@@ -294,19 +294,8 @@ pub(crate) async fn collect_leader_settlements<P>(
...
+        propose_persisted_settlement_range(&provider, &commands, &context, 1, head).await;
```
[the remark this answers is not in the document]
adityapk00: Fixed in a41ab5a

adityapk00 (approved):

mattsse:
+1

PR #897, opened 2026-07-29
fix: decrypt deposits with indexed historical keys

mattsse:
## Summary

- Bind configured deposit-decryption secrets to finalized `ZonePortal` key indexes and select the exact indexed key while preparing encrypted deposits.
- Reconstruct still-valid historical key bindings from the Portal snapshot on startup and fail closed when a required key is unavailable.
- Add startup configuration for pre-provisioned/historical keys, operational rotation guidance, unit coverage, and an end-to-end K0/K1 grace-period test.

## Why

Encrypted deposits carry the Portal `keyIndex` they were accepted under, but the node previously decrypted every deposit with only the current sequencer key. A deposit accepted for K0 during K1's grace period would therefore fail decryption after rotation. The Portal remains authoritative for admission and grace validity; the node now follows the accepted deposit's immutable key index.

## Operational impact

Sequencing-capable nodes must preload and restart with the next private key before its Portal registration, and retain previous keys while the Portal's grace period accepts them. The active sequencer key remains included automatically.

[ZONE-133](https://linear.app/tempoxyz/issue/ZONE-133/t1-006-encrypted-deposits-accepted-under-historical-keys-are-decrypted)


crates/node/src/node.rs
```diff
@@ -510,6 +532,10 @@ where
 
         self.resolve_and_seed_tokens(&l1_provider, tempo_block_number)
             .await?;
+        if let Some(keys) = self.l1_config.encryption_keys.clone() {
+            self.resolve_and_seed_encryption_keys(&l1_provider, tempo_block_number, &keys)
+                .await?;
+        }
```
klkvr, line 538: not sure if i like this pattern where empty container is initialized early and then in a random place we are actually filling it with the initial data. we have something similar with enabled tokens and leadership schedule but not sure if we can do much to simplify this

klkvr (dismissed):
lgtm

klkvr (approved):

klkvr:
cyclops audit

PR #899, opened 2026-07-29
feat(scripts): add standalone zone deployment without the factory [do not merge]

legion2002:
## Status

Operational standalone-deployment workaround. **Do not merge** while production uses this path: the script temporarily patches the reference contracts because the target chain has no native TIP-1091 ZoneFactory.

## What changed

Adds `scripts/deploy-standalone-zone.sh` to deploy and initialize a standalone Verifier, ZoneMessenger, and ZonePortal.

The script now:

1. requires the complete `SEQUENCERS` set and `THRESHOLD` explicitly;
2. validates the quorum, rejects zero or duplicate addresses, and ensures the encryption transaction signer belongs to the set before spending gas;
3. installs the complete sequencer set atomically in `ZonePortal.initialize`, leaving `sequencerSetVersion == 0`;
4. registers the shared deposit-encryption key using separate proof and transaction signers; and
5. verifies the on-chain count, threshold, membership, leader, set version, and encryption key.

It also binds the standalone ZoneMessenger directly to the deployed Portal and restores both patched reference source files on exit.

## Multi-sequencer usage

Load private keys from the normal secret manager rather than putting them in shell history.

```bash
export PRIVATE_KEY='<funded deployer key>'
export SEQUENCERS='[0x1111111111111111111111111111111111111111,0x2222222222222222222222222222222222222222,0x3333333333333333333333333333333333333333]'
export THRESHOLD=2
export SEQUENCER_TRANSACTION_PRIVATE_KEY='<private key for one address in SEQUENCERS>'
export SEQUENCER_ENCRYPTION_PRIVATE_KEY='<shared deposit-decryption key>'
export ETH_RPC_URL='https://rpc.example.invalid'

./scripts/deploy-standalone-zone.sh
```

Key roles:

- `PRIVATE_KEY`: deploys and initializes the standalone contracts.
- `SEQUENCER_TRANSACTION_PRIVATE_KEY`: a registered settlement signer that submits and pays for `setSequencerEncryptionKey`.
- `SEQUENCER_ENCRYPTION_PRIVATE_KEY`: the shared node encryption key that signs the proof of possession; it does not need to be a registered settlement signer.
- P2P Ed25519 identity keys remain per-node infrastructure inputs and are not used by this L1 deployment script.

## Why initialization must be atomic

Calling `setSequencerSet` after deployment would first expose a temporary deployer-only 1-of-1 portal and increment the set version. Passing the intended set and threshold directly to `initialize` creates the correct leader and quorum at version 0.

## Validation

- rebased onto current `main`
- `bash -n scripts/deploy-standalone-zone.sh`
- `git diff --check origin/main...HEAD`
- negative preflight cases: missing quorum, invalid threshold, duplicate signer, non-member transaction signer, mismatched factory authority
- complete mocked 3-sequencer, threshold-2 deployment path, including post-initialization quorum and encryption-key assertions
- prior live Moderato deployment validation remains documented in the PR discussion

### DIFF
--- a/scripts/deploy-standalone-zone.sh
+++ b/scripts/deploy-standalone-zone.sh
@@ -1,9 +1,13 @@
 #!/usr/bin/env bash
 #
 # Deploy a standalone zone — Verifier, ZoneMessenger, and an initialized ZonePortal — without the
-# native TIP-1091 ZoneFactory. Needs only a private key and an RPC URL.
+# native TIP-1091 ZoneFactory. Needs a deployer key, an active sequencer transaction signer, an
+# encryption key, and an RPC URL.
 #
-#   PRIVATE_KEY=0x... ETH_RPC_URL=http://localhost:8545 ./scripts/deploy-standalone-zone.sh
+#   PRIVATE_KEY=0x... \
+#   SEQUENCER_ENCRYPTION_PRIVATE_KEY=0x... \
+#   ETH_RPC_URL=http://localhost:8545 \
+#   ./scripts/deploy-standalone-zone.sh
 #
 # Why this needs to patch the reference contracts
 # ------------------------------------------------
@@ -31,6 +35,11 @@ if [ -z "${PRIVATE_KEY:-}" ]; then
   exit 1
 fi
 
+if [ -z "${SEQUENCER_ENCRYPTION_PRIVATE_KEY:-}" ]; then
+  echo "SEQUENCER_ENCRYPTION_PRIVATE_KEY must be set (the shared Zone deposit-decryption key)" >&2
+  exit 1
+fi
+
 ETH_RPC_URL="${ETH_RPC_URL:-http://localhost:8545}"
 
 for tool in forge cast jq; do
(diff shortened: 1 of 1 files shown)


legion2002:
Tested on moderato. The deployment half works against the live testnet — all three contracts went out fine:

| | address |
|---|---|
| Verifier | `0xf8dc675Fc74A4F13ee76fC1A990774D106373D37` |
| ZoneMessenger | `0xf722982836f7fB6c716F9F34027674Fac260eF21` |
| ZonePortal | `0xD77ccfBB2dC6b7931D0d7EE56935078d3CeDD46e` |

`initialize` then reverted with `UnknownFunctionSelector(0x23143aff)`. That selector is `tokenTransferPolicyId(address)` — `_enableTokenInternal` reaches the TIP-403 registry, and those entry points are TIP-1092, which activates at **T9**. Moderato is still on T8; its `t9Time` is 2026-08-05 14:00 UTC. Same call on a local T10 node returns `(false, 1)`, on moderato it reverts, so it's purely the fork gate.

Not a script bug, but it burned gas on three deployments before failing, so I added a pre-flight check on `tokenTransferPolicyId` that exits before patching or deploying and explains why. Verified it now fails fast on moderato and still completes on a T10 node.

So a standalone zone can't be initialized on moderato until T9 lands on Aug 5.

scripts/deploy-standalone-zone.sh
```diff
@@ -0,0 +1,153 @@
...
+
+ETH_RPC_URL="${ETH_RPC_URL:-http://localhost:8545}"
+
+for tool in forge cast jq; do
+  command -v "$tool" >/dev/null || { echo "$tool is required but not installed" >&2; exit 1; }
+done
+
+DEPLOYER="$(cast wallet address "$PRIVATE_KEY")"
+
+# Zone parameters. Defaults produce a single-sequencer, fully open zone owned by the deployer.
```
jenpaff: missing shared sequencer encryption key

legion2002:
cyclops audit super-fast

PR #900, opened 2026-07-29
fix(rpc): allow sequencers to read foreign account state

mattsse:
## Summary

- allow Portal-configured sequencers to read foreign balances and nonces through private RPC
- keep ordinary foreign reads masked as `0x0`
- cover both paths with nonzero balance and nonce assertions

## Why

`eth_getBalance` and `eth_getTransactionCount` returned the privacy dummy before considering the caller's sequencer role. Authorized sequencer tooling therefore observed false zero state.

Fixes ZONE-139

[Linear issue](https://linear.app/tempoxyz/issue/ZONE-139/t1-028-private-rpc-masks-foreign-account-state-from-configured)


0xrusowsky (dismissed):

legion2002 (dismissed):

legion2002:
closed becaused ZONE 139 was marked as won't fix

PR #902, opened 2026-07-29
test(evm): cover reverted advanceTempo openers

legion2002:
## Summary

- add an end-to-end executor regression for a reverted `advanceTempo` opener
- verify Tempo rejects the failed system transaction before commit
- verify the rejected opener does not satisfy the zone block guard
- leave production execution behavior unchanged

## Why

CYCLOPS-530 assumed that a failed system transaction could reach Zones as a normal execution result. Current `TempoEvm` already rejects failed system transactions, so that path is invalid and Zones does not need a duplicate consensus check. This test locks in the dependency behavior and catches a future regression without changing production code.

## Validation

- `rustup run nightly cargo test -p zone-evm reverted_advance_tempo_does_not_satisfy_block_guard -- --nocapture`
- `rustup run nightly cargo test -p zone-evm`
- `rustup run nightly cargo clippy -p zone-evm --all-targets -- -D warnings`

### DIFF
--- a/crates/evm/src/executor.rs
+++ b/crates/evm/src/executor.rs
@@ -255,13 +255,19 @@ where
 #[cfg(test)]
 mod tests {
     use super::{
-        ADVANCE_TEMPO_SELECTOR, ZONE_INBOX_ADDRESS, ZONE_OUTBOX_ADDRESS, ZoneBlockPhase,
-        ZoneTransactionKind,
+        ADVANCE_TEMPO_SELECTOR, ZONE_INBOX_ADDRESS, ZONE_OUTBOX_ADDRESS, ZoneBlockExecutor,
+        ZoneBlockPhase, ZoneTransactionKind,
     };
 
-    use alloy_consensus::{Signed, TxLegacy};
-    use alloy_primitives::{Address, Bytes, Signature, U256};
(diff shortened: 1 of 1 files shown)


crates/evm/src/executor.rs
```diff
@@ -177,20 +179,53 @@ fn validate_advance_tempo(
...
 }
 
+fn validate_advance_tempo_execution<H>(
+    is_advance_tempo: bool,
+    result: &ResultAndState<H>,
+) -> Result<(), BlockExecutionError> {
+    // TempoEvm currently rejects failed system transactions before returning a result. Keep the
+    // zone's consensus invariant local as well so upstream execution changes cannot turn a failed
+    // opener into a committed `has_advanced_tempo` marker.
+    if is_advance_tempo && !result.result.is_success() {
```
klkvr: i don't think this is necessary? on tempo side we already enforce that system txs never fail https://github.com/tempoxyz/tempo/blob/63ee158051cfcc98fca9567d8142d8a6d5070f4c/crates/evm/src/evm.rs#L291-L293

klkvr (changes requested):

klkvr (dismissed):

klkvr (approved):

PR #904, opened 2026-07-29
fix(portal): cap deposits per Tempo block

legion2002:
## Summary
- cap each `ZonePortal` at 640 total deposit queue entries per Tempo block
- cap public regular/encrypted deposits at 620, reserving 20 slots for one complete withdrawal bounce-back batch
- add gas-budget, withdrawal-liveness, planner-bound, storage-layout, and portal behavior regression coverage

## Why
`advanceTempo` must atomically process every deposit emitted in one finalized Tempo block under the 250,000,000 gas system budget. Without admission control in the portal, an oversized block can make the next L1 advance unexecutable.

The internal reserve also prevents public deposits from consuming every queue slot before `processWithdrawals`: the official 20M-gas withdrawal batch can contain at most 19 simple withdrawals, so 20 reserved slots let at least one complete batch advance the FIFO under sustained public deposit load.

## Capacity calculation
The regression conservatively exercises 640 encrypted deposits whose recipients reject the transfer, including decryption and receive-policy handling. This is more expensive than the reachable maximum mix of 620 public deposits and 20 regular bounce-backs.

- T9 gas used: 185,349,414
- System budget: 250,000,000
- Headroom: 64,650,586 gas (25.86%)

The regression enforces a stricter 200,000,000-gas ceiling, preserving at least a 20% buffer.

## Validation
- `forge test --match-path test/tempo/ZonePortal.t.sol` — 169 passed
- `rustup run nightly cargo test -p zone-precompiles` — 120 passed
- `rustup run nightly cargo test -p zone-sequencer` — 62 passed
- regenerated and verified `ZonePortal` storage layout
- `forge build --sizes` — `ZonePortal` runtime 22,257 bytes, 2,319-byte margin
- formatting and diff checks passed

Linear: ZONE-167

### DIFF
--- a/crates/precompiles/src/inbox/tests.rs
+++ b/crates/precompiles/src/inbox/tests.rs
@@ -5,11 +5,14 @@ use alloy_primitives::{Bytes, address, keccak256};
 use alloy_rlp::Encodable as _;
 use alloy_sol_types::{SolCall, SolError};
 use revm::precompile::PrecompileResult;
+use tempo_chainspec::hardfork::TempoHardfork;
 use tempo_precompiles::{
     PATH_USD_ADDRESS, TIP403_REGISTRY_ADDRESS,
+    receive_policy_guard::ReceivePolicyGuard,
     storage::{ContractStorage, Handler, StorageCtx},
     test_util::TIP20Setup,
     tip20::{ITIP20, TIP20Token},
+    tip403_registry::{ALLOW_ALL_POLICY_ID, ITIP403Registry, REJECT_ALL_POLICY_ID, TIP403Registry},
     zone_factory::{ZonePortalStorage, zone_portal_slots},
 };
 use tempo_primitives::TempoHeader;
@@ -47,7 +50,16 @@ impl Harness {
     }
 
     fn with_l1(l1: MockL1Reader) -> eyre::Result<Self> {
+        Self::with_l1_and_spec(l1, TempoHardfork::default())
+    }
+
+    fn with_spec(spec: TempoHardfork) -> eyre::Result<Self> {
+        Self::with_l1_and_spec(MockL1Reader::default(), spec)
+    }
+
+    fn with_l1_and_spec(l1: MockL1Reader, spec: TempoHardfork) -> eyre::Result<Self> {
         let mut ctx = test_context();
+        ctx.cfg.spec = spec;
         let genesis_rlp = encode_header(&TempoHeader::default());
         let genesis_hash = keccak256(&genesis_rlp);
         {
@@ -56,6 +68,9 @@ impl Harness {
                 TempoState::new().initialize(&genesis_rlp)?;
                 ZoneInbox::new().initialize()?;
                 ZoneOutbox::new().initialize()?;
+                if spec.is_t6() {
+                    ReceivePolicyGuard::new().initialize()?;
+                }
                 TIP20Setup::path_usd(ALICE)
                     .with_issuer(ALICE)
                     .with_issuer(ZONE_INBOX_ADDRESS)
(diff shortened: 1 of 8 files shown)


crates/precompiles/src/inbox/tests.rs
```diff
@@ -56,6 +68,9 @@ impl Harness {
                 TempoState::new().initialize(&genesis_rlp)?;
                 ZoneInbox::new().initialize()?;
                 ZoneOutbox::new().initialize()?;
+                if spec.is_t6() {
```
0xrusowsky: zones exist post-T6 so this can always be on

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -779,6 +789,18 @@ contract ZonePortal is IZonePortal {
         internal
         returns (uint64 thisDeposit)
     {
+        uint64 currentBlock = uint64(block.number);
+        if (_depositCountBlock != currentBlock) {
+            _depositCountBlock = currentBlock;
+            _depositsInCurrentBlock = 0;
+        }
+        if (_depositsInCurrentBlock >= MAX_DEPOSITS_PER_TEMPO_BLOCK) {
+            revert DepositBlockCapacityExceeded(MAX_DEPOSITS_PER_TEMPO_BLOCK);
```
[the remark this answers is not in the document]
0xrusowsky: my understanding is that we already leave a buffer to avoid this issue

crates/precompiles/src/inbox/tests.rs
```diff
@@ -47,7 +50,16 @@ impl Harness {
     }
 
     fn with_l1(l1: MockL1Reader) -> eyre::Result<Self> {
+        Self::with_l1_and_spec(l1, TempoHardfork::default())
```
0xrusowsky: zones exist post T8, so can we force it here to not have to pass specs around?

specs/ref-impls/src/tempo/ZonePortal.sol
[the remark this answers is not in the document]
legion2002: Saved 20 slots for bouncebacks. Addressed in https://github.com/tempoxyz/zones/commit/b5d5e2e1fe2f262a74652fcac5c7f2effecb4eec

crates/precompiles/src/inbox/tests.rs
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/b5d5e2e1fe2f262a74652fcac5c7f2effecb4eec

crates/precompiles/src/inbox/tests.rs
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/b5d5e2e1fe2f262a74652fcac5c7f2effecb4eec

legion2002:
cyclops audit fast

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -775,10 +789,25 @@ contract ZonePortal is IZonePortal {
         }
     }
 
-    function _recordDeposit(bytes32 newCurrentDepositQueueHash)
+    function _recordDeposit(
+        bytes32 newCurrentDepositQueueHash,
+        uint64 maximum
```
0xKitsune, line 794: Do we need a param for this since its defined using consts above?

0xKitsune (approved):

0xrusowsky (approved):

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002: think it's fine, it's useful for testing ig

PR #913, opened 2026-07-30
fix(precompiles): verify receive policies on portal <> zone

0xrusowsky:
ref: `ZONES-166`


0xrusowsky:
cyclops audit fast note="full issue description can be found on linear issue `ZONES-166`"

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -1097,6 +1098,14 @@ contract ZonePortal is IZonePortal {
         internal
         returns (bool success)
     {
+        try TIP403_REGISTRY.validateReceivePolicy(token, address(this), to) returns (
```
0xrusowsky: addressed in [`09d0a37` (this PR)](https://github.com/tempoxyz/zones/pull/913/commits/09d0a371d68e79c9cc4aa89324dfa9f79899c887)

0xrusowsky:
cyclops audit fast note="full issue description can be found on linear issue `ZONES-166`"

legion2002:
as discussed on slack, think it is just expected behavior that TIP 20 receive policy redirected transfers would not go the original user, and it's the user's job to ensure the correct receive policy config is set.

I don't think the portal should try to give out any guarantees, except that we will call transfer once.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -1097,6 +1098,23 @@ contract ZonePortal is IZonePortal {
         internal
         returns (bool success)
     {
+        address effectiveRecipient;
+        try StdPrecompiles.ADDRESS_REGISTRY.resolveRecipient(to) returns (address resolved) {
+            effectiveRecipient = resolved;
+        } catch {
+            return false;
+        }
```
0xKitsune, line 1138: IIRC the address registry is disabled on zones?

0xrusowsky: we enabled it and also storage credits cause they were potentially causing the state root issues

crates/precompiles/src/inbox/mod.rs
```diff
@@ -284,38 +285,54 @@ impl ZoneInbox {
...
+            if err.is_system_error() {
+                Err(err)
+            } else {
+                Ok(false)
             }
-            Err(_) => Ok(false),
+        };
+
+        let can_receive = AddressRegistry::new()
+            .resolve_recipient(to)
```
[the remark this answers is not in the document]
0xrusowsky: disabled until we enable the address registry on zones

crates/precompiles/src/test_utils/local.rs
```diff
@@ -33,9 +33,9 @@ pub(crate) use crate::ecies::{build_plaintext, compressed_x_and_parity, encrypt_
 pub(crate) type TestContext =
     Context<TempoBlockEnv, TxEnv, CfgEnv<TempoHardfork>, CacheDB<EmptyDB>>;
 
-/// Create an empty test EVM context at the default Tempo hardfork.
+/// Create an empty test EVM context at the 1st Tempo hardfork with zone deployments.
 pub(crate) fn test_context() -> TestContext {
-    Context::new(CacheDB::new(EmptyDB::new()), TempoHardfork::default())
+    Context::new(CacheDB::new(EmptyDB::new()), TempoHardfork::T8)
 }
```
klkvr, line 39: hmm `TempoHardfork::default()` seems more future proof?

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -1084,6 +1085,23 @@ contract ZonePortal is IZonePortal {
...
+
+        try TIP403_REGISTRY.validateReceivePolicy(
+            token, address(this), effectiveRecipient
+        ) returns (
+            bool authorized, ITIP403Registry.BlockedReason
+        ) {
+            if (!authorized) return false;
+        } catch {
+            return false;
+        }
```
klkvr, line 1148: this API does feel a bit footgunish, makes me wonder if we should at least make `validateReceivePolicy` fail for virtual addresses

klkvr (dismissed):

crates/precompiles/src/test_utils/local.rs
0xrusowsky: default is T2 iiric, maybe we should bump it

specs/ref-impls/src/tempo/ZonePortal.sol
0xrusowsky: let's discuss it in a follow-up for tempo

crates/precompiles/src/test_utils/local.rs
0xrusowsky: and zones is +T8

0xKitsune (approved):

PR #894, opened 2026-07-29
fix(evm): enforce withdrawal finalizer boundary (ZONE-203)

0xKitsune:
This PR rejects zone blocks that emit a withdrawal request after finalizing a withdrawal batch, preserving the finalizer as the block's withdrawal boundary.

Closes ZONE-203

PR #914, opened 2026-07-30
fix(evm): reject unexpected zone system transactions

rkrasiuk:
## Summary

Rejecting unexpected Tempo system transactions during zone block execution.

After the required opening advanceTempo, the executor now permits only:
- Ordinary non-system transactions
- Canonically encoded `ZoneOutbox.finalizeWithdrawalBatch` system transactions

All other system transactions, including privileged outbox setters, cause block validation to fail.

### DIFF
--- a/crates/evm/src/executor.rs
+++ b/crates/evm/src/executor.rs
@@ -20,8 +20,10 @@ use tempo_primitives::{TempoReceipt, TempoTxEnvelope, TempoTxType};
 use tempo_revm::evm::TempoContext;
 use zone_chainspec::ZoneChainSpec;
 use zone_l1::state::L1StateProvider;
-use zone_precompiles::{ADVANCE_TEMPO_SELECTOR, L1StorageReader, tx_context};
-use zone_primitives::constants::ZONE_INBOX_ADDRESS;
+use zone_precompiles::{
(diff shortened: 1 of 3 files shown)


crates/evm/src/executor.rs
```diff
@@ -164,25 +168,48 @@ fn validate_advance_tempo(
...
+    if tx.is_system_tx()
+        && !tx.calls().any(|(kind, input)| {
+            kind.to() == Some(&ZONE_OUTBOX_ADDRESS) && is_finalize_withdrawal_batch_calldata(input)
+        })
+    {
+        return Err(BlockValidationError::msg(
+            "system transactions after advanceTempo must call ZoneOutbox.finalizeWithdrawalBatch",
+        )
+        .into());
+    }
```
klkvr, line 198: should we enforce that there are no txs after `finalizeWithdrawalBatch`?

rkrasiuk: yes, but it's a separate issue (ZONE-203), left my comment here
https://github.com/tempoxyz/zones/pull/894#pullrequestreview-4819218448

0xrusowsky (commented):
lgtm, pending incorporation of arsenii's cmnt

0xrusowsky (approved):
approving for now, pending a follow-up to refactor into a state-machine

PR #916, opened 2026-07-30
fix(precompiles): validate fee beneficiary before execution

rkrasiuk:
## Summary

Move TIP-403 fee-beneficiary authorization into the pre-execution fee hook.

Previously, `collect_fee_post_tx` checked whether the block beneficiary was authorized to receive the selected fee token. Because this hook runs after transaction execution, a rejected beneficiary produced a fatal EVM error and aborted payload construction. The offending transaction remained in the pool and could repeatedly halt zone block production.

### DIFF
--- a/crates/precompiles/src/zone_fee_manager/mod.rs
+++ b/crates/precompiles/src/zone_fee_manager/mod.rs
@@ -37,10 +37,13 @@ impl ZoneFeeManager {
         fee_payer: Address,
         fee_token: Address,
         max_amount: U256,
-        _beneficiary: Address,
+        beneficiary: Address,
     ) -> Result<Address> {
         let mut token = TIP20Token::from_address(fee_token)?;
-        token.ensure_authorized_as(&[(fee_payer, AuthRole::sender())])?;
+        token.ensure_authorized_as(&[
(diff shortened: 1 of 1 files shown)


0xKitsune (approved):

0xKitsune:
cyclops audit fast

crates/precompiles/src/zone_fee_manager/mod.rs
```diff
@@ -37,10 +37,13 @@ impl ZoneFeeManager {
         fee_payer: Address,
         fee_token: Address,
         max_amount: U256,
-        _beneficiary: Address,
+        beneficiary: Address,
     ) -> Result<Address> {
         let mut token = TIP20Token::from_address(fee_token)?;
-        token.ensure_authorized_as(&[(fee_payer, AuthRole::sender())])?;
+        token.ensure_authorized_as(&[
```
0xKitsune, line 43: Will this cause valid transactions to be rejected when the sequencer changes to a new leader? Pool validation uses the previous block beneficiary, so the incoming leader could be authorized while the transaction is still checked against the outgoing leader.

PR #920, opened 2026-07-30
refactor(node): require attestation store

Rjected:
Makes the attestation store required in `AttestationContext`, this is now never constructed as `None`


mattsse (approved):
LGTM

PR #924, opened 2026-07-30
fix(payload): finalize withdrawals in the request block

mattsse:
## Summary

Finalize withdrawals in the same zone block that executes the withdrawal request.

## Root cause

The payload builder read `getPendingWithdrawals` before `advanceTempo` and pool transactions, then carried a `has_prior_withdrawals` boolean into the finalization decision. A withdrawal created by a pool transaction was therefore absent from that block-start snapshot, adding one deterministic zone tick between `WithdrawalRequested` and `BatchFinalized`.

## Change

Remove the block-start read and stale boolean. After pool transactions execute, read the current pending withdrawal list once and append `finalizeWithdrawalBatch` when it is non-empty. Empty batches still finalize at `withdrawal_batch_interval_blocks`, and the finalization system transaction remains last in the block.

## Impact

A newly requested withdrawal now reaches its `BatchFinalized` boundary in the request block, removing one zone-block of deterministic latency. This only changes the zone-side batching boundary; the later L1 batch settlement and withdrawal processing stages are unchanged.

### DIFF
--- a/crates/node/tests/it/e2e.rs
+++ b/crates/node/tests/it/e2e.rs
@@ -958,37 +958,76 @@ async fn submit_withdrawal(
     dev_address: Address,
     amount: u128,
 ) -> eyre::Result<u64> {
-    let outbox = IZoneOutbox::new(ZONE_OUTBOX_ADDRESS, provider.clone());
-    let pending = outbox
-        .requestWithdrawal(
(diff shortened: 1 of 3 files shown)


Rjected (approved):
lgtm, makes sense to do this at block end

PR #925, opened 2026-07-30
fix(sequencer): settle batches anchored to current tip

mattsse:
## Summary

Allow batch and certificate anchors equal to the observed L1 head, avoiding the forced retry and two-second delay for current-tip batches. Future anchors remain rejected and the EIP-2935 history-window checks are unchanged.

Apply a bounded gas limit only when the resolved settlement anchor equals the observed head, while preserving Alloy/Tempo gas estimation for every other submission path. Add focused unit coverage plus a dev-L1 integration test that proves a batch anchored to block N settles successfully in N+1.

## EIP-2935 timing

EIP-2935 keeps recent block hashes in state so contracts such as `ZonePortal` can verify an L1 block hash. The important boundary is the current head:

1. After block N has been observed as the L1 head, the state at N contains historical hashes only through N-1. Hash(N) cannot have been stored while N was being executed because the block had not yet been sealed.
2. A settlement transaction created after observing N cannot be included in N itself. Its earliest possible inclusion is N+1.
3. During processing of N+1, the EIP-2935 system update stores hash(N) before normal transaction execution. The portal can therefore validate an anchor to N when the settlement transaction executes in N+1 or any later block inside the history window.

The previous off-chain eligibility check rejected `tempo_block_number >= current_l1_block`. That treated the valid equality case as a future anchor, forcing the submitter to retry and sleep until the head advanced. This change rejects only `tempo_block_number > current_l1_block`; the equivalent certificate-anchor validation now follows the same rule.

## Gas-estimation wrinkle

Alloy/Tempo performs `eth_estimateGas` against the latest completed state, which is still state N immediately after the submitter observes head N. At that point the EIP-2935 history does not contain hash(N), so simulating `submitBatch` reverts with `InvalidTempoBlockNumber` even though the real transaction can only execute in N+1, where hash(N) is available.

The dev-L1 reproduction confirmed that behavior: without an explicit gas limit, estimation failed before the transaction was broadcast. The workaround is deliberately scoped to submissions whose resolved anchor equals the observed head; those receive a bounded 2,000,000 gas limit, while all other settlement paths continue using normal estimation. In the successful integration run, the batch anchored to block 10 and the settlement transaction was included in block 11.


mattsse:
Superseded by #931, which ports this change unchanged (head-equal anchors, the bounded 2M gas limit for tip-anchored submissions, the certificate-anchor validation, and the current-tip stepping test) and extends the same condition to the atomic withdrawal fast path so tip-anchored combined settlement skips estimation as well.

PR #931, opened 2026-07-31
feat(sequencer): process eligible withdrawals atomically with settlement

mattsse:
## Summary

- Atomically combines `submitBatch` and `processWithdrawals` when the newly created withdrawal slot fits the planner's gas budget.
- Falls back to the existing standalone processor for backlogs, oversized slots, ancestry submissions, retries, and restart recovery.
- Accepts batches anchored to the observed L1 head and uses bounded gas for tip-anchored submissions, avoiding the previous forced retry.
- Logs the confirmed settlement transaction hash, inclusion block, and gas used.

## Why this matters

Today, a zone withdrawal takes two independent L1 transactions on the happy path:

1. `submitBatch` commits the zone block and appends its withdrawals to the portal queue.
2. The standalone withdrawal processor discovers the new queue entry on a later polling cycle and submits `processWithdrawals`.

That split adds at least one extra L1 inclusion and a polling delay before an otherwise-ready withdrawal is processed. It also creates an intermediate state where the batch is accepted but its eligible withdrawals are still waiting in the queue, and requires a second transaction to be signed, broadcast, tracked, and recovered independently.

This PR adds a fast path that places the two calls, in order, in one Tempo transaction:

```text
submitBatch(batch)
processWithdrawals(withdrawals)
```

For an eligible withdrawal slot, `BatchSubmitted` and `WithdrawalProcessed` therefore share the same transaction hash and L1 block. This:

- removes the extra block and polling-cycle latency from the common case;
- makes batch acceptance and withdrawal processing atomic—if either call fails, neither state transition is committed;
- avoids the operational overhead and failure surface of a second transaction; and
- gives operators and users one transaction receipt to follow for the complete settlement.

### Observed example

On the deployed Deel Zone, [this batch submission](https://explore.tempo.xyz/tx/0x96c50d25743d140f776028282b4b03091d522cf5cb51adc47ba29c3896705509) landed in L1 block `36,650,746`, while its corresponding [withdrawal processing transaction](https://explore.tempo.xyz/tx/0x8b5c4a0004dacce7c90caed7aa2d437cbefc827b7e4e8d6c36be85585fa41aaa) landed in block `36,650,747`. With this fast path, an otherwise eligible case like this is processed as the second call of the batch-settlement transaction, in the same L1 block.

```text
Current main (observed on Deel Zone)

L1 block 36,650,746
+-----------------------------------+
| Transaction A                     |
| submitBatch                       |
|   -> append withdrawal to queue   |
+-----------------------------------+
                  |
                  | standalone poll
                  | + another L1 transaction
                  v
L1 block 36,650,747
+-----------------------------------+
| Transaction B                     |
| processWithdrawals                |
|   -> process queued withdrawal    |
+-----------------------------------+


This PR (eligible fast path)

L1 block N
+-----------------------------------+
| Transaction A (atomic)            |
| 1. submitBatch                    |
|    -> append withdrawal to queue  |
| 2. processWithdrawals             |
|    -> process queued withdrawal   |
+-----------------------------------+
        same transaction hash
        same L1 block
```

## Safety and fallback behavior

The optimization only applies when the newly created withdrawal slot can be processed safely within the configured gas budget. The existing standalone processor remains responsible for queue backlogs, oversized slots, ancestry submissions, retries, and restart recovery.

The combined path is atomic: a failure reverts both calls and enters the existing retry and resync handling. This changes the happy-path packaging and latency, not the underlying recovery guarantees. It also continues to use the sequencer's existing sequential nonce keys; it does not introduce expiring nonces.

## Benchmark

[Full-journey run 30910891112](https://github.com/tempoxyz/zones/actions/runs/30910891112) on `2db465b` did not reach the workload, so it produced no latency sample. Zone creation succeeded in L1 block 5, but provisioning stopped because the factory returned a zeroed leader snapshot instead of the expected leader at epoch 1, activation block 5.


mattsse:
cyclops audit

crates/sequencer/src/monitor.rs
```diff
@@ -566,13 +570,27 @@ impl<P: ZoneSequencerProvider> ZoneMonitor<P> {
                 return Ok(());
             }
 
+            // Only the first attempt may drain the slot atomically. A combined transaction
+            // declares more gas than settlement alone, so after any failure, including bounded
+            // non-inclusion of that transaction, retries submit the batch by itself and leave
+            // the slot to the standalone processor.
+            let atomic_slot = (attempt == 1).then_some(AtomicWithdrawalSlot {
```
0xKitsune: After a combined send times out, that transaction is already in the 2D pool at this nonce. The next attempt changes the calldata to submit-only but reuses the same nonce and max fee, so Tempo rejects it as an underpriced replacement. That can leave settlement waiting on the larger transaction under congestion. Could the speculative send use a separate nonce lane, or could this retry explicitly replace it with a sufficient fee bump? A timeout-path test would make this clear.

PR #936, opened 2026-07-31
test: cover all node-spawning suites in nextest heavy-test filter

mattsse:
`stepping_e2e` and the three `demo_*` suites each boot a full in-process Tempo L1 dev node (`demo_cross_zone` boots two zone nodes on top of it), and `enable_token::test_enable_token_via_real_l1` does the same, but none of them were covered by the `threads-required = 8` override — they ran at full parallelism next to 8-thread-reserved neighbours, which is a likely source of the flaky timeouts the override exists to prevent.

This also drops the `test(advance_tempo)` entry entirely: all three tests in that binary are sync `#[test]` fns running revm against `CacheDB<EmptyDB>` in-process, so reserving 8 threads for them only serialized CI for no benefit.

More serialization can shift the two shard wall-times; the consolidation PRs later in this stack delete the most expensive redundant tests as the offset.

---

Independent of the chain below (targets `main` directly); listed with the stack for navigation. Next: #937.

1. #936 — nextest heavy-test filter ← this PR

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -6,5 +6,5 @@ test-threads = "num-cpus"
 # Integration tests spin up full reth nodes — limit concurrency to avoid
 # resource contention and flaky timeouts on CI runners.
 [[profile.ci.overrides]]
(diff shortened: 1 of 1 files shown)


mattsse:

PR #940, opened 2026-07-31
test: dedupe e2e harness helpers

mattsse:
Collapses the harness's duplicated internals and adds the helpers the suites adopt in the next PRs.

The eight `ZoneTestNode` constructors funnelled through three launch layers, six of them pass-throughs differing only in hardcoded defaults, ending in an eight-positional-argument function behind `#[allow(clippy::too_many_arguments)]`. There is now one `launch(ZoneNodeParams)` with a `Default`-able params struct; the named `start_*` constructors stay as thin wrappers with unchanged signatures (29 `start_from_l1` call sites untouched). `start_from_l1_at_block` merges into `start_from_l1_genesis_block` — their bodies were identical. The copy-paste genesis-builder pair becomes one fn with `at_block: Option<u64>`, the three private-RPC POST helpers become one `rpc_call_raw(.., auth: Option<&str>)` behind a shared lazy reqwest client, the five token builders share `fresh_token_fields` with a named `AUTH_TOKEN_TTL_SECS`, the portal role setters share `set_role_on_portal`, and the blacklist/whitelist policy constructors share `create_policy`. The gateway-mode assert's inverted `isGatewayOpen` check gets an explanatory comment rather than a merge — the inversion is contract semantics, not a copy-paste bug.

New adoption-ready helpers land under temporary `allow(dead_code)` markers that the follow-up PRs remove: `start_l1_and_zone` (the real-L1 preamble), `blacklist_on_token`, `ZoneAccount::deposit_raw`, `ZoneTestNode::wait_for_withdrawal_batch_index` and `decode_finalize_batch_call`, `expect_result`/`expect_error_code`, shared `L1_TIMEOUT`/`WITHDRAWAL_TIMEOUT` constants, and `build_secp256k1_token` in the shared `auth_tokens.rs` (the r ‖ s ‖ parity ‖ fields blob is currently hand-assembled in three places). Two deliberate behavior changes to review: the single-sequencer harness path now polls withdrawals at the same 500ms as `spawn_sequencer_with_config` instead of a silently different 5s, and `start_local_p2p_cluster` waits out the Commonware peer-handshake window itself instead of every caller sleeping around it.

---

Stacked on #939. Next: #941.

**Test de-slop stack** (merge in order; bases retarget as predecessors merge):
1. #936 — nextest heavy-test filter
2. #937 — remove dead code and debug noise
3. #938 — consolidate redundant e2e tests
4. #939 — split it/utils.rs into modules
[...]

PR #939, opened 2026-07-31
test: split it/utils.rs into modules

mattsse:
Move-only split of the 4.5k-line `tests/it/utils.rs` into `tests/it/utils/`, one module per concern: `node` (ZoneTestNode + the mock L1 RPC), `l1` (L1TestNode, TIP-403 seeding, genesis builders), `accounts` (ZoneAccount, withdrawal args), `p2p` (cluster + sequencer spawning), `fixture` (L1Fixture, queue injection), `private_rpc` (auth tokens, RPC contexts), and a `mod.rs` facade whose `pub(crate) use` globs keep every existing import path working — zero call-site changes in the 17 test files.

Three mechanical adjustments ride along: the `#[path]` include of `auth_tokens.rs` gains one directory level, items that now cross module boundaries become `pub(crate)`, and each module keeps only the imports it uses. `ZoneTestNode`'s doc block, which had been orphaned onto the `RpcApiFuture` type alias, is reattached to the struct. A second commit hoists the ~80 `use` statements that were buried inside function bodies (`ITIP20` and the TIP-403 registry pair were each imported up to six times).

Review with `git diff --color-moved=dimmed-zebra`.

---

Stacked on #938. Next: #940.

**Test de-slop stack** (merge in order; bases retarget as predecessors merge):
1. #936 — nextest heavy-test filter
2. #937 — remove dead code and debug noise
3. #938 — consolidate redundant e2e tests
4. #939 — split it/utils.rs into modules ← this PR
5. #940 — dedupe e2e harness helpers
6. #941 — adopt shared harness helpers
7. #942 — dedupe private-RPC, WS, and earn suites
8. #943 — clean up hot suites and refresh test docs

PR #938, opened 2026-07-31
test: consolidate redundant e2e tests

mattsse:
Removes tests whose assertions are strict subsets of surviving tests and merges one near-identical pair. Coverage mapping — every deleted test with the survivor covering its assertions:

- `stepping_e2e::test_batch_submission_after_configured_short_l1_gap` → `test_boundary_ancestry_submission_uses_recent_anchor` (its 59-line setup was byte-identical and is now the shared `start_short_gap_harness`; "a batch was submitted" is implied by the survivor's calldata assertions)
- `e2e::test_deposit_via_queue_injection` → `test_multiple_deposits_across_blocks` (mint amounts) + `test_zone_inbox_events_on_deposit` (event emission)
- `e2e::test_empty_l1_blocks_advance_zone` → `test_tempo_state_advances_with_l1_blocks` (also asserts the hash)
- `e2e::test_large_deposit_batch` → folded into `test_multiple_deposits_across_blocks` as a third injected block carrying the ten-deposit fan-out and per-recipient assertions
- `l1_e2e::test_deposit_via_real_l1` → `test_multiasset_deposit_withdrawal` (the pathUSD round trip is its first leg)
- `earn_zone_e2e::matrix_deposit_private_private_succeeds` and `zone_lifecycle_swapped` → `matrix_redeem_private_private_succeeds` (shares assertion carried over, making it the full lifecycle test)

The two 85%-identical router-bounce tests in `l1_e2e.rs` now share one body parameterized over `RouterCallbackKind`; both test names remain. `zone_swap_slippage_bounces` is renamed to `zone_deposit_min_vault_assets_bounces` — the swap it referenced no longer exists. Each merged scenario is one fewer full L1+zone spawn in CI.

---

Stacked on #937. Next: #939.

**Test de-slop stack** (merge in order; bases retarget as predecessors merge):
1. #936 — nextest heavy-test filter
2. #937 — remove dead code and debug noise
3. #938 — consolidate redundant e2e tests ← this PR
4. #939 — split it/utils.rs into modules
5. #940 — dedupe e2e harness helpers
6. #941 — adopt shared harness helpers
7. #942 — dedupe private-RPC, WS, and earn suites
8. #943 — clean up hot suites and refresh test docs

PR #937, opened 2026-07-31
test: remove dead code and debug noise from e2e suite

mattsse:
Pure deletions, no surviving code changes behavior (−760 lines).

Removes `advance_tempo_repro`: a 180-line debug scratchpad that printed every execution outcome — success, revert, halt, transport error — and swallowed all of them, so it could not fail. Also removes stepping's `#[ignore]`d extended-gap test, whose ignore reason pointed at a nightly CI that does not exist (nothing in the repo passes `--run-ignored`, so it never ran anywhere), and ~130 lines of dead harness code: 55 lines of hand-rolled ECIES no test exercises, `start_zone_with_private_rpc_l1_with_encryption` (byte-identical to its non-`_with_encryption` sibling — the inner fn always registers a key), `build_bad_token`, and a duplicated nested `available_address`.

The rest is the narration layer: 40 `// --- Step N ---` banners, 56 numbered doc-comment lines restating test bodies, three ASCII sequence diagrams in `l1_e2e.rs` (one visually broken), and four per-test "requires forge build" notes now stated once per module doc. Comments explaining ordering constraints and invariants stay; the demo files keep their flow diagrams.

---

First PR of the chain (targets `main`). Next: #938.

**Test de-slop stack** (merge in order; bases retarget as predecessors merge):
1. #936 — nextest heavy-test filter
2. #937 — remove dead code and debug noise ← this PR
3. #938 — consolidate redundant e2e tests
4. #939 — split it/utils.rs into modules
5. #940 — dedupe e2e harness helpers
6. #941 — adopt shared harness helpers
7. #942 — dedupe private-RPC, WS, and earn suites
8. #943 — clean up hot suites and refresh test docs

PR #941, opened 2026-07-31
test: adopt shared harness helpers in core e2e suites

mattsse:
The adoption pass over the core suites (−395/+219): `l1_e2e.rs` uses `start_l1_and_zone`, `blacklist_on_token`, and `deposit_raw` for its repeated preambles and raw portal deposits, replaces fifteen inline 60-second literals with `WITHDRAWAL_TIMEOUT`, polls instead of sleeping to check L1 block production, and turns an `unwrap_or_default()` that would have masked "no batches finalized" into an explicit `ensure!`. `e2e.rs` replaces three copy-pasted `withdrawalBatchIndex` poll closures with `wait_for_withdrawal_batch_index` (one tautological follow-up assert died with them), the three `BatchFinalized` decode blocks with `decode_finalize_batch_call`, and its four fund+approve preambles with a local helper; both post-cluster sleeps are gone. `restart_e2e.rs` replaces the four abort-sleep(1s)-respawn sequences with a `restart_sequencer` helper that awaits task teardown via the existing `abort_task` instead of sleeping over the race, and the suite's worst flake bet — sleep(5s) then assert new batches appeared — is now a poll.

`submit_withdrawal`/`submit_withdrawals` move to `utils::fixture` and replace `tip403_transfers`' hand-rolled copy. The original plan's `withdraw_token_with` retarget was wrong for these files: injection-based tests have no `L1TestNode`, so `ZoneAccount` cannot be constructed there. Two more deliberate keeps: `stepping_e2e`'s raw-JSON `fetch_submit_batch_call` stays for now (its calls-array fallback for AA transactions needs runtime verification), and `restart_e2e` keeps its 120-second withdrawal timeout since restart cycles span two sequencer lifecycles.

---

Stacked on #940. Next: #942.

6. #941 — adopt shared harness helpers ← this PR

### DIFF
--- a/crates/node/tests/it/e2e.rs
+++ b/crates/node/tests/it/e2e.rs
@@ -8,11 +8,9 @@
 use std::{net::TcpListener, time::Duration};
 
 use alloy::primitives::{Address, B256, Bytes, TxKind, U256, address};
-use alloy_consensus::Transaction;
 use alloy_eips::NumHash;
 use alloy_provider::{DynProvider, Provider};
 use alloy_rpc_types_eth::TransactionRequest;
-use alloy_sol_types::SolCall;
 use tempo_chainspec::spec::TEMPO_T0_BASE_FEE;
 use tempo_contracts::precompiles::ITIP20;
 use tempo_precompiles::PATH_USD_ADDRESS;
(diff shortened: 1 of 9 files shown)


mattsse:

PR #942, opened 2026-07-31
test: dedupe private-RPC, WS, and earn suites

mattsse:
`ws.rs` gains `request()`/`request_skipping_notifications()`, replacing ~14 copies of the send-then-parse chain and the three drain-until-id loops; the JSON-RPC error codes get named constants; the invalid-token test reuses `connect_with_token` instead of inlining the 13-line upgrade request; and the secp token blob comes from the shared `auth_tokens` helper. The two close-on-revocation tests now assert the socket actually terminates — the token-expiry test had been passing on any frame whatsoever, and strengthening it revealed the server ends the connection with an abrupt TCP reset rather than a Close frame, so `assert_socket_terminated` accepts Close, clean end, or reset, but no longer a delivered message.

`private_rpc_e2e.rs` merges the three auth-token tables into one test, folds the public-methods table into `test_method_tiers`, and folds the all-enabled-tokens assertion into `test_zone_metadata_methods` — four fewer full node spawns with identical assertions. Tier-table error extraction goes through `expect_error_code` so failures print the server's actual response, the hardcoded zoneTokens literal becomes `format!("{PATH_USD_ADDRESS:#x}")` (the sibling assertion proves that is the right constant), and the keychain-expiry sleep gets a comment explaining why it cannot be polled away (the key's expiry is on-chain wall-clock state).

`earn_zone_e2e.rs` merges the two stateless retired-router-surface tests into one fixture spawn (~30 L1 transactions saved), drops the unused `_output_token` parameter from `callback_data`, and messages the four bare U256-arithmetic assertions. The `EarnEncryptedDepositPayload` redeclaration stays — `sol!` cannot reference externally generated types in struct fields — but now carries a comment saying exactly that.

---

Stacked on #941. Next: #943.

7. #942 — dedupe private-RPC, WS, and earn suites ← this PR

### DIFF
--- a/crates/node/tests/it/earn_zone_e2e.rs
+++ b/crates/node/tests/it/earn_zone_e2e.rs
@@ -78,6 +78,9 @@ alloy_sol_types::sol! {
         ExcessReturnFee excess;
     }
 
+    // Mirrors `tempo_zone_contracts::EncryptedDepositPayload`; sol! cannot
+    // reference externally generated types in struct fields, so the layout is
+    // redeclared here and converted by `map_encrypted_payload`.
     struct EarnEncryptedDepositPayload {
         bytes32 ephemeralPubkeyX;
         uint8 ephemeralPubkeyYParity;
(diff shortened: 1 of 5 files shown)


mattsse:

PR #943, opened 2026-07-31
test: clean up hot suites and refresh test docs

mattsse:
The hot-file pass, deliberately last in the stack to minimize rebase pressure on files with recent feature work.

`handoff_e2e.rs` drops the four post-cluster sleeps (the harness waits out the Commonware handshake itself since the readiness change earlier in the stack) and extracts the thrice-copied producer-boundary assertion loop, the twice-copied block-4 funding preamble, and the pool-forwarding polls into local helpers — deletions and extractions only, no reordering. `tip403_policy.rs` loses `test_tip20_transfer_on_zone` (a weaker copy of `tip403_transfers::test_deposit_then_transfer`, minus that test's native-balance and gas checks), merges the structurally identical whitelist/blacklist tests into one table-driven test, names the magic addresses and compound-policy-graph ids, and uses `local_dev_zone_account` instead of hand-rolled mnemonic plumbing.

`enable_token.rs` consolidates its two import blocks, adopts the shared timeout constants, and documents that its real-L1 test is the live-`TokenEnabled`-event counterpart of `demo_asset_swap`'s genesis-carried path — they exercise different enable pipelines, so both stay. The demo files retire the last per-file timeout constants, `L1Fixture::make_deposit`/`make_deposit_for_block` (identical bodies, one taking an unused `&self`) become a single free `make_deposit()`, and the README swaps its rotting 10-of-97 per-test inventory for a per-module table, documents the `utils/` layout, fixes examples that no longer compiled, and drops the "chain ID collisions" known-issue fixed by `next_unique_chain_id` long ago.

---

Stacked on #942. Top of the stack.

8. #943 — clean up hot suites and refresh test docs ← this PR

### DIFF
--- a/crates/node/tests/it/README.md
+++ b/crates/node/tests/it/README.md
@@ -39,6 +39,13 @@ The harness provides two independent testing paths:
               └─────────────────────────┘
 ```
 
+The shared harness lives in `utils/`, one module per concern: `node`
+(`ZoneTestNode` + the mock L1 RPC), `l1` (`L1TestNode`, TIP-403 seeding,
+genesis builders), `fixture` (`L1Fixture` + queue injection), `accounts`
+(`ZoneAccount`, withdrawal args), `p2p` (multi-sequencer cluster), and
+`private_rpc` (auth tokens, RPC test contexts). Everything is re-exported
+through `utils::`.
+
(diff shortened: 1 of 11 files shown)


mattsse:

PR #949, opened 2026-07-31
feat: configure forced recovery via manifest

klkvr:
## Summary

- replace the force variant of `zone_setLeader` with an operator-declared `[forced_recovery]` manifest directive
- require the configured recovery hash to equal the local canonical head at startup, then install the runtime leadership override before role-dependent tasks start
- let the replacement leader produce from the next Tempo anchor until the first subsequent finalized portal transition reaches its activation boundary
- keep on-chain recovery explicit: nodes never submit `setLeader` automatically, and the operator can use the ordinary RPC when ready
- document the restart/removal requirements and update recovery, replication, RPC, and manifest tests

## Why

Waiting for a recovery `setLeader` transaction to finalize before activating the replacement can deadlock after enough downtime because L1 ingestion is bounded relative to zone progress. A manifest directive lets the operator restart every survivor with the same pinned canonical tip and apply the temporary runtime authority immediately, while the next finalized portal transition still restores the ordinary on-chain schedule.

### DIFF
--- a/crates/node/src/engine.rs
+++ b/crates/node/src/engine.rs
@@ -58,8 +58,9 @@ use zone_payload::{ZonePayloadAttributes, ZonePayloadTypes};
 
 /// Per-anchor production permit backed by the effective leadership schedule.
 ///
-/// The permit is a single schedule lookup: produce anchor `N` only if the portal schedule or an
-/// active bounded forced-recovery override assigns `N` to this node.
+/// The permit is a single schedule lookup: produce anchor `N` only if the portal schedule or a
+/// forced-recovery override assigns `N` to this node. An optimistic override is open-ended until
+/// the next finalized portal transition supplies the ordinary-authority boundary.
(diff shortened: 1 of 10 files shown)


adityapk00:
During manifest recovery, leader_for(anchor) correctly assigns the next anchor to the replacement B, but is_scheduled_leader(B) (in `runtime.rs`) only searches finalized portal transitions, where B is absent until LeaderUpdated is finalized. 

As a result, B’s locally produced blocks are dropped here instead of broadcast. The receive-side check at line 911 and settlement checks use the same predicate, so this should be fixed centrally or all effective-authority checks should  include forced_recovery.

adityapk00:
Note, the path in the backfill eventually gets triggered after the timeout and it does allow C to proceed with the block, but the live path stalls

adityapk00 (dismissed):

adityapk00 (approved):

PR #950, opened 2026-07-31
feat(bench): warm earn state and lower callback gas to 2.5M

mattsse:
Stacked on #931 so the pull-request benchmark job exercises these changes together with the atomic settlement fast path.

Under TIP-1016, the first-ever earn callback pays one-time system-level state creation — roughly 248k gas in-frame per new storage slot for the router, vault, portal, and DEX first touches — and that state gas never appears in receipt `gas_used`, which reports the regular dimension only. A failed cold callback reverts its creations, so a too-small allowance can never warm the state itself. This is why simply lowering the callback default failed deterministically (runs 30637943025 at 2.5M, 30640369281 at 3M, 30641666224 at 5M): the first journey wave wedged cold and every retry re-paid the full creation cost.

The fix is one untimed warm-up journey that runs the preset scenario with the full 10M allowance whenever the measured callback gas is configured lower, before the measured phase. Warm measured callbacks then only need execution gas plus per-transaction state headroom, so the measured default drops to 2.5M, where the withdrawal planner packs four callbacks per batch instead of one 12.25M singleton each and callback-bearing slots become eligible for atomic settlement. Defaults are unchanged when the measured allowance is not lower than the warm-up allowance, and the warm-up report is included in the benchmark artifact.

Dispatch run 30645610512 validates the warm-up with a 2.5M measured allowance on this branch.

### DIFF
--- a/.github/workflows/zones-benchmark.yml
+++ b/.github/workflows/zones-benchmark.yml
@@ -97,7 +97,7 @@ on:
         description: L1 callback gas requested by composable withdrawals (maximum 10000000).
         type: string
         required: true
-        default: "10000000"
+        default: "2500000"
       swap-liquidity:
         description: Untimed liquidity seeded into the selected swap mechanism.
         type: string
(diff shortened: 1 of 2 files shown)


mattsse:
Closing: the warm-up mechanism worked (one full cold journey at 10M completed in ~5s), but the measured wave at 2.5M still exhausted every callback allowance against a warm system (dispatch run 30645610512), while warm callbacks at 10M consume ~30k (run 30632161477). Same code and state consuming 30k at one allowance and >2.5M at another points at TIP-1016 child-frame gas accounting (the reservoir is passed in full to child frames, bypassing gas caps, but only exists when the outer transaction exceeds the per-tx regular budget), not at workload. The callback allowance is not client-sizeable until that is understood on the tempo side; the 10M default stays.

PR #951, opened 2026-07-31
fix: scope private account getters

klkvr:
Restricts account-indexed NonceManager, AccountKeychain, and TIP-20 permit nonce reads to the owner or an active sequencer at the precompile boundary, including Multicall forwarding.

fixes ZONE-188


crates/precompiles/src/ztip20/mod.rs
```diff
@@ -83,17 +84,20 @@ impl<P: L1StorageReader> CallRules for TIP20Rules<P> {
...
-                    self.check_auth_with_sequencer(caller, &[call.account])
+                    check_caller_or_sequencer(&self.l1, caller, &[call.account])
                 })
             }
             ITIP20::allowanceCall::SELECTOR => {
                 decode_and_check::<ITIP20::allowanceCall>(args, |call| {
-                    self.check_auth_with_sequencer(caller, &[call.owner, call.spender])
+                    check_caller_or_sequencer(&self.l1, caller, &[call.owner, call.spender])
                 })
             }
```
0xKitsune, line 94: Looks like we could make this a bit easier to reason about by keeping the caller/account check explicit at each call site and only extracting the shared sequencer fallback.

```rust

ITIP20::balanceOfCall::SELECTOR => {
    decode_and_check::<ITIP20::balanceOfCall>(args, |call| {
        if caller == call.account {
            CallCheck::Continue
        } else {
            self.check_sequencer(caller)
        }
    })
}

ITIP20::allowanceCall::SELECTOR => {
    decode_and_check::<ITIP20::allowanceCall>(args, |call| {
        if caller == call.owner || caller == call.spender {
            CallCheck::Continue
        } else {
            self.check_sequencer(caller)
        }
    })
}
```

crates/precompiles/src/privacy.rs
```diff
@@ -0,0 +1,28 @@
...
+) -> CallCheck {
+    if accounts.contains(&caller) {
+        return CallCheck::Continue;
+    }
+
+    match l1.read_portal(|portal| &portal.is_sequencer[caller]) {
+        Ok(true) => CallCheck::Continue,
+        Ok(false) => CallCheck::Revert(Unauthorized {}.abi_encode().into()),
+        Err(error) => CallCheck::Error(CallRuleError::Tempo(error)),
+    }
```
0xKitsune, line 27: Will this read directly from the l1 every time? If so we should probably update to read from cache.

klkvr: there's already a cache layer within `l1` state provider

crates/precompiles/src/ztip20/mod.rs
klkvr: hmm we have quite a lot of callsites of `check_caller_or_sequencer` so seems appropriate to extract it into a shared helper to save 5 lines for each?

0xrusowsky (approved):

PR #955, opened 2026-07-31
fix: make withdrawal finalization system-only

0xKitsune:
This PR makes withdrawal batch finalization system-only and rejects transactions after the finalization transaction. This ensures the finalization event is unique and always at the tail of the block.

Closes CYCLOPS-527

### DIFF
--- a/crates/precompiles/src/outbox/mod.rs
+++ b/crates/precompiles/src/outbox/mod.rs
@@ -283,11 +283,13 @@ impl ZoneOutbox {
 
     fn finalize_withdrawal_batch<P: L1StorageReader>(
         &mut self,
-        l1: &L1State<P>,
+        _l1: &L1State<P>,
         caller: Address,
         call: IZoneOutbox::finalizeWithdrawalBatchCall,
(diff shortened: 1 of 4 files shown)


crates/precompiles/src/outbox/mod.rs
```diff
@@ -283,11 +283,13 @@ impl ZoneOutbox {
 
     fn finalize_withdrawal_batch<P: L1StorageReader>(
         &mut self,
-        l1: &L1State<P>,
+        _l1: &L1State<P>,
         caller: Address,
         call: IZoneOutbox::finalizeWithdrawalBatchCall,
     ) -> ZoneResult<B256> {
```
0xrusowsky, line 289: if we don't use `L1State` let's remove it from the input params

0xrusowsky (commented):
the approach works for now, but as discussed internally we should move to a state machine design

klkvr (approved):
lgtm, pending @0xrusowsky

0xrusowsky (approved):

PR #956, opened 2026-07-31
fix: require exact-height L1 cache entries

0xKitsune:
This PR requires L1 storage cache entries to match the requested block height exactly. This prevents log-free L1 state changes from being hidden by stale cached values.

Closes CYCLOPS-522


0xrusowsky (commented):
this solution makes the cache pointless. instead we should follow the approach of https://github.com/tempoxyz/zones/pull/855

PR #958, opened 2026-07-31
feat(builder): prewarm `advanceTempo` L1 reads

0xrusowsky:
Add builder-local optimization for Zone payload construction:

- Prepare encrypted deposits in parallel off the engine task while preserving canonical queue and decryption ordering.
- Before canonical `advanceTempo`, enqueue one-deposit throwaway executions directly on Reth's prewarming pool. These discover and warm exact L1 reads through the real Inbox, policy, encrypted-deposit, and token-enable paths.

Canonical `advanceTempo` remains the only committed system transaction. Prewarming state and receipts are discarded, pool workers bound concurrency, and queued work is cancelled once canonical `advanceTempo` completes.

## Performance

The full-journey workload contains eight 12-deposit payloads per run. The dedicated encrypted-deposit workload submits 50 deposits concurrently into one L1 block. All runs use Tempo T11 with Tempo `a843c91a`, 1 GiB of L1 state bloat, and seed `505050`.

Times below are means for the eight full-journey payloads and the single dedicated payload. Combined time is L1 deposit preparation plus payload construction; parenthesized percentages are changes relative to main.

| Workload | Revision | Preparation | Payload build | Combined |
| --- | --- | ---: | ---: | ---: |
| [Full journey](https://github.com/tempoxyz/zones/actions/runs/34493343514) | main (`5dcad06d`) | 1.80 ms | 3.27 ms | 5.07 ms |
| [Full journey](https://github.com/tempoxyz/zones/actions/runs/34494324644) | initial prewarming (`47f8038a`) | 1.73 ms (−4.1%) | 2.67 ms (−18.1%) | 4.40 ms (−13.1%) |
| [Full journey](https://github.com/tempoxyz/zones/actions/runs/34864766538) | parallel preparation (`372dd92f`) | 0.57 ms (−68.1%) | 2.95 ms (−9.6%) | 3.53 ms (−30.4%) |
| [Dedicated encrypted deposit](https://github.com/tempoxyz/zones/actions/runs/34495773286) | main (`5dcad06d`) | 7.21 ms | 11.31 ms | 18.52 ms |
| [Dedicated encrypted deposit](https://github.com/tempoxyz/zones/actions/runs/34496438157) | initial prewarming (`47f8038a`) | 7.38 ms (+2.3%) | 8.27 ms (−26.9%) | 15.65 ms (−15.5%) |
| [Dedicated encrypted deposit](https://github.com/tempoxyz/zones/actions/runs/34864778070) | parallel preparation (`372dd92f`) | 1.22 ms (−83.1%) | 9.05 ms (−20.0%) | 10.26 ms (−44.6%) |

Compared with main, the current revision reduces combined preparation-and-build time by **30.4%** for the 12-deposit workload and **44.6%** for the 50-deposit workload. Preparation itself improves by **68.1%** and **83.1%**, respectively. Payload construction remains **9.6%** and **20.0%** faster than main, although it was roughly 9–10% slower than the initial prewarming runs; that smaller difference should not be over-interpreted across hosts.


crates/payload/src/builder.rs
```diff
@@ -266,6 +297,14 @@ where
                 );
                 err
             })?;
+        // Canonical `advanceTempo` can no longer benefit from additional reads. Stop dispatching
+        // without waiting for workers that are already running.
```
0xrusowsky:
```suggestion
        // Canonical `advanceTempo` can no longer benefit from additional reads.
```

crates/payload/src/builder.rs
```diff
@@ -266,6 +297,14 @@ where
...
+        drop(prewarming);
+        info!(
+            target: "zone::payload",
+            elapsed = ?canonical_advance_start.elapsed(),
+            "canonical advanceTempo completed"
+        );
```
0xrusowsky:
```suggestion
```

0xrusowsky:
cyclops audit

crates/payload/src/prewarming.rs
```diff
@@ -0,0 +1,159 @@
...
+                enabled_tokens: context.prepared.enabled_tokens.clone(),
+                queued_deposits: vec![deposit.clone()],
+                decryptions,
+            };
+            let (ctx, cancel, tx) = (context.clone(), cancel.clone(), tx.clone());
+
+            pool.spawn(move || {
+                // Do not initialize a worker after canonical `advanceTempo` has completed.
+                if !cancel.is_cancelled() {
+                    let _ = ctx.prewarm_deposit(&partial);
```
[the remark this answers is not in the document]
0xrusowsky: this improvement can be done in a follow-up

crates/payload/src/prewarming.rs
```diff
@@ -0,0 +1,159 @@
...
+    /// already-running workers.
+    pub(crate) fn start(self) -> AdvanceTempoPrewarming {
+        let prewarming = AdvanceTempoPrewarming::default();
+        let num_deposits = self.prepared.queued_deposits.len();
+        let pool = self.task_executor.prewarming_pool();
+        let limit = self
+            .l1_fetch_concurrency
+            .saturating_sub(1)
+            .min(pool.current_num_threads())
+            .min(num_deposits);
```
klkvr: idk how useful propagating l1 concurrency here is given that we have reasonable prewarming pool worker count limits already

crates/payload/src/prewarming.rs
```diff
@@ -0,0 +1,150 @@
...
```
0xKitsune: [P2] This limit resets for each payload, so it doesn't bound L1 fetches once workers outlive a build. With `l1_fetch_concurrency = 2` and at least two prewarming threads, a one-deposit build can leave its worker in a slow RPC after canonical execution finishes; the next build can start another worker, allowing two speculative fetches plus canonical execution against a budget of two. Could we share speculative permits across builds and hold them until workers exit? A regression test with consecutive builds and a delayed first worker would cover this.

0xKitsune: [P2] The next payload gets a fresh budget while workers from the previous payload can still be fetching. With concurrency set to `2` and at least two pool threads, two consecutive one-deposit builds can leave two prewarm requests running alongside the canonical request. That exceeds the configured limit and adds RPC load when requests are already slow. Can we share the limit across builds and release each slot only when its worker finishes?

mattsse (commented):
what does this actually prewarm?

looks a bit complicated and this is probably not the right location for this, shouldnt this be triggered as soon as we've prepped the PreparedL1Block and not tied to the builder

or does this look like this because we cant directly fetch the necessary slots, it used to be the case that we did that iirc

crates/payload/src/prewarming.rs
```diff
@@ -0,0 +1,150 @@
...
+            };
+            let (ctx, cancel, tx) = (context.clone(), cancel.clone(), tx.clone());
+
+            pool.spawn(move || {
+                // Do not initialize a worker after canonical `advanceTempo` has completed.
+                if !cancel.is_cancelled() {
+                    let _ = ctx.prewarm_deposit(&partial);
+                }
+                let _ = tx.send(());
+            });
```
klkvr, line 70: hmm i'm wondering if we could just spawn a task per deposit onto the pool at once and let it figure out the concurrency? workers are limited anyway

crates/payload/src/prewarming.rs
0xrusowsky: addressed in 372dd92f4

crates/payload/src/prewarming.rs
0xrusowsky: addressed in 372dd92f4

crates/l1/src/block.rs
```diff
@@ -21,107 +23,46 @@ impl L1BlockDeposits {
...
+                .map(|deposit| prepare_deposit(deposit, &keys, portal_address, l1_block_number))
+                .collect::<Vec<_>>()
+        } else {
+            tokio::task::spawn_blocking(move || {
+                deposits
+                    .par_iter()
+                    .map(|deposit| prepare_deposit(deposit, &keys, portal_address, l1_block_number))
+                    .collect::<Vec<_>>()
+            })
+            .await?
```
klkvr, line 57: `spawn_blocking` and rayon are not free so it might be reasonable to gate the parallelization to only run if there's more than N deposits so that overhead doesn't outweight the gains for smaller blocks

0xrusowsky: addressed in [`c251b16`](https://github.com/tempoxyz/zones/pull/958/commits/c251b1646224268dd2d3bffbe80f0f35734ef17c)

klkvr (approved):

klkvr (approved):

PR #959, opened 2026-08-02
fix(l1): meter all L1 reads

0xrusowsky:
this PR adds gas metering for direct L1-backed precompile reads. It does so by keeping warmth tracking separate from the L1 value cache so host cache state cannot affect consensus gas accounting.

additionally, it removes the legacy generic L1-read methods from `TempoState`.


0xrusowsky:
cyclops audit

crates/precompiles/src/storage.rs
```diff
@@ -124,7 +146,34 @@ impl<P: L1StorageReader> L1State<P> {
...
+        let key = (account, slot);
+        let is_cold = self.access_set.borrow_mut().insert(key);
+        let result = (|| {
+            StorageCtx.deduct_gas(if is_cold {
+                COLD_SLOAD_COST
+            } else {
+                WARM_STORAGE_READ_COST
+            })?;
+            self.read_l1_storage_unmetered(account, slot, block_number)
+                .map_err(|err| TempoPrecompileError::Fatal(err.to_string()))
```
klkvr: ah i see, i guess this works but feels a bit hacky

crates/precompiles/src/storage.rs
```diff
@@ -124,7 +146,34 @@ impl<P: L1StorageReader> L1State<P> {
...
+        })();
+
+        // Failed accesses do not warm a previously cold key. An already-warm key remains warm.
+        if is_cold && result.is_err() {
+            self.access_set.borrow_mut().remove(&key);
+        }
```
klkvr: should we just only insert after succesful load?

crates/evm/src/database.rs
```diff
@@ -93,8 +93,9 @@ impl<DB: Database, L1: L1StorageReader> L1OverlayDB<DB, L1> {
         slot: U256,
         anchor: u64,
     ) -> Result<U256, ZoneDbError<DB::Error>> {
+        // REVM already charges TIP-403 overlay SLOADs; use the unmetered path to avoid double-charging.
```
klkvr: let's inline this fn into `Database::storage` so that it's obvious that only 403 reads are unmetered

0xrusowsky: addressed in [`106012a` (this PR)](https://github.com/tempoxyz/zones/pull/959/commits/106012a8e0f4a82310f88cdd2e9bdb641a946d98)

crates/precompiles/src/storage.rs
0xrusowsky: addressed in [`106012a` (this PR)](https://github.com/tempoxyz/zones/pull/959/commits/106012a8e0f4a82310f88cdd2e9bdb641a946d98)

klkvr (approved):

crates/evm/src/database.rs
```diff
@@ -93,8 +93,9 @@ impl<DB: Database, L1: L1StorageReader> L1OverlayDB<DB, L1> {
...
         self.l1
-            .read_l1_storage(address, B256::from(slot), anchor)
+            .read_l1_storage_unmetered(address, B256::from(slot), anchor)
```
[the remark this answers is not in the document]
0xrusowsky: in practice it will always be set beforehand

PR #960, opened 2026-08-03
refactor(p2p): extract protocol frame and routing policy

adityapk00:
Refactors the P2P runtime by separating protocol framing and routing
  authorization into dedicated modules.

  - Extracts backfill request/response encoding and decoding into protocol.rs
  - Centralizes peer membership and routing decisions in routing.rs
  - Uses atomic leadership snapshots for consistent routing decisions
  - Makes quorum membership checks explicit and manifest-backed
  - Fixes malformed backfill completions consuming outstanding requests
  - Adds protocol unit tests and end-to-end backfill response/replay coverage


crates/p2p/src/runtime.rs
```diff
@@ -952,76 +903,71 @@ async fn run_receivers(
...
             // Got backfill response (for an existing request)
             result = backfill_responses.recv() => {
                 let (peer, bytes) = result.map_err(|err| eyre::eyre!("backfill response receive failed: {err}"))?;
-                // Backfill responses are accepted from any quorum member — the request side
-                // only ever asks them, so a page from an internet-facing standby is unsolicited.
-                if peer == local_ed25519_public_key
-                    || !manifest.contains_ed25519_public_key(&peer)
-                    || !leadership.is_quorum_member(&peer)
-                {
+                let frame = match ResponseFrame::decode(bytes.as_ref()) {
```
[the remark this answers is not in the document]
adityapk00: wont fix. The sequencers are all trusted, and in case the key leaks, we already have limits (20mb block, rate limit for the cw channel), so this extra check is not needed.

crates/p2p/src/routing.rs
```diff
@@ -0,0 +1,287 @@
...
+        self.quorum.contains(peer)
+    }
+
+    pub(crate) fn other_peers(&self, local: &PublicKey) -> Vec<PublicKey> {
+        self.peers
+            .iter()
+            .filter(|peer| *peer != local)
+            .cloned()
+            .collect()
+    }
```
klkvr: let's maybe store our key as part of `RoutingMembership`? so that those methods don't take an arg and we can use `membership.local()` to access it

crates/p2p/src/runtime.rs
```diff
@@ -754,7 +709,7 @@ async fn run_commands(
             }
 
             P2pCommand::BroadcastSettlementProposal(proposal) => {
-                if !leadership.is_scheduled_leader(&local_ed25519_public_key) {
+                if !policy.may_broadcast_settlement_proposal() {
```
klkvr: maybe let's just invoke something like `policy.am_i_leader()` or similar? that way it's obvious from reading the code that we only broadcast when we are leading without looking deeper into the fn

crates/p2p/src/runtime.rs
```diff
@@ -716,7 +671,7 @@ async fn run_commands(
                 // `producer == leader_for(anchor)` fence. Recipients are all other manifest
                 // members — during a scheduled handoff the incoming leader must keep
                 // receiving live blocks.
-                if !leadership.is_scheduled_leader(&local_ed25519_public_key) {
+                if !policy.may_broadcast_block() {
```
klkvr: same here

crates/p2p/src/runtime.rs
```diff
@@ -727,7 +682,7 @@ async fn run_commands(
                     continue;
                 }
 
-                let recipients = &others;
+                let recipients = policy.block_recipients();
```
klkvr: simialrly here imo `policy.other_peers()` is better for clarify than `block_recipients()`

crates/p2p/src/runtime.rs
```diff
@@ -632,24 +593,20 @@ fn run(
...
 ) -> eyre::Result<()> {
-    let others: Vec<PublicKey> = peers
-        .iter()
-        .filter(|peer| *peer != &local_ed25519_public_key)
-        .cloned()
-        .collect();
-
     while let Some(command) = commands.recv().await {
+        let authority = leadership.authority_snapshot();
+        let policy = RoutingPolicy::new(&local_ed25519_public_key, &membership, &authority);
```
klkvr: every time we construct `RoutingPolicy` we also call `leadership.authority_snapshot()`, should we simplify by passing `leadership` into `RoutingPolicy::new`

crates/p2p/src/routing.rs
adityapk00: fixed

crates/p2p/src/runtime.rs
adityapk00: this is a good idea. Changed these names ( `am_i_retained_leader()`, `other_peers()`)

crates/p2p/src/runtime.rs
adityapk00: fixed

crates/p2p/src/runtime.rs
adityapk00: fixed

crates/p2p/src/runtime.rs
adityapk00: good idea, moved it into RoutingPolicy::new()

crates/p2p/src/runtime.rs
```diff
@@ -569,17 +535,14 @@ fn run(
...
-        let peers: Vec<PublicKey> = config
-            .manifest
-            .nodes()
-            .iter()
-            .map(|node| node.ed25519_public_key().clone())
-            .collect();
+        let membership = RoutingMembership::from_manifest(
+            &config.manifest,
+            local_ed25519_public_key.clone(),
+        );
```
klkvr, line 541: nit: could be `from_config(&config)` bc it contains both the key and the manifest

crates/p2p/src/runtime.rs
```diff
@@ -904,12 +855,11 @@ async fn run_receivers(
...
                 // A lagging follower must not drop the rightful
                 // producer of in-between anchors just because a later transition is already
                 // the "current" record.
-                if peer == local_ed25519_public_key
-                    || !leadership.is_scheduled_leader(&peer)
-                {
+                if !policy.may_accept_block(&peer) {
                     warn!(target: "zone::p2p", %peer, "Ignoring live block from non-leader");
                     continue;
                 }
```
klkvr, line 865: should we just remove this check? it's checked downstream in `validate_live_block_sender` anyway iiuc

klkvr (approved):

PR #961, opened 2026-08-03
fix: validate zone block timestamps against L1

decofe:
ref: `CYCLOPS-539`, `ZONE-171`

## Summary

- Validate replicated Zone block timestamp seconds and millisecond components against the anchored L1 header.
- Reject mismatched timestamps before execution and forkchoice.
- Add focused regression coverage for both mismatch cases.
- The companion nonzero base-fee validation is merged in [#962](https://github.com/tempoxyz/zones/pull/962).

## Testing

- `git diff --check`
- `cargo test -p zone-node replication::tests::validates_zone_block_timestamp_against_anchored_l1_header`

Prompted by: @0xrusowsky


0xrusowsky:
cyclops audit fast

0xrusowsky:
all cyclops findings are unrelated to the fixes in place

0xrusowsky:
closing as won't fix given that base_fee = 0 is already enforced + timestamp is not really used

PR #963, opened 2026-08-03
refactor(evm): model zone block execution as state machine

rkrasiuk:
## Summary

Refactors `ZoneBlockExecutor` transaction ordering into an explicit `ZoneBlockPhase` state machine for the opening `advanceTempo`, regular execution, and terminal withdrawal finalization phases.

Phase transitions occur only when transactions are committed and remain monotonic, preserving speculative execution behavior while making invalid ordering states unrepresentable. Adds exhaustive transition and regression tests for invalid system transactions, discarded results, and selector lookalikes.


0xrusowsky (approved):
impl looks good to me.
would be nice to have @klkvr's review too

klkvr (approved):

PR #964, opened 2026-08-03
fix: interrupt unauthorized websocket sessions

klkvr:
Cancel in-flight WebSocket dispatch and queued output when token expiry or keychain revocation terminates the session.

Fixes [CYCLOPS-529](https://linear.app/tempoxyz/issue/CYCLOPS-529/cyclops-revokedexpired-private-rpc-websocket-sessions-keep-streaming)


mattsse (approved):
LGTM

PR #965, opened 2026-08-03
fix(contracts): pin encryption key snapshot

legion2002:
## Summary

- read the current L1 block number before fetching the active encryption key
- use `encryptionKeyAtBlock` pinned to that exact block so the public key and key index come from one Portal state snapshot
- add a mocked-provider regression test for the atomic key/index response

## Why

`ZonePortal::encryption_key()` previously combined `sequencerEncryptionKey()` and `encryptionKeyCount()` from two independent unpinned `latest` calls. A key rotation between those calls could pair a key with the wrong index, causing an accepted encrypted deposit to fail decryption and bounce.

This keeps the existing helper API while making every current caller use a consistent key/index pair.

## Validation

- `rustup run nightly cargo test -p tempo-zone-contracts`
- `rustup run nightly cargo check -p tempo-zone-contracts --no-default-features`
- `rustup run nightly cargo check -p tempo-zone-contracts --all-features`
- `rustup run nightly cargo check -p tempo-xtask`
- `rustup run nightly cargo clippy -p tempo-zone-contracts --all-targets --all-features -- -D warnings`

Fixes [CYCLOPS-544](https://linear.app/tempoxyz/issue/CYCLOPS-544/cyclops-ripple-effect-audit-report).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11591,6 +11591,7 @@ dependencies = [
  "alloy-provider",
  "alloy-rpc-types-eth",
  "alloy-sol-types",
+ "alloy-transport",
  "const-hex",
  "futures",
  "serde",
--- a/crates/contracts/Cargo.toml
+++ b/crates/contracts/Cargo.toml
@@ -26,6 +26,7 @@ tokio = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
 
 [dev-dependencies]
+alloy-transport.workspace = true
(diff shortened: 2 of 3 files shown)


legion2002:
cyclops audit fast

0xrusowsky (approved):

legion2002:
the cyclops finding is unrelated to this PR, and will be addressed separately

PR #967, opened 2026-08-03
feat: require encrypted zone deposits

rkrasiuk:
## Summary

Deprecate regular zone deposits in favor of encrypted ones. The former `deposit` method is changed to be an API compatible alias to `depositEncrypted`.


rkrasiuk:
cyclops audit fast

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -885,10 +823,23 @@ contract ZonePortal is IZonePortal {
         address _token,
         uint128 amount,
         uint256 keyIndex,
-        EncryptedDepositPayload calldata encrypted,
+        DepositPayload calldata encrypted,
         address tempoRefundRecipient
     )
```
0xrusowsky, line 828: i guess we preserve it to minimize the diff?
since the `deposit()` change is already breaking the ABI, maybe we should mark this one as deprecated and eventually remove it?

xtask/src/demo_blacklist.rs
```diff
@@ -801,16 +761,16 @@ async fn wait_for_encrypted_result<P: Provider<TempoNetwork>>(
 ) -> eyre::Result<bool> {
     let processed_filter = Filter::new()
         .address(tempo_zone_contracts::ZONE_INBOX_ADDRESS)
-        .event_signature(IZoneInbox::EncryptedDepositProcessed::SIGNATURE_HASH)
+        .event_signature(IZoneInbox::DepositProcessed::SIGNATURE_HASH)
         .from_block(from_block);
     let failed_filter = Filter::new()
         .address(tempo_zone_contracts::ZONE_INBOX_ADDRESS)
-        .event_signature(IZoneInbox::EncryptedDepositFailed::SIGNATURE_HASH)
+        .event_signature(IZoneInbox::DepositFailed::SIGNATURE_HASH)
```
0xrusowsky, line 768: **Low — `demo-blacklist` waits for the gross deposit amount, but events contain the net amount**\n\nThis helper now matches `DepositProcessed` by requiring `event.data.amount == amount`, while all callers pass the original gross amount. `ZonePortal` deducts the deposit fee before queueing the deposit, so the zone event amount is lower whenever the fee is nonzero. The demo can therefore time out even when the deposit was processed successfully. Please either pass the net amount or match using a fee-independent deposit identifier.

.codex/skills/zones-deploy-debug/references/commands.md
```diff
@@ -119,7 +119,7 @@ tail -f /tmp/tempo-zone-my-zone*/logs/*/reth.log
 Useful patterns:
 
 ```bash
-rg -n "Prepared L1 block deposits|Including advanceTempo|TokenEnabled|DepositProcessed|WithdrawalProcessed" /tmp/tempo-zone-my-zone*/logs/*/reth.log
+rg -n "Prepared L1 block deposits|Including advanceTempo|TokenEnabled|EncryptedDepositProcessed|WithdrawalProcessed" /tmp/tempo-zone-my-zone*/logs/*/reth.log
```
0xrusowsky: **Low — debugging reference was updated in the wrong direction**\n\nThis command now searches for `EncryptedDepositProcessed`, but the new implementation emits `DepositProcessed`. The documented `rg` command will therefore miss the event emitted by this branch. Please keep `DepositProcessed` in the pattern.

crates/precompiles/src/inbox/mod.rs
```diff
@@ -263,8 +234,15 @@ impl ZoneInbox {
     fn process_withdrawal_bounce_back(
         &mut self,
         outbox: &mut ZoneOutbox,
-        deposit: Deposit,
+        portal: Address,
+        deposit: WithdrawalBounceBackDeposit,
     ) -> ZoneResult<()> {
+        if deposit.sender != portal
+            || !deposit.tempoRefundRecipient.is_zero()
+            || !deposit.memo.is_zero()
```
klkvr: should we remove those fields if they are always set to the same values?

specs/ref-impls/src/tempo/ZonePortal.sol
rkrasiuk: yup, we will deprecate shortly after reconciling everything to work with `deposit` function

xtask/src/demo_blacklist.rs
rkrasiuk: issue is pre-existing, can be addressed separately

klkvr (approved):

0xrusowsky (approved):

PR #968, opened 2026-08-03
fix(portal): reject registered portal bounceback recipients

decofe:
## Summary
- reject zero-address and registered ZonePortal bounce-back recipients through one shared validation helper
- cover regular and encrypted deposits with regression tests
- document the non-portal bounce-back invariant in the protocol spec

## Validation
- `forge fmt --check`
- `git diff --check`
- `forge test --match-path test/tempo/ZonePortal.t.sol` *(blocked: sandbox submodules are empty and Foundry could not clone them because GitHub credentials were rejected)*

Linear: ZONE-233 (reported as CHAIN-1248)

Prompted by: @0xrusowsky

### DIFF
--- a/specs/ref-impls/src/tempo/ZonePortal.sol
+++ b/specs/ref-impls/src/tempo/ZonePortal.sol
@@ -11,6 +11,7 @@ import {
     EncryptedDepositPayload,
     EncryptionKeyEntry,
     IVerifier,
+    IZoneFactory,
     IZoneMessenger,
     IZonePortal,
     MAX_WITHDRAWAL_CALLBACK_GAS,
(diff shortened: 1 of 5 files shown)


legion2002:
needs a corresponding tempo PR to bump portal bytecode

legion2002 (approved):

legion2002 (changes requested):
we should only merge this close to T10, otherwise people won't be able to test without the zone factory live

0xKitsune (approved):

Welith:
@0xrusowsky the PR does not reject deterministic future `ZonePortal` addresses. `isZonePortal()` only returns true for already-created IDs, so if `nextZoneId == 2`, the canonical Portal 2 address still passes `_validateBouncebackRecipient`

0xrusowsky:
since users cannot recover funds from any address that isn't their own wallet, which is almost all addresses. We decided we won't start making lists of random addresses that you can't send to.

PR #971, opened 2026-08-03
refactor(p2p): move backfill behind typed ports

adityapk00:
Move backfill traffic out of the shared P2pCommand/P2pEvent channel and into dedicated bounded command, request, and response channels.

Introduce a single-owner backfill coordinator responsible for request selection, outstanding request state, wire encoding, authorization, and response correlation. This removes the shared mutex-backed lifecycle state while preserving leader-first fallback and wire compatibility.

Keep incoming requests process-scoped so every node role can serve backfill, while routing accepted responses only to the active follower generation. Update node replication, role switching, and P2P lifecycle integration to use the typed interfaces.


adityapk00:
cyclops audit fast

crates/p2p/src/backfill.rs
```diff
@@ -0,0 +1,467 @@
...
+            attempts.push(vec![leader]);
+        }
+        attempts.push(candidates);
+
+        for (attempt, sources) in attempts.into_iter().enumerate() {
+            let leader_only = leader_first && attempt == 0;
+            let request = self.job.begin_request(&sources, now);
+            let Some((request_id, request_peers)) = request else {
+                debug!(target: "zone::p2p", start, sources = sources.len(), leader_only, "Skipping block backfill request because all eligible peers already have outstanding responses");
+                continue;
```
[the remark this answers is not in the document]
adityapk00: fixed

crates/p2p/src/backfill.rs
```diff
@@ -0,0 +1,467 @@
...
+        self.outstanding
+            .retain(|peer, request| request.request_id != request_id || sent.contains(peer));
+    }
+
+    fn cancel_request(&mut self, request_id: u64) {
+        self.outstanding
+            .retain(|_, request| request.request_id != request_id);
+    }
+
+    fn accepts(&self, peer: &PublicKey, request_id: u64, now: Instant) -> bool {
```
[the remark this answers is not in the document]
adityapk00: won't fix - we're trusting all the peers in this model, so the peer is expected not to send garbage (a compromised peer has many other ways of stalling the node, and this issue is bounded)

crates/p2p/src/runtime.rs
```diff
@@ -708,7 +568,15 @@ async fn run_commands(
...
             P2pCommand::BroadcastSettlementProposal(proposal) => {
-                if !policy.am_i_retained_leader() {
+                let (may_broadcast, recipients) = leadership.with_authority(|authority| {
+                    let policy =
+                        RoutingPolicy::new(&local_ed25519_public_key, &membership, authority);
+                    (
+                        policy.may_broadcast_settlement_proposal(),
+                        policy.settlement_proposal_recipients(),
+                    )
+                });
```
klkvr: hmm this is quite messy

crates/p2p/src/manifest.rs
```diff
@@ -377,17 +396,10 @@ impl LeadershipSchedule {
...
         let state = self.inner.read().expect("poisoned");
-        AuthoritySnapshot {
-            retained_leaders: state
-                .transitions
-                .values()
-                .map(|record| record.leader.clone())
-                .collect(),
-            next_anchor_record: state.next_anchor_record(),
-        }
+        f(Authority { state: &state })
```
klkvr: why not just keep a `&LeadershipSchedule` on `RoutingPolicy` and avoid the extra `Authority` struct and this closure altogether?

crates/p2p/src/runtime.rs
adityapk00: fixed

crates/p2p/src/manifest.rs
adityapk00: The direct &LeadershipSchedule version would re-acquire the lock every time. So, like If a transition landed between two lookups, it could observe mixed old/new schedule state. with_authority holds one lock for the entire decision, so its atomic.

klkvr (approved):

PR #972, opened 2026-08-03
fix(node): retain encryption keys across rotations

legion2002:
## Summary

- build a static sequencer ECIES keyring at startup by matching the current key and configured historical keys against a pinned Portal key-history snapshot
- select the private key using each encrypted deposit's exact `keyIndex`
- require every historical key that is still accepted during the Portal grace period
- fail closed when a queued deposit references a key that was not retained
- add repeatable `--sequencer-key-history-file` / `SEQUENCER_KEY_HISTORY_FILES` configuration

## Why

The Portal accepts deposits encrypted to a superseded key for `86,400` blocks after a rotation. The node currently decrypts every encrypted deposit with only the current private key, so a valid old-key deposit can be turned into a failed decryption/proof and bounced.

This addresses the key-retention issue raised in the [Cyclops review](https://github.com/tempoxyz/zones/pull/965#pullrequestreview-4846325508). It is separate from #965's key/index snapshot consistency fix.

The keyring is deliberately static: rotations require updating the configured key files and restarting the sequencer. Expired historical keys may be omitted; if an older queued deposit still references one, production stops at that deposit until the key is restored.

## Validation

- `rustup run nightly cargo test -p zone-l1 --lib`
- `rustup run nightly cargo test -p zone-node --features cli --lib`
- `rustup run nightly cargo clippy -p zone-l1 -p zone-node --features zone-node/cli --lib -- -D warnings`

### DIFF
--- a/crates/contracts/src/precompiles/zone_portal.rs
+++ b/crates/contracts/src/precompiles/zone_portal.rs
@@ -10,6 +10,9 @@ use alloy_primitives::{Address, B256, Bytes, keccak256};
 use alloy_sol_types::SolValue;
 use zone_primitives::constants::EMPTY_SENTINEL;
 
+/// Old encryption keys remain valid for new deposits for one day at one-second blocks.
+pub const ENCRYPTION_KEY_GRACE_PERIOD: u64 = 86_400;
+
 crate::sol! {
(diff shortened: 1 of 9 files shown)


legion2002:
superseded with #897

PR #974, opened 2026-08-03
fix(precompiles): lock down zone TIP20 administration

klkvr:
## Summary

- Reject external TIP20 admin and role mutations through exhaustive typed dispatch.
- Make ZoneInbox the pathUSD genesis admin; the zone --admin receives no pathUSD role.
- Keep bridge mint/burn on native internal paths and preserve ordinary token behavior.

Fixes [CYCLOPS-563](https://linear.app/tempoxyz/issue/CYCLOPS-563/cyclops-ungated-tip20changetransferpolicyid-writes-l1-mirrored-tip-403).

## Existing selector routing changes

- mint, mintWithMemo, burn, burnWithMemo: Inbox/Outbox-authorized external routes → rejected for every external EVM call. Deposits and withdrawals use direct native calls instead.
- distributeReward, setRewardRecipient, claimRewards, optedInSupply: upstream pass-through → Unauthorized. Zone rewards are deprecated.
- setSupplyCap, setLogoURI, pause, unpause, setNextQuoteToken, completeQuoteTokenUpdate, changeTransferPolicyId, burnBlocked: upstream pass-through → Unauthorized. Zones expose no token-admin operations; notably changeTransferPolicyId would write L1-mirrored TIP-403 state during zone execution.
- grantRole, revokeRole, renounceRole, setRoleAdmin: upstream pass-through → Unauthorized. Zone token roles are protocol-defined and immutable externally.
- hasRole: account privacy gate → public pass-through. Fixed role metadata does not expose private token activity.

Unchanged: balanceOf, allowance, and nonces remain privacy-gated; reward reads already denied remain denied; ordinary reads/transfers and getRoleAdmin pass through; malformed or unknown calldata keeps upstream dispatch; fixed-gas routing is unchanged.

## Tests

- cargo test -p zone-precompiles --lib
- cargo test -p tempo-xtask
- cargo fmt --all -- --check

### DIFF
--- a/crates/precompiles/src/ztip20/mod.rs
+++ b/crates/precompiles/src/ztip20/mod.rs
@@ -2,20 +2,16 @@
 //!
 //! `TIP20Token` remains the source of truth for token and TIP403 policy behavior.
 //! Before forwarding a call to Tempo, `TIP20Rules` applies only zone-specific checks:
-//! privacy-gated reads, fixed gas for selected selectors, and bridge mint/burn callers.
+//! privacy-gated reads, fixed gas for selected selectors, and blocked system/admin entry points.
 //!
 //! Accepted calldata and callers are forwarded unchanged to Tempo. Ordinary token state remains
 //! Zone-local while the EVM context's database adapter exposes selected policy values from the
 //! finalized Tempo L1 state.
 
 use alloy_primitives::Address;
-use alloy_sol_types::{SolCall, SolError};
-use tempo_precompiles::{
-    dispatch::selector_from_calldata,
-    tip20::{IRolesAuth, ITIP20},
-};
+use alloy_sol_types::{SolCall, SolError, SolInterface};
+use tempo_precompiles::tip20::{IRolesAuth, ITIP20};
 use tempo_zone_contracts::Unauthorized;
-use zone_primitives::constants::{ZONE_INBOX_ADDRESS, ZONE_OUTBOX_ADDRESS};
 
 use crate::{
     execution::{CallCheck, CallRules},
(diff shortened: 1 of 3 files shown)


crates/precompiles/src/ztip20/mod.rs
```diff
@@ -68,54 +57,84 @@ impl<P: L1StorageReader> CallRules for TIP20Rules<P> {
             .then_some(TIP20_FIXED_TRANSFER_GAS)
     }
 
-    /// Apply zone privacy and bridge-path checks before upstream execution.
+    /// Apply zone privacy and selector restrictions before upstream execution.
     fn admit(&self, data: &[u8], caller: Address) -> CallCheck {
-        let Some(selector) = selector_from_calldata(data) else {
+        if let Ok(call) = ITIP20::ITIP20Calls::abi_decode(data) {
```
mattsse, line 62: [P1] Meter calldata before decoding the interface

`admit` runs before `charge_input_cost` in `create_precompile`, but this eagerly ABI-decodes every recognized TIP20 call. `setLogoURI(string)` therefore allocates and copies attacker-controlled dynamic input before any precompile gas is charged. An EVM contract can reuse a large input buffer and repeatedly call this selector with insufficient callee gas, making every call do unmetered native work before it eventually returns out-of-gas. Please route restricted calls by selector (and only decode the bounded privacy arguments), or otherwise charge input before performing this full decode.

mattsse (commented):
One calldata-metering concern.

mattsse (dismissed):

mattsse (approved):

PR #984, opened 2026-08-04
fix(precompiles): preserve successful TIP-20 return data

decofe:
Fixes the successful TIP-20 return-data corruption described in CYCLOPS-844 by applying balance-error redaction only to reverted precompile outputs. Adds a regression test that stores an allowance whose ABI word starts with the upstream InsufficientBalance selector and verifies allowance() still returns all 32 bytes.

Validation: cargo +nightly fmt --check passed. The reported revm 41 build failure was fixed by using PrecompileOutput::is_revert(). The local targeted cargo test remains blocked because Cargo could not fetch the pinned alloy-evm revision; GitHub returned HTTP 401.

Prompted by: @0xrusowsky

### DIFF
--- a/crates/precompiles/src/lib.rs
+++ b/crates/precompiles/src/lib.rs
@@ -190,7 +190,7 @@ where
 {
     // Redacts TIP20 transfer from reverts that reveal user balances to the spender.
     let redact = |mut res: revm::precompile::PrecompileOutput| {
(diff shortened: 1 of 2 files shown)


0xrusowsky (dismissed):

0xrusowsky (approved):

legion2002 (approved):
makes sense

PR #989, opened 2026-08-04
fix(bench): align Tempo revision with ZoneFactory

Rjected:
Updates the Zones benchmark Tempo L1 pin to match the workspace dependency revision. This revision includes ZoneFactory leader initialization, preventing provisioning from rejecting a successful createZone transaction with a zeroed leader snapshot. The Tempo revision participates in the snapshot cache key, so the next benchmark run rebuilds the L1 baseline.


mattsse (approved):
LGTM

PR #990, opened 2026-08-04
ci(bench): track earn main in benchmarks

Rjected:
Updates the Zones benchmark workflow to check out `tempoxyz/earn` from `main`, matching the e2e test workflow.

Removes the benchmark-only Earn revision lock and its resolution and verification steps. The workflow derives the checked-out commit for Solidity cache keys, fixture metadata, and benchmark reports.


mattsse (approved):
LGTM

PR #991, opened 2026-08-04
fix(xtask): validate L1 anchor hash before createZone

decofe:
ref: `CYCLOPS-902`

## Summary

- Validate that the selected pre-creation L1 header RPC hash matches `TempoHeader::hash_slow()` before sending `ZoneFactory.createZone`.
- Add a regression test for matching and mismatched hashes.

## Validation

- `git diff --check`
- `cargo test -p tempo-xtask canonical_header_hash_must_match_rpc_hash` *(blocked before compilation: Cargo could not fetch the pinned `alloy-evm` revision; GitHub returned HTTP 401)*

Prompted by: @0xrusowsky


0xrusowsky:
cyclops audit fast

legion2002:
should be a won't fix, an honest reth L1 RPC wouldn't have this

PR #994, opened 2026-08-04
fix(ci): ignore unstable Solidity AST IDs in layout check

decofe:
Normalize compiler-generated Solidity storage-layout metadata before comparison so merge commits do not fail on unstable AST IDs.

The check still fails on slot, offset, label, or type changes.

Prompted by: @0xrusowsky


decofe:
Superseded by the direct CI fix on PR #967.

PR #1002, opened 2026-08-04
fix: Don't load full block only to redact transactions during block_by_hash

adityapk00:
## Summary

  Avoid loading private block bodies for redacted eth_getBlockByNumber and eth_getBlockByHash requests.

  ## Details

  Reth’s rpc_block(full=false) loads the recovered block and materializes every transaction hash before the Zone redaction layer discards them. This caused unnecessary body I/O,
  allocation, CPU work, and transaction-count-dependent timing.

  The redacted RPC now:

  - Reads only the sealed block header.
  - Synthesizes an empty transaction list.
  - Preserves the existing redaction behavior.
  - Keeps full=true restricted to sequencer-only handling.

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -32,7 +32,7 @@ use reth_rpc_eth_api::{
     helpers::{EthApiSpec, EthBlocks, EthCall, EthFees, EthState, EthTransactions, FullEthApi},
 };
 use reth_rpc_eth_types::logs_utils;
-use reth_storage_api::{BlockNumReader, StateProviderFactory};
+use reth_storage_api::{BlockNumReader, BlockReaderIdExt, StateProviderFactory};
 use tempo_alloy::{
     TempoNetwork,
     rpc::{TempoCallBuilderExt as _, TempoHeaderResponse, TempoTransactionRequest},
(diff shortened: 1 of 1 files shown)


crates/node/src/rpc.rs
```diff
@@ -737,36 +737,24 @@ where
     fn block_by_number(
         &self,
         number: BlockNumberOrTag,
-        full: bool,
+        _full: bool,
         _auth: AuthContext,
     ) -> BoxFut<'_> {
         Box::pin(async move {
-            let block = EthBlocks::rpc_block(&self.eth.api, number.into(), full)
```
mattsse, line 744: can we not simply set this to false always

crates/node/src/rpc.rs
```diff
@@ -737,36 +737,24 @@ where
...
-            redact_block(&mut block);
-
             to_raw(&block)
         })
     }
 
-    fn block_by_hash(&self, hash: B256, full: bool, _auth: AuthContext) -> BoxFut<'_> {
+    fn block_by_hash(&self, hash: B256, _full: bool, _auth: AuthContext) -> BoxFut<'_> {
         Box::pin(async move {
-            let block = EthBlocks::rpc_block(&self.eth.api, hash.into(), full)
```
mattsse, line 760: same here

mattsse (changes requested):
I'd rather keep using the regular rpc handler forwarding than reimplement this entirely.

since these are always fetched in full, it doesnt really matter if we forward the bool, but we can set them to false.

to remove code bloat we can introduce block_by_id as unifying function.

decofe:
Superseded by [#1003](https://github.com/tempoxyz/zones/pull/1003), which implements the requested `block_by_id` unification and `full=false` forwarding.

PR #1019, opened 2026-08-05
ci: normalize AST ids in storage layout check

0xrusowsky:
this PR normalizes AST ids in the solidity layout check so that we don't need to constantly regenerate when there are non-breaking changes


0xrusowsky:
cyclops audit fast

decofe (approved):
Approved: this is limited to normalizing unstable Solidity AST IDs in the storage-layout comparison, with explicit type-key collision detection. The required CI checks are passing.

grandizzy (approved):

PR #1022, opened 2026-08-05
fix(portal): preserve failed bounceback fees

decofe:
ref: `ZONE-317`

## Summary
- Count the bounce-back fee only when the admin transfer succeeds.
- Refund or park the full amount when the fee transfer fails.
- Emit the actual collected fee and document the accounting semantics.
- Add regression coverage for failed fee and refund transfers.

## Validation
- `forge fmt --check`
- `git diff --check`
- `forge test --match-path test/tempo/ZonePortalGasLimit.t.sol` *(blocked: GitHub rejected missing forge-std and tempo-std submodule downloads with invalid credentials)*

Prompted by: @0xrusowsky


legion2002:
cyclops audit fast

legion2002 (approved):
lgtm, will create an associated tempo bump PR

PR #1023, opened 2026-08-05
fix(rpc): validate redacted fee history percentiles

decofe:
## Summary

- Inline Reth-equivalent `rewardPercentiles` count, `[0, 100]` range, and nondecreasing-order checks before constructing the redacted reward matrix.
- Document why the checks are repeated locally: passing `None` to Reth avoids loading private block bodies and bypasses its validation.
- Add an end-to-end regression test proving 101 percentiles return `-32602` before allocation.

Addresses ZONE-415.

## Testing

- `git diff --check`
- `cargo test -p zone-node --test it redacted_rpc_e2e::test_fee_history_rejects_excessive_reward_percentiles --no-run` *(blocked before compilation: Cargo could not fetch the pinned `alloy-evm` revision; GitHub returned HTTP 401)*

Prompted by: aditya

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -716,14 +716,26 @@ where
                 .await
                 .map_err(internal)?;
             if let Some(reward_percentiles) = reward_percentiles.as_deref() {
-                validate_reward_percentiles(
-                    reward_percentiles,
-                    self.eth
+                // These checks mirror upstream Reth's `eth_feeHistory` validation. We repeat
+                // them here because passing `None` above avoids loading private block bodies,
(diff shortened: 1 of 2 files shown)


adityapk00:
cyclops audit fast

adityapk00:
The cyclops issue is already tracked: [ZONE-427]

legion2002 (commented):
we could do this, but tbh feels pretty messy to duplicate reth logic like this upstream

would prefer to just revert the #1001  PR instead, as it was just a nit anyways. We shouldn't care about RPC latency based leaks

decofe:
Closing this PR because timing attacks on RPC are wont fix.

PR #1026, opened 2026-08-05
fix(rpc): restore fee history reward handling

decofe:
## Summary

- Revert PR #1001 by passing `reward_percentiles` back to Reth’s `eth_feeHistory` implementation.
- Restore redaction of the returned reward values while preserving upstream validation and response sizing.

Timing attacks on RPC are wont fix, so the bespoke side-channel mitigation from #1001 should be removed.

## Testing

- `git diff --cached --check`

Reverts #1001.

Prompted by: aditya

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -709,17 +709,10 @@ where
         reward_percentiles: Option<Vec<f64>>,
     ) -> BoxFut<'_> {
         Box::pin(async move {
-            // Avoid loading private block bodies to calculate reward percentiles.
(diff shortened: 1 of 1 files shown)


adityapk00:
Disucssion Summary: The original PR #1001 was fixed but it caused us to skip the upstream reth checks that created more resource exhaustion. Reverting #1001 because the checks are bounded and the callers to the RPC are anyway checked and timing side channel attacks via RPC are not reliable/feasible

adityapk00 (approved):

PR #1030, opened 2026-08-06
docs(ref-impl): support batched Tempo imports

0xrusowsky:
## Summary

This PR updates the Solidity protocol/reference contracts to support importing a contiguous range of Tempo headers in one Zone `advanceTempo` call as introduced in https://github.com/tempoxyz/zones/pull/981.

It intentionally contains only the protocol/reference-contract side of the change. The native Rust precompile, node, payload-builder, queue, replication, and generated Zone genesis updates will follow in a separate implementation PR.


0xrusowsky:
cyclops audit fast

PR #1037, opened 2026-08-06
fix(replication): preserve durable leader handoffs

Rjected:
Completes the persisted-only replication invariant during leader shutdown and a responder that has not reached the requested backfill range.

It waits for the leader engine to stop, then broadcasts each final block only after it persists. If a responder completes below the requested start, the follower retains its catch-up state and immediately falls back from the preferred leader.

Fixes ZONE-265.

### DIFF
--- a/crates/node/src/role.rs
+++ b/crates/node/src/role.rs
@@ -24,7 +24,10 @@ use reth_provider::HeaderProvider;
 use reth_storage_api::{BlockNumReader, BlockReader, ReceiptProvider, StateProviderFactory};
 use reth_transaction_pool::TransactionPool;
 use tempo_primitives::{Block, TempoHeader};
-use tokio::{sync::mpsc, task::JoinSet};
+use tokio::{
+    sync::{mpsc, oneshot},
+    task::JoinSet,
+};
 use tokio_util::{sync::CancellationToken, task::AbortOnDropHandle};
 use tracing::{debug, error, info, warn};
 use zone_chainspec::ZoneChainSpec;
@@ -47,7 +50,7 @@ mod zone_transaction_pool_alias {
 use crate::{
     EngineExit, ProductionPermit, ZoneEngine, ZoneSequencerAddOnsConfig,
     replication::{
-        AttestationContext, PeerTipRegistry, broadcast_persisted_blocks,
+        AttestationContext, BroadcasterShutdown, PeerTipRegistry, broadcast_persisted_blocks,
         collect_follower_settlement_signatures, run_follower_block_sync,
     },
     settlement_attestation::collect_leader_settlements,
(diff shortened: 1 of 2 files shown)


adityapk00:
cyclops audit fast

adityapk00:
The backfill responder stuff wasmerged in #1017

crates/node/src/role.rs
```diff
@@ -368,6 +374,36 @@ impl RunningGeneration {
...
+        if let Some(engine_done) = self.engine_done.take() {
+            match tokio::time::timeout_at(deadline, engine_done).await {
+                Ok(Ok(())) | Ok(Err(_)) => {
+                    if let Some(drain) = self.broadcast_drain.take() {
+                        let _ = drain.send(());
+                    } else if let Some(stop) = self.broadcast_stop.take() {
+                        stop.cancel();
+                    }
+                }
+                Err(_) => {
```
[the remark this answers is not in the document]
adityapk00: fixed

crates/p2p/src/backfill.rs
```diff
@@ -418,9 +448,24 @@ where
                     .map_err(|_| eyre::eyre!("backfill response event channel closed"))?;
             }
             ResponseFrame::Complete { request_id, tip } => {
-                if !self.job.complete(&peer, request_id, received_at) {
+                let Some(request) = self.job.complete(&peer, request_id, received_at) else {
                     warn!(target: "zone::p2p", %peer, request_id, "Ignoring unsolicited or stale backfill completion");
                     return Ok(());
+                };
+                if tip.zone_height < request.start {
```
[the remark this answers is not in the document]
adityapk00: This is an issue with #1017 and needs a follow up PR

adityapk00 (dismissed):

crates/node/src/role.rs
```diff
@@ -360,14 +373,80 @@ struct RunningGeneration {
     id: u64,
     role: DesiredRole,
     token: CancellationToken,
+    /// Fires when the leader engine task returns.
+    engine_done: Option<oneshot::Receiver<()>>,
```
legion2002: Could we bundle these leader-only shutdown handles into one optional struct, with a single `Drain`/`Stop` command for the broadcaster? That would make partial states impossible and simplify `stop`.

crates/node/src/role.rs
```diff
@@ -586,8 +674,11 @@ pub(crate) async fn run_role_controller<P, Pool>(
             .as_ref()
             .is_some_and(|generation| generation.role.same_variant(desired))
         {
-            if let Some(generation) = current.take() {
-                generation.stop(&sinks).await;
+            if let Some(generation) = current.take()
+                && generation.stop(&sinks).await == GenerationStopOutcome::Failed
```
legion2002: Think this block appears multiple times below, can we extract this into a helper

legion2002 (approved):
lgtm

crates/node/src/role.rs
adityapk00: fixed

crates/node/src/role.rs
adityapk00: fixed

crates/p2p/src/backfill.rs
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/zones/pull/1097

PR #1041, opened 2026-08-07
fix(node): confirm zone leadership transfers

decofe:
## Summary

- refetch the committed admin nonce lane before every zone leadership transfer
- set the nonce and Tempo fee fields explicitly so retries replace a missing transaction instead of creating a nonce gap
- wait for a successful receipt before reporting the transfer as submitted

## Validation

- `cargo check -p zone-node --lib`
- `cargo test -p zone-node rpc::tests --lib` (11 passed)
- `cargo +nightly clippy -p zone-node --all-targets -- -D warnings` *(blocked by existing `semicolon-in-expressions-from-non-local-macros` errors in `crates/l1/src/subscriber.rs`)*

Prompted by: @kamsz

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -38,6 +38,7 @@ use reth_storage_api::{BlockNumReader, StateProviderFactory};
 use reth_trie_common::ExecutionWitnessMode;
 use tempo_alloy::{
     TempoNetwork,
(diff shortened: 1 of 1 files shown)


rkrasiuk (approved):
lgtm pending @adityapk00

mattsse (approved):
lgtm

PR #1052, opened 2026-08-07
refactor(rpc): remove redacted RPC self-connection

mattsse:
## Summary

- remove the operator HTTP URL and self-connected zone provider from redacted RPC construction
- read `TempoState.tempoBlockNumber` from the EthApi's local canonical state
- create an L1 provider for the redacted RPC server during node launch
- keep `RedactedRpcConfig` focused on server and authentication configuration

## Why

The redacted RPC server connected back to its own operator RPC solely to read the local Tempo block number. That value is already available from the EthApi's local state. The redacted RPC still needs L1 access for zone metadata, so node launch creates a provider and passes it to `ZoneRpc`.

## Impact

Redacted RPC behavior, authentication, privacy enforcement, the handwritten dispatcher, WebSocket transport, and `ZoneProvider` remain unchanged. The change removes the self-connection and uses one L1 connection owned by the redacted RPC server.


0xrusowsky:
cyclops audit fast

klkvr (approved):

0xrusowsky (approved):

PR #1054, opened 2026-08-07
refactor(sequencer): simplify withdrawal task lifecycle

decofe:
## Summary

- remove the startup-log-only L1 RPC URL from `WithdrawalProcessorConfig`
- run one withdrawal processor until shutdown instead of wrapping it in unreachable restart supervision
- retain per-cycle error logging/retry behavior and poll-boundary graceful shutdown
- use shared references for the processor loop, queue processing, and metrics helpers

## Validation

- `cargo test -p zone-sequencer withdrawals::tests`
- `cargo check -p zone-sequencer`
- `cargo +nightly clippy -p zone-sequencer --lib -- -D warnings`

Prompted by: @mattsse


mattsse (approved):
LGTM

PR #1056, opened 2026-08-07
refactor(sequencer): simplify withdrawal queue reconciliation

mattsse:
## Summary

- Replace repeated suffix rehashing during withdrawal queue reconciliation with one backward hash-chain walk.
- Inline the remaining queue suffix hash at the withdrawal submission call site and remove the redundant wrapper.
- Consolidate reconciliation offset cases into table-driven coverage.

## Why

`find_processed_offset` previously recomputed the queue hash for every candidate suffix. Those suffixes overlap, so the helper repeatedly hashed the same withdrawals and took O(n²) work. Reusing the tail hash while walking backward makes reconciliation O(n).

## Impact

Protocol behavior is unchanged: full, partial, fully consumed, empty, corrupted, and sentinel-related queue states retain their existing semantics, and an empty submitted suffix still hashes to `B256::ZERO`.

### DIFF
--- a/crates/sequencer/src/settlement.rs
+++ b/crates/sequencer/src/settlement.rs
@@ -27,7 +27,10 @@ use std::{collections::BTreeMap, fmt, sync::OnceLock};
 
 use crate::{
     ZoneSequencerProvider,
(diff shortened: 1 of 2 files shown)


0xrusowsky:
cyclops audit fast

0xrusowsky (approved):

PR #1057, opened 2026-08-07
refactor(rpc): remove phantom params and fix portal-less zone info

mattsse:
Cleans up the zone RPC surface in three parts:

- Fixes zone_getZoneInfo erroring on nodes running without an L1 portal: the zero-portal case is now handled once with static defaults instead of per-read guards, one of which was missing. The token list now comes from the node's synced enabled-token registry instead of a separate L1 read, matching the log-scoping path.

- Removes parameters from ZoneRpcApi that implementations can never receive (state overrides, full block bodies) — the dispatch layer rejects both before any trait method runs. Signatures now match the actual contract; rejection behavior is unchanged.

- Corrects docs that described non-existent sequencer bypasses, and replaces the stringly-typed zone_getSequencerInfo mode and zone_setLeader status fields with enums. Wire format is unchanged.

### DIFF
--- a/crates/contracts/src/precompiles/zone_portal.rs
+++ b/crates/contracts/src/precompiles/zone_portal.rs
@@ -340,6 +340,24 @@ impl<P: alloy_provider::Provider<N>, N: alloy_network::Network>
         futures::future::try_join_all(futs).await
     }
 
+    /// Returns all sequencer addresses currently registered on this [`ZonePortal`].
+    ///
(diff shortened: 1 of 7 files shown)


mattsse:
Closing: the rpc crate surface this touches is being restructured, and the one real fix (portal-less zone_getZoneInfo defaults + registry-sourced token list) will be folded into that work instead.

PR #1064, opened 2026-08-07
fix: prevent unauthorized encrypted deposit payload replay

0xrusowsky:
## Summary

   - Bind encrypted deposit key derivation to the portal and deposit sender,
     preventing payloads from being replayed by another account.
   - Derive one-time, portal-scoped sender tags from client-generated sender
     witnesses.
   - Bind routed deposit ciphertexts and replay nullifiers to authenticated source
     withdrawals.
   - Preserve the existing deposit gas capacity by deriving sender tags instead of
     duplicating them in pending withdrawal storage.

   ## Security properties

   - A direct encrypted deposit payload cannot authenticate under a different
     deposit sender.
   - A routed payload cannot authenticate for another source withdrawal.
   - A source withdrawal cannot be consumed by the router more than once.


0xrusowsky:
closing for now as zones will launch in a closed loop setup

PR #1065, opened 2026-08-07
spec: prevent unauthorized encrypted deposit payload replay

0xrusowsky:
## Summary

- bind direct encrypted deposit key derivation to the portal and public deposit sender
- define portal-scoped, one-time sender commitments using a client-generated `senderWitness`
- derive a router-scoped source-withdrawal identity for both replay nullification and routed GCM nonce binding
- update the protocol invariants, Solidity reference implementation, and reference tests

## Motivation

An encrypted deposit payload must not authenticate when copied into a different authorization context. Direct deposits therefore bind the public deposit sender into HKDF, while routed deposits bind their ciphertext and nullifier to the authenticated source portal and sender tag.

## Validation

- `forge fmt --check`
- `forge test` — 680 passed


0xrusowsky:
closing for now as zones will launch in a closed loop setup

PR #1070, opened 2026-08-10
fix(precompiles): reject non-canonical deposit encodings

decofe:
## Summary

- require queued deposits to match their canonical ABI encoding
- reject trailing bytes and other non-canonical encodings before deposit execution
- cover the encrypted-only deposit format with a regression test

Addresses [ZONE-257](https://linear.app/tempoxyz/issue/ZONE-257/cyclops-advancetempo-deposit-calldata-has-two-different-acceptance).

## Tests

- `cargo test -p zone-precompiles`
- `cargo +nightly clippy -p zone-precompiles --all-targets -- -D warnings`

Prompted by: @0xrusowsky

### DIFF
--- a/crates/precompiles/src/inbox/mod.rs
+++ b/crates/precompiles/src/inbox/mod.rs
@@ -20,7 +20,7 @@ use alloc::vec::Vec;
 
 use alloy_evm::precompiles::DynPrecompile;
 use alloy_primitives::{Address, B256, U256};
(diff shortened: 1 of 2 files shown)


0xrusowsky (approved):

klkvr (approved):

klkvr:
cyclops audit

PR #1072, opened 2026-08-10
fix(ci): pass Git SHA to reproducible build

decofe:
## Summary

- pass the checked-out commit SHA into the reproducible Docker build
- use that SHA when vergen cannot inspect Git metadata because .git is excluded
- preserve the existing no-.git reproducible build context

## Validation

- bash -n scripts/reproducible-build.sh
- cargo +nightly fmt --all -- --check
- cargo check --locked -p zone-node from a source copy without .git

The full wrapper could not run locally because the sandbox Docker installation lacks Buildx support for --output.

Prompted by: @grandizzy


grandizzy:
cyclops audit super-fast

grandizzy (approved):

PR #1073, opened 2026-08-10
refactor(checker): reuse shared protocol surfaces

decofe:
## Summary

- remove checker-owned protocol ABI, constants, storage layouts, and manual slot decoding
- read exact-block state through Tempo read-only storage contexts and generated Zone handlers
- decode shared event/call wire types while retaining independent topics, bounds, canonicalization, commitments, fees, and state-transition verification
- move broad runtime and kernel-boundary tests out of root source modules and consolidate obsolete ABI coverage

## Validation

- `cargo +nightly clippy -p zone-checker --all-targets -- -D warnings`
- `cargo test -p zone-checker -- --test-threads=1` (116 passed)

Prompted by: @grandizzy


decofe:
Closing per requester; these commits will be pushed directly to `horsefacts/zone-checker`.

PR #1077, opened 2026-08-10
ci: sync private copies from upstream

kuyziss:
## Summary

- add an hourly, manually dispatchable upstream-sync workflow for forks and private copies of `tempoxyz/zones`
- authenticate mirror pushes with a GitHub App token that can update both contents and workflow files
- replace the long-lived `EARN_DEPLOY_KEY` checkout credential with a short-lived STS token in test and benchmark workflows
- skip the mirror job in the public upstream repository

## Why

A private mirror must retain the sync workflow after each force-reset to upstream `main`. STS grants each job only temporary read access to `tempoxyz/earn`, avoiding a deploy key in Zones and its private mirror.

## Dependency

Requires `tempoxyz/earn` PR for the `zones-read` STS policy to merge first.

## Validation

- `git diff --check`
- `actionlint` (with the existing custom benchmark runner label ignored)
- STS policy YAML parse

### DIFF
--- a/.github/workflows/sync-from-upstream.yml
+++ b/.github/workflows/sync-from-upstream.yml
@@ -1,0 +1,38 @@
+# This workflow is intended for forks or private copies of tempoxyz/zones.
+# It keeps their main branch aligned with the upstream repository every hour.
+
+name: Sync main branch with upstream
+
+on:
+  schedule:
+    - cron: "0 * * * *"
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


kamsz (approved):

jenpaff (approved):

laibe (approved):

kuyziss:
cyclops audit super-fast

PR #1086, opened 2026-08-10
perf(spf): compose `ZoneChainSpec` once

shekhirin:
## Problem

Creating `ZoneEvmConfig` is expensive because [`compose_chain_spec`](https://github.com/tempoxyz/zones/blob/153d466df2935005432bd80c34243ff6ec6d9cad/crates/evm/src/lib.rs#L560-L563) clones the whole `ChainSpec`.
https://github.com/tempoxyz/zones/blob/153d466df2935005432bd80c34243ff6ec6d9cad/crates/evm/src/lib.rs#L560-L563
image

## Solution

Prepare `ZoneChainSpec` once, and clone as an `Arc` for `ZoneEvmConfig` creations.


rkrasiuk (approved):
lgtm

PR #1095, opened 2026-08-10
refactor: simplify leader shutdown coordination

legion2002:
## Summary

- bundle leader-only shutdown handles into one `LeaderShutdown` state
- replace separate broadcaster stop/drain signals with one typed shutdown command
- centralize the controller's repeated stop-and-fence sequence

## Why

This is a behavior-preserving cleanup stacked on #1037. It makes partial leader shutdown states unrepresentable and keeps the fail-closed teardown policy in one place.

## Validation

- `rustup run nightly cargo clippy -p zone-node --all-targets -- -D warnings`
- `rustup run nightly cargo test -p zone-node --lib` (56 passed)
- `rustup run nightly cargo test -p zone-node --test advance_tempo` (3 passed after building the required Foundry artifacts)
- `git diff --check`

### DIFF
--- a/crates/node/src/replication.rs
+++ b/crates/node/src/replication.rs
@@ -19,7 +19,7 @@ use std::{
 };
 use tempo_alloy::TempoNetwork;
 use tempo_primitives::{Block, TempoHeader, TempoTxEnvelope};
-use tokio::sync::mpsc;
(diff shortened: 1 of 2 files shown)


legion2002:
already fixed in base PR

PR #1097, opened 2026-08-10
fix(p2p): Fix off by one causing spurious backfill requests

adityapk00:
## Summary

  Fix backfill handling for responders that are fully synced with the requester.

  Production backfill requests begin at local_best + 1. A responder at the same
  height therefore completes with tip.zone_height == request.start - 1. This
  valid empty completion was incorrectly classified as coming from a behind peer.

  ## Root cause

  The coordinator previously used:

  tip.zone_height < requested_start

  For a requester and responder both at height N, this evaluates as:

  tip = N
  requested_start = N + 1
  N < N + 1 // true

  The completion was suppressed, the responder was placed on cooldown, and the
  request was fanned out to other peers. Because no completion reached
  BackfillProgress, fully synced followers could continue retrying indefinitely.

  ## Fix

  Only classify a responder as behind when its tip is below the requester’s
  current head:

  tip.zone_height.saturating_add(1) < requested_start

  A responder at requested_start - 1 now follows the normal Completed path,
  allowing the follower to record completion and stop retrying.

  ## Testing

  Expanded the P2P integration test to verify:

  - A level responder completing at requested_start - 1 produces
    BackfillResponse::Completed.

  - The level completion does not trigger fallback fanout to another peer.
  - Non-empty backfill responses still deliver all blocks before completion.
  - Replaying a completed request ID is still rejected.


legion2002 (approved):
makes sense

PR #1099, opened 2026-08-10
fix: bound private RPC response sizes

klkvr:
## Summary

- apply reth's configured `rpc.max-response-size` to private HTTP RPC responses
- stop serialization at the per-response byte limit and return jsonrpsee-compatible `-32008` errors
- enforce the same limit across batch responses, returning `-32011` and stopping subsequent dispatch
- avoid introducing private-only log-count or work limits

## Testing

- `cargo +nightly clippy -p zone-rpc --all-targets -- -D warnings`
- `cargo +nightly clippy -p zone-node --lib -- -D warnings`
- `cargo +nightly test -p zone-rpc response_ --lib`

Fixes ZONE-201

### DIFF
--- a/crates/node/src/node.rs
+++ b/crates/node/src/node.rs
@@ -758,6 +758,12 @@ where
         }
 
         let chain_id = ctx.node.provider().chain_spec().genesis().config.chain_id;
+        let max_response_size = ctx
+            .config
+            .rpc
+            .rpc_max_response_size
+            .get()
+            .saturating_mul(1024 * 1024) as usize;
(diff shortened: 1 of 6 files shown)


klkvr:
cyclops audit fast

mattsse (approved):
Reviewed as scoped to private HTTP RPC responses. The configured response-size propagation, bounded serialization, and aggregate batch handling look correct. WebSocket response parity can be handled separately.

mattsse (approved):

PR #1101, opened 2026-08-10
ci: allow transitive bitmaps advisory

mattsse:
Cargo deny now rejects the locked `bitmaps 3.2.1` dependency because RUSTSEC-2026-0247 marks every release unmaintained. The crate is pulled in transitively by `reth-transaction-pool` through `imbl`, and the advisory provides no safe upgrade.

Add a documented temporary advisory exception so repository lint remains usable until the upstream dependency chain replaces `bitmaps`. This keeps the repository-wide CI fix separate from the unrelated xtask changes in #1088.


0xrusowsky (approved):

legion2002:
cyclops audit super-fast

legion2002 (approved):

PR #1102, opened 2026-08-10
Allow zone admins to update the leader

legion2002:
## Summary

- allow `ZonePortal.setLeader` to be called by the current portal admin as well as any registered sequencer
- keep the new leader restricted to the active sequencer set
- preserve the existing unauthorized-caller revert behavior and document the expanded caller set
- add focused coverage for admin-authorized transitions and unrelated-account rejection

## Why

Leader rotation was previously relayable only by registered sequencer keys. Allowing the zone admin key to submit the same compare-and-set transition provides an administrative recovery path while retaining the existing target membership, epoch fencing, and one-transition-per-block checks.

## Validation

- `forge fmt --check`
- `forge test --match-path test/tempo/ZonePortal.t.sol --match-test 'test_setLeader' -vv` (8 passed)
- `forge test --match-path test/tempo/ZonePortal.t.sol -q`
- `git diff --check`

### DIFF
--- a/specs/ref-impls/src/interfaces/IZone.sol
+++ b/specs/ref-impls/src/interfaces/IZone.sol
@@ -774,8 +774,8 @@ interface IZonePortal {
     function leaderActivationTempoBlock() external view returns (uint64);
 
     /// @notice Transfer block-production leadership to another sequencer-set member.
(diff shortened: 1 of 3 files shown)


legion2002:
cyclops audit fast

0xKitsune (approved):

PR #1107, opened 2026-08-10
fix(sequencer): submit current-tip batches immediately

0xKitsune:
This PR submits batches immediately instead of waiting for the next retry when `current_block == anchor_block`, and reduces the fallback settlement retry from five seconds to 500ms.


mattsse (approved):
LGTM

PR #1109, opened 2026-08-10
feat(xtask): add sequencer set rotation

decofe:
## Summary

- add a `rotate-sequencer-set` xtask command
- retain the active leader in a joint set before handing leadership off
- use the portal admin key for both set updates and the `setLeader` handoff
- validate the admin key on-chain and verify final state
- make interrupted rotations resumable

## Why

This stacks on #1102, which authorizes the portal admin to call `setLeader`; the rotation therefore requires only `ADMIN_KEY`.

## Validation

- `cargo test -p tempo-xtask rotate_sequencer_set`
- `cargo run -p tempo-xtask -- rotate-sequencer-set --help`
- `git diff --check`

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -6,7 +6,8 @@ use crate::{
     deploy_router::DeployRouter, deposit::Deposit, generate_p2p_key::GenerateP2pKey,
     generate_zone_genesis::GenerateZoneGenesis,
     install_reference_zone_factory::InstallReferenceZoneFactory,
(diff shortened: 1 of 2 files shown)


0xalpharush:
cyclops audit super-fast

decofe:
Operationally superseded by #1180 and #1181, which implement guarded, one-at-a-time identity and sequencer-membership rotation. They replace the intended rotation workflow, though not this PR's identical arbitrary whole-set CLI.

PR #1113, opened 2026-08-10
feat(xtask): add portal access lockdown

0xalpharush:
## Summary

- Add a `lock-down-access` incident-response xtask.
- Close access enforcement, reconstruct account membership from `RoleUpdated` logs, and revoke current account roles.
- Submit bounded concurrent reads and explicitly-nonced revocations; document safe backfill use.

## Validation

- `cargo run -p tempo-xtask -- lock-down-access --help`
- `git diff --check`

A full local native-L1 E2E attempt was interrupted at requester direction after the standalone dev binary hit an unrelated `OpcodeNotFound` provisioning incompatibility.


0xalpharush:
Superseded by #1114, which contains only the portal access lockdown changes.

PR #1114, opened 2026-08-10
feat(xtask): add portal access lockdown

0xalpharush:
## Summary

- Add a `lock-down-access` incident-response xtask.
- Close access enforcement, reconstruct account membership from `RoleUpdated` logs, and revoke current account roles.
- Submit bounded concurrent reads and explicitly-nonced revocations; document safe backfill use.

## Validation

- `cargo test -p tempo-xtask`
- `cargo +nightly clippy -p tempo-xtask --all-targets`
- `cargo run -p tempo-xtask -- lock-down-access --help`
- `git diff --check`

A full local native-L1 E2E attempt was interrupted at requester direction after the standalone dev binary hit an unrelated `OpcodeNotFound` provisioning incompatibility.

Prompted by: @0xalpharush


0xalpharush:
cyclops audit super-fast

legion2002 (commented):
think we don't need this anymore since we added the pause button on the portal, and the pause-portal xtask

PR #1121, opened 2026-08-11
refactor: derive zone identity from chain ID

klkvr:
## Summary

- derive the parent Tempo chain ID and zone ID from the genesis `chainId`, including the reserved mainnet and Moderato ranges
- validate the derived zone ID against `ZonePortal.zoneId()` and use it for node, RPC, P2P, sequencing, and SPF configuration
- remove manifest identity as a source of truth while continuing to accept legacy `zone_id` and `sequencer_set_version` fields
- keep `zone_getSequencerInfo` compatibility fields populated from the chain-derived zone ID and startup-pinned Portal sequencer-set version
- update admin invariants and sequencer-set validation to stop reading removed manifest identity fields
- pin the Portal sequencer-set version during startup membership validation and fail closed if it changes before settlement
- retain `--zone.id` / `ZONE_ID` as a deprecated compatibility check instead of removing it

## Verification

- `cargo check -p zone-primitives -p zone-chainspec -p zone-p2p -p zone-node`
- `cargo test -p zone-primitives -p zone-chainspec -p zone-p2p -p zone-node --lib`
- `cargo test -p zone-node --features cli cli::tests::deprecated_zone_id_is_optional_and_validated -- --exact`
- `cargo +nightly clippy -p zone-primitives -p zone-chainspec -p zone-p2p -p zone-node -p zone-sequencer --all-features --tests` (passes with pre-existing warnings)
- `cargo check --workspace --all-features --locked`
- `cargo test -p tempo-xtask --locked`
- `cargo test -p zone-node operator_rpc_module_exposes_sequencer_methods_without_auth --locked`
- `cargo test -p zone-node settlement_attestation::tests::settlement_rejects_runtime_sequencer_set_rotation --locked -- --exact`
- `cargo test -p zone-node settlement_attestation::tests::synthetic_test_harness_can_skip_an_unconfigured_portal --locked -- --exact`
- `cargo +nightly clippy -p zone-node --all-targets --all-features --locked` (passes with pre-existing warnings)
- `cargo +nightly clippy --all-targets --all-features --locked` (passes with pre-existing warnings)
- `git diff --check`

Prompted by: @klkvr

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -86,6 +86,8 @@ pub struct SequencerRpcContext {
     pub(crate) peer_tips: PeerTipRegistry,
     /// Validated static topology manifest.
     pub manifest: Arc<ZoneManifest>,
+    /// Portal sequencer-set version validated against the manifest at startup.
+    pub pinned_sequencer_set_version: Option<u64>,
     /// This node's individual secp256k1 address (the `setLeader` relayer identity).
     ///
     /// `None` on an rpc-only member: it holds no individual key, so it cannot relay.
@@ -105,6 +107,7 @@ impl SequencerRpcContext {
         status: SharedRoleStatus,
         peer_tips: PeerTipRegistry,
         manifest: Arc<ZoneManifest>,
+        pinned_sequencer_set_version: Option<u64>,
         local_secp256k1_address: Option<Address>,
         local_ed25519_public_key: zone_p2p::P2pPeerId,
         relayer: Option<DynProvider<TempoNetwork>>,
@@ -115,6 +118,7 @@ impl SequencerRpcContext {
             status,
             peer_tips,
             manifest,
+            pinned_sequencer_set_version,
             local_secp256k1_address,
             local_ed25519_public_key,
             relayer,
@@ -196,6 +200,7 @@ where
 
 /// Build the unauthenticated Zone extension installed on the node's operator HTTP RPC.
 pub(crate) fn operator_zone_rpc_module<P>(
+    zone_id: u32,
     portal_address: Address,
     sequencer: Arc<std::sync::OnceLock<SequencerRpcContext>>,
     provider: P,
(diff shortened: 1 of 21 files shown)


klkvr:
cyclops audit

adityapk00:
Looks pretty good and makes sense. I think there might be a merge conflict.

klkvr:
cyclops audit

crates/node/src/settlement_attestation.rs
```diff
@@ -213,11 +213,6 @@ where
         verifier_call.call(),
         portal_tip_call.call(),
     )?;
-    eyre::ensure!(
-        set_version == context.domain.sequencer_set_version,
-        "portal signer-set version {set_version} does not match manifest version {}",
-        context.domain.sequencer_set_version
-    );
     eyre::ensure!(
```
0xrusowsky, line 232: is this relevant? @klkvr

0xrusowsky (commented):
LGTM besides the unresolved cyclops finding (which may be overstated)

crates/node/src/rpc.rs
```diff
@@ -487,11 +499,16 @@ where
     let local_node = context
         .manifest
         .node_by_ed25519_public_key(&context.local_ed25519_public_key);
+    let sequencer_set_version = ZonePortal::new(portal_address, l1_provider)
```
adityapk00: zone_getSequencerInfo should report the node’s startup-pinned sequencer-set version rather than re-reading the live value from ZonePortal.

Currently, a node can pin version 7 at startup, observe the Portal advance to version 8, and refuse settlement because its pinned version is stale—while this RPC reports version 8. Admin invariants can therefore incorrectly conclude that the node loaded the current version. 

Recommendation: add pinned_sequencer_set_version to SequencerRpcContext in crates/node/src/rpc.rs, pass it into SequencerRpcContext::new from the value already obtained in crates/node/src/node.rs, and return that value as manifest_sequencer_set_version.

adityapk00 (dismissed):

adityapk00 (approved):

0xrusowsky (approved):

PR #1132, opened 2026-08-11
feat(xtask): derive zone genesis anchor from portal

decofe:
## Summary
- derive the genesis anchor from the canonical finalized `ZoneCreated` block
- make `create-zone` wait for finality and use the same `creation block - 1` anchor as regeneration
- require an explicit `--l1-rpc-url` when deriving from `--tempo-portal`
- verify event, deployment-code, header-number, and header-hash post-conditions before writing genesis

## Validation
- `cargo +nightly clippy -p tempo-xtask --all-targets --no-deps` (passes with pre-existing future-incompatibility warnings)

Prompted by: @kuyziss

### DIFF
--- a/xtask/src/generate_zone_genesis.rs
+++ b/xtask/src/generate_zone_genesis.rs
@@ -1,8 +1,13 @@
 use alloy::{
+    consensus::BlockHeader as _,
     genesis::{ChainConfig, Genesis, GenesisAccount},
-    primitives::{Address, Bytes, U256, address},
+    primitives::{Address, B256, Bytes, U256, address, keccak256},
+    providers::{Provider, ProviderBuilder},
 };
-use eyre::{WrapErr as _, eyre};
+use alloy_eips::BlockNumberOrTag;
+use alloy_rlp::Encodable;
+use alloy_rpc_types_eth::BlockId;
+use eyre::{WrapErr as _, ensure, eyre};
 use reth_evm::{
     Evm as _, EvmEnv, EvmFactory,
     revm::{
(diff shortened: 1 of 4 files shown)


kuyziss:
cyclops audit fast

xtask/src/generate_zone_genesis.rs
```diff
@@ -311,6 +320,63 @@ impl GenerateZoneGenesis {
...
+            .wrap_err_with(|| format!("failed to fetch portal code at Tempo block {mid}"))?;
+        if code.is_empty() {
+            low = mid + 1;
+        } else {
+            high = mid;
+        }
+    }
+
+    let creation_block = low;
+    let anchor_block = creation_block.checked_sub(1).ok_or_else(|| {
```
0xalpharush, line 375: can use https://github.com/tempoxyz/zones/blob/b430c7f0ecc367ce7a036a9e271a6675b1d082f6/xtask/src/verify_closed_loop.rs#L293
once merged

xtask/src/generate_zone_genesis.rs
```diff
@@ -68,11 +71,16 @@ pub(crate) struct GenerateZoneGenesis {
     #[arg(long)]
     pub(crate) tempo_portal: Address,
 
+    /// Tempo L1 HTTP RPC URL used to derive the pre-creation genesis anchor.
+    #[arg(long, default_value = "https://rpc.moderato.tempo.xyz")]
+    pub(crate) l1_rpc_url: String,
```
klkvr: feels a bit odd to have this default to testnet? especially given that portal addresses are identical on both testnet and mainnet? let's require passing this explicitly? or default to mainnet at least

klkvr (dismissed):

xtask/src/generate_zone_genesis.rs
decofe: Done — portal-based derivation now requires an explicit --l1-rpc-url; there is no Moderato default on this path.

xtask/src/generate_zone_genesis.rs
decofe: Done — this now reuses the merged event-based ZoneCreated lookup from verify-closed-loop, moved into the shared xtask utilities.

klkvr (approved):

PR #1133, opened 2026-08-11
ci(bench): remove unsupported private transfer

decofe:
## Summary

- remove the direct TIP-20 transfer and receipt steps from the full neobank journey
- remove the unused private transfer workload template and recipient binding
- complete the journey from the supported onramp, Earn, and offramp paths

## Validation

- `git diff --check`
- parsed both edited scenario files after environment placeholder substitution

Prompted by: @shekhirin


0xrusowsky (approved):
note: this is only temporary. transfers will be re-enabled on phase 2

PR #1140, opened 2026-08-11
fix(rpc): stabilize private log indices

mattsse:
## Summary

Private RPC log queries now derive redacted `logIndex` values from the complete caller-visible set of whitelisted logs in each transaction before applying the caller's requested address and topic filter.

Persisted filters retain the scoped user filter for post-index matching, and WebSocket subscriptions use the same ordering pipeline.

## Root cause

The previous implementation passed the already-filtered query result into the log-ordering redactor, so a log's ordinal depended on the filter shape. The same log could therefore receive different `(transactionHash, logIndex)` identities across receipts, queries, filters, and subscriptions.

## Impact

Log identity is stable across private RPC surfaces without exposing block-global ordering. Narrowing a query no longer renumbers matching logs, preventing deduplication and reconciliation errors in downstream consumers.

Closes ZONE-306

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -527,9 +527,16 @@ fn stale_filter_owner_ids(
         .collect()
 }
 
+#[derive(Clone)]
+struct OwnedFilter {
+    caller: Address,
+    /// The scoped user filter for log filters. Block filters do not need post-filtering.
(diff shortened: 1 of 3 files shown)


0xalpharush:
cyclops audit fast

PR #1147, opened 2026-08-12
feat: introduce checker & observe mode

grandizzy:


grandizzy:
cyclops audit

grandizzy:
cyclops audit

grandizzy:
cyclops audit

grandizzy:
cyclops audit

grandizzy:
cyclops audit

crates/checker/src/adapter/tempo/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+                let Some(L1ProtocolEvent::Portal(Portal::ZonePortalEvents::DepositMade(event))) =
+                    all_events.get(event_cursor).copied()
+                else {
+                    return Err(AdapterFindingCode::Grammar
+                        .failure("deposit is not followed by DepositMade"));
+                };
+                event_cursor += 1;
+                let deposit = ordinary_deposit_event(event, "deposit")?;
+                operations.push(ImportedOperation::AppendDeposit(deposit));
+                effects.push(Effect::DepositAppended {
```
0xalpharush: This seems very verbose

crates/checker/src/adapter/tempo/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+    zone_id: u32,
+) -> Result<ImportedAdaptation, Failure> {
+    let mut operations = Vec::new();
+    let mut effects = Vec::new();
+    for tx in observation.protocol_transactions() {
+        let all_events: Vec<_> = tx.outcomes().iter().map(|x| x.event()).collect();
+        let mut event_cursor = 0;
+        for direct_call in tx.direct_calls() {
+            if let Some(call) = direct_call.as_set_bounceback_gas() {
+                let Some(L1ProtocolEvent::Portal(Portal::ZonePortalEvents::BouncebackGasUpdated(
```
0xalpharush: why do we need this event?

crates/checker/src/adapter/tempo/withdrawals.rs
```diff
@@ -0,0 +1,344 @@
...
+    kernel::{Effect, PortalCallbackOperation, RefundClaim, TokenEnable, WithdrawalOutcome},
+    observe::events::{L1ProtocolEvent, Portal},
+};
+
+use super::super::{AdapterFindingCode, deposits::ordinary_deposit_event};
+
+/// Parsed effects and outcomes for one `processWithdrawals` call.
+pub(crate) struct WithdrawalAdaptation {
+    pub(crate) outcomes: Vec<WithdrawalOutcome>,
+    pub(crate) effects: Vec<Effect>,
```
0xalpharush: This seems like duplicating equivalent info?

crates/checker/src/adapter/deposits.rs
```diff
@@ -0,0 +1,55 @@
+//! Converts authenticated ordinary deposits into kernel facts.
+
+use alloy_primitives::FixedBytes;
+
+use crate::{failure::Failure, kernel::OrdinaryDeposit};
+
+use super::AdapterFindingCode;
+
+/// Convert an `advanceTempo` ordinary deposit into checker-owned facts.
+pub(super) fn ordinary_deposit(
```
0xalpharush: it's totally unnecessary to have a 2nd type and perform these conversions?

crates/checker/src/bootstrap/ancestry.rs
```diff
@@ -0,0 +1,94 @@
...
+                .into());
+            }
+            descending.reverse();
+            let mut path = Vec::with_capacity(descending.len() + 1);
+            path.push(ancestor.clone());
+            path.extend(descending);
+            return Ok(path);
+        }
+        let parent = acquire_l1_header(provider, parent_hash).await?;
+        if parent.number().checked_add(1) != Some(current.number()) {
```
0xalpharush, line 80: i'm not sure we need to duplicate all this that should be in the verifier

crates/checker/src/bootstrap/creation.rs
```diff
@@ -0,0 +1,192 @@
...
+            count: candidates.len(),
+        }
+        .into());
+    };
+    let header = acquire_l1_header(provider, *candidate).await?;
+    let observation = observe_l1(provider, &header, portal).await?;
+    let facts = adapt_imported(&observation, &header, header.hash(), zone_id)
+        .map_err(|failure| eyre::eyre!(failure.message))?
+        .facts;
+    let identity = validate_creation(&facts.operations, portal, zone_id)?;
```
0xalpharush: this seems redundant. we are searching for an event with a given address and ID and then checking it a 2nd time

decofe:
My concern is that the checker validates the implementation's exact sequence of calls, events, and intermediate updates, effectively creating a second implementation:

```rust
require(call == ProcessWithdrawal);
require(next_event == WithdrawalProcessed);
require(next_event == FallbackHeld);
require(next_event == RefundCreated);
```

It should parse enough authenticated data to reconstruct state, then check properties independently of that sequence. For example:

```rust
for fallback in held_fallbacks {
    let withdrawal = withdrawals[fallback.withdrawal_id];

    assert!(withdrawal.is_pending_user() || withdrawal.is_finalized_user());
    assert_eq!(withdrawal.fallback_id, fallback.id);
    assert_eq!(withdrawal.token, fallback.token);
    assert_eq!(withdrawal.amount, fallback.amount);
}
```

Parsing is unavoidable, but requiring every intermediate event to appear in a prescribed order duplicates the implementation and can reject a valid implementation change. The checker should focus on final cross-system properties such as matching ownership links and value conservation.

crates/checker/src/kernel/tests/invariants.rs
0xalpharush, line 1: move to benches/?

crates/checker/src/kernel/transition/tempo.rs
```diff
@@ -0,0 +1,621 @@
...
+                initial_token,
+            } => {
+                let portal = portal(&overlay)?;
+                let PortalState::AwaitingCreation(expected) = portal else {
+                    return Err(TransitionError::PortalAlreadyCreated);
+                };
+                if expected != *identity {
+                    return Err(TransitionError::PortalIdentityMismatch);
+                }
+                if portal_address(identity.zone_id) != identity.portal {
```
0xalpharush, line 26: i think we should be checking this where we receive the event

decofe:
After reading `transition/mod.rs` and `transition/tempo.rs`, I think the same operation is represented too many times. For example, `append_deposit` updates aggregate accounting, inserts a deposit row, advances a cursor, and emits an expected effect.

Could this instead use a typestate FSM for each bridge flow?

```rust
struct Flow<S> {
    id: FlowId,
    token: Address,
    amount: U256,
    state: S,
}

struct PendingDeposit;
struct PendingWithdrawal { fallback: FallbackId }
struct PendingRefund { recipient: Address, side: Side }
struct Settled;

impl Flow<PendingDeposit> {
    fn minted(self) -> Flow<Settled>;
    fn failed(self) -> Flow<PendingWithdrawal>;
}

impl Flow<PendingWithdrawal> {
    fn delivered(self) -> Flow<Settled>;
    fn bounced(self) -> Flow<PendingDeposit>;
    fn refund_pending(self, recipient: Address) -> Flow<PendingRefund>;
}

impl Flow<PendingRefund> {
    fn claimed(self) -> Flow<Settled>;
}
```

This makes invalid lifecycle transitions unrepresentable without needing the `StateKey`/`StateValue`, owner, fallback, and refund combinations to be checked repeatedly at runtime. A small tagged enum can wrap the typestates for persistence:

```rust
enum OpenFlow {
    Deposit(Flow<PendingDeposit>),
    Withdrawal(Flow<PendingWithdrawal>),
    Refund(Flow<PendingRefund>),
}
```

`apply_imported` would only apply the relevant typed transition and advance protocol heads. Deposit and withdrawal totals would be derived from the open flows and compared with observed collateral and supply, instead of maintaining parallel `TokenAccounting` counters and expected `Effect` values.

This seems substantially smaller while preserving the important checks: legal lifecycle transitions, ownership links, queue commitments, and collateral conservation.

crates/checker/src/observe/events/portal.rs
0xalpharush, line 1: can we derive this with alloy?

crates/checker/src/observe/abi/bounds.rs
```diff
@@ -0,0 +1,219 @@
+//! Allocation-bounded ABI surface navigation.
```
decofe: I think this module can be removed and the generated Alloy decoders used directly. Every current call site decodes a successful system or Portal transaction; on Tempo, the relevant transaction must also have emitted the expected protocol event, so malformed ABI lengths have already been rejected by onchain execution. Alloy made decoder allocations fallible in [alloy-rs/core#513](https://github.com/alloy-rs/core/pull/513), released in 0.6.3, and this PR uses 1.6.1. If we still want defense in depth for valid worst-case inputs, a small calldata/count cap before generated decoding is enough; reimplementing ABI offset walking here adds substantial complexity and parser-divergence risk.

crates/checker/src/adapter/tempo/mod.rs
grandizzy: protocol emits this from `setBouncebackGas`, check here makes sure no contract/event drift (call authenticates requested value and the event authenticates the resulting Portal transition)

grandizzy:
will continue simplify in https://github.com/tempoxyz/zones/pull/1216

crates/checker/src/adapter/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+}
+
+impl TryFrom<&Inbox::WithdrawalBounceBackDeposit> for BounceBackDeposit {
+    type Error = Failure;
+
+    fn try_from(deposit: &Inbox::WithdrawalBounceBackDeposit) -> Result<Self, Failure> {
+        let bytes = deposit.to.as_slice();
+        if bytes[..12].iter().any(|byte| *byte != 0) {
+            return Err(AdapterFindingCode::EventSequence
+                .failure("bounceback recipient has non-canonical high bytes"));
```
0xalpharush, line 228: remove

crates/checker/src/adapter/mod.rs
0xalpharush, line 1: i still don't understand point of fact and effects

PR #1154, opened 2026-08-12
fix(evm): require same-block withdrawal finalization

klkvr:
A Zone block can currently emit WithdrawalRequested without including finalizeWithdrawalBatch, even though the builder and settlement recovery assume all requests are finalized in the same block. This PR makes that builder behavior a consensus rule by rejecting such blocks in ZoneBlockExecutor::finish; because SPF replay uses the same executor, the rule is enforced in both node execution and proof replay.


0xrusowsky (approved):
nice!

mattsse (approved):

PR #1156, opened 2026-08-12
fix(genesis): remove account prefunding

mattsse:
## What changed

Remove the hard-coded native balance assignments for the zone admin and sequencer from generated genesis allocations. Refresh the bundled node genesis and checked-in root genesis fixture so both reflect the generator output.

## Why

Zones charge transaction fees in the configured TIP-20 fee token, so these operational accounts do not need placeholder native balances. Keeping the assignments unnecessarily mutates the generated genesis state and suggests that native funds are required.

## Impact

Newly generated zone genesis files no longer prefund the admin or sequencer with native balance. The bundled development genesis now has the same behavior. Existing deployed zones are unaffected.

### DIFF
--- a/crates/node/assets/zone-dev-genesis.json
+++ b/crates/node/assets/zone-dev-genesis.json
@@ -74,7 +74,7 @@
       "nonce": "0x1"
     },
     "0xaaaaaaaa00000000000000000000000000000000": {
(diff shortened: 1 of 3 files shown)


0xalpharush (approved):

0xalpharush:
cyclops audit super-fast

PR #1134, opened 2026-08-11
feat: pause portal deposits and withdrawals

decofe:
## Summary

- add a portal-wide emergency pause for new deposits, Zone withdrawal admission, and L1 withdrawal processing while proof-verified batch settlement continues
- allow the admin, active sequencers, and explicitly assigned `PauseGuardian` accounts to initiate a pause; only the admin can resume early
- make each pause expire automatically after 30 days and prevent extending an active pause
- add delayed, capability-scoped abdication for `PausePortal` and `AccessPolicy`
- keep the settlement monitor active during a pause while the withdrawal processor remains pause-aware
- make all dedicated role setters idempotent and prevent them from overwriting a different stored role
- use typed Rust storage accessors for `pauseExpiry` and `abdicationEffectiveAt`
- update tooling, ABI bindings, specs, tests, and the checked storage-layout artifact

## Stack

- based on [Zones #1157](https://github.com/tempoxyz/zones/pull/1157), which contains the sequencer-role refactor
- depends on [Tempo #7171](https://github.com/tempoxyz/tempo/pull/7171), pinned at `cce96c897ac3a1fcef07fb89e48aadb2ebeb19ba`

## Validation

- `forge fmt --check`
- `forge test` (716 passed)
- `cargo +nightly fmt --check`
- `cargo check -p tempo-zone-contracts -p tempo-xtask -p zone-node`

PR #1157, opened 2026-08-12
refactor: derive portal sequencer membership from roles

0xalpharush:
## Summary

- make stored Portal roles mutually exclusive enum values (`None`, `Sequencer`, `Account`, `CallbackGateway`)
- derive sequencer authorization from the shared role mapping instead of maintaining both `isSequencer` and role mappings
- store roles in slot 20 and leave slot 19 available for future use
- expose public `hasRole` and route sequencer checks through public `isSequencer`
- keep admin as a separate singleton authority; an admin may independently hold the `Sequencer` role
- keep account and callback-gateway membership behind their dedicated setters; changing between roles requires clearing the current role first
- add and remove sequencer membership only through `setSequencerSet`, validating the full proposed set before mutating roles
- align Solidity fixtures, Rust constants, native storage initialization, and installed-runtime regression coverage

## Rationale

I went ahead and made sequencer a role because we had two mappings performing authorization and they are now shared.

## Stack

This is the base PR for #1134 and depends on [Tempo #7175](https://github.com/tempoxyz/tempo/pull/7175) for native factory initialization and the embedded runtime.

## Validation

- `forge fmt --check`
- `forge test`
- `cargo check -p zone-precompiles --tests`
- `cargo test -p zone-precompiles outbox --lib`


0xalpharush:
cyclops audit fast

crates/contracts/src/precompiles/zone_portal.rs
```diff
@@ -22,6 +22,7 @@ crate::sol! {
         // -- Shared types --
         enum Role {
             None,
+            Sequencer,
```
0xalpharush, line 25: fine to break backwards compat as not yet deployed

0xalpharush:
cyclops audit super-fast

crates/precompiles/src/lib.rs
```diff
@@ -86,6 +86,11 @@ use tempo_precompiles::{
     tip20::{ITIP20::InsufficientBalance as TIP20InsufficientBalance, TIP20Token},
     tip403_registry::TIP403Registry,
 };
+use tempo_zone_contracts::ZonePortal as IZonePortal;
+
+pub(crate) fn has_portal_role(actual: u8, expected: IZonePortal::Role) -> bool {
+    actual == u8::from(expected)
+}
```
0xrusowsky: this is kinda useless. it would be better to pass `to: Address` rather than the actual role, and make it read the role from the l1 portal

0xrusowsky: addressed in [`9b46d67` (this PR)](https://github.com/tempoxyz/zones/pull/1157/commits/9b46d67b5498c3a130fa6daaeb6faa9404524247)

specs/spec.md
```diff
@@ -302,7 +303,7 @@ A single [`ZoneFactory`](#izonefactory) on Tempo creates zones and maintains the
...
 
-The factory's shared `ZoneMessenger` is fixed when each portal is initialized. It is separated from the portal so callback code does not execute with the fund-owning portal as `msg.sender`. Portal roles are managed atomically with `setRole(account, role)`. An account has exactly one of `None`, `Account`, or `CallbackGateway`; the messenger cannot have the `Account` role. `setAccessMode` and `setGatewayMode` activate or deactivate enforcement of the corresponding roles without clearing them.
+The factory's shared `ZoneMessenger` is fixed when each portal is initialized. It is separated from the portal so callback code does not execute with the fund-owning portal as `msg.sender`. An account has exactly one of `None`, `Sequencer`, `Account`, or `CallbackGateway`. The admin manages account and gateway membership through `setAllowedAccount` and `setGateway`, while sequencer membership changes only through `setSequencerSet`. Account and gateway setters atomically replace the prior non-sequencer role. `setAccessMode` and `setGatewayMode` activate or deactivate enforcement of the corresponding roles without clearing them.
```
0xrusowsky: This says the setters atomically replace a prior non-sequencer role, but the implementation requires two transactions (`Account -> None`, then `None -> CallbackGateway`, or vice versa). That introduces an intermediate authorization gap; the old `setRole` API did perform the replacement atomically. Should we retain an atomic transition API, or update this text to explicitly document the two-step behavior?

specs/spec.md
```diff
@@ -190,7 +190,7 @@ Each zone has two privileged roles registered on the [`ZonePortal`](#izoneportal
 - Any active sequencer may perform a sequencer-authorized portal operation. Batch settlement additionally requires a threshold certificate.
 - Hold the encryption private keys corresponding to the portal's encryption public keys and used to decrypt [deposits](#deposits).
 
-A zone MAY include its admin in the sequencer set. The protocol still treats each privileged call as belonging to its role.
+A zone MUST NOT include its admin in the sequencer set.
```
0xrusowsky: The overlap policy is contradictory: this PR intentionally tests and permits `admin` to be a sequencer, while this line says a zone MUST NOT include its admin in the sequencer set. The preceding paragraph and `docs/ZONES.md` also say overlap is supported. Can we choose one policy and align the contract, tests, PR summary, and docs? If overlap is forbidden, it needs enforcement in initialization, `setSequencerSet`, and admin transfer.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -168,8 +168,10 @@ contract ZonePortal is IZonePortal {
     uint8 public sequencerThreshold;
     uint256 public zoneHeight;
     address[] internal _sequencers;
-    mapping(address => bool) public isSequencer;
-    mapping(address => Role) public role;
+    /// @dev Reserved slot 19, available for future use.
+    uint256 private _reservedSlot19;
```
klkvr, line 172: why reserve the slot if we're breaking layout anyway?

0xrusowsky (commented):
other than the cmnts, directionally looks good

specs/ref-impls/src/tempo/ZonePortal.sol
0xalpharush: just minimizing diff

specs/spec.md
0xalpharush: Fixed in dbd11d9. The spec now documents that each setter manages only its own role, so changing Account ↔ CallbackGateway requires clearing the current role and then assigning the new one.

specs/spec.md
0xalpharush: Fixed in dbd11d9. We are keeping admin/sequencer overlap supported, matching the contract and regression test, and removed the contradictory MUST NOT statement from the spec.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -445,7 +460,7 @@ contract ZonePortal is IZonePortal {
     function acceptAdmin() external {
         if (pendingAdmin == address(0) || msg.sender != pendingAdmin) revert NotPendingAdmin();
         address previousAdmin = admin;
-        admin = pendingAdmin;
+        _setAdmin(pendingAdmin);
         pendingAdmin = address(0);
```
0xalpharush, line 467: Resolved in 27d06a9. Admin remains a separate singleton authority and may also hold the stored Sequencer role; the old overlap restriction was removed and the regression test covers that supported lifecycle.

specs/ref-impls/src/tempo/ZonePortal.sol
```diff
@@ -474,24 +489,39 @@ contract ZonePortal is IZonePortal {
 
     /// @notice Add or remove an account from closed-loop portal flows.
     function setAllowedAccount(address account, bool allowed) external onlyAdmin {
-        _setRole(account, allowed ? Role.Account : Role.None);
+        Role previous = role[account];
+        require(previous != Role.Sequencer);
+        Role next = allowed ? Role.Account : Role.None;
```
0xalpharush, line 498: Fixed in 1c03dd0. setAllowedAccount rejects the messenger when adding Account, and initialize applies the same invariant to every initial allowed account.

specs/ref-impls/src/interfaces/IZone.sol
```diff
@@ -11,6 +11,7 @@ address constant ZONE_MESSENGER_ADDRESS = 0x5A4d00000000000000000000000000000000
 /// @notice Mutually exclusive authorization role assigned to a Tempo account.
 enum Role {
     None,
+    Sequencer,
```
0xalpharush, line 14: Not applicable to this rollout: the Portal has not been deployed, so there is no persisted pre-change role state or historical anchor decoding to preserve. The enum/storage transition lands atomically across the stacked Zones and Tempo changes.

0xalpharush:
cyclops audit super-fast

specs/ref-impls/src/tempo/ZonePortal.sol
legion2002, line 172: why are we reserving this slot? we can just delete this as zones is not live yet.

legion2002 (dismissed):

specs/ref-impls/src/interfaces/IZone.sol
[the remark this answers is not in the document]
0xalpharush: No migration is needed because this Portal layout has not been deployed. The stacked Zones and Tempo changes install the new enum values and native storage initialization atomically; there is no persisted pre-change role state to reinterpret. Covered by fresh factory/runtime initialization tests.

specs/ref-impls/src/tempo/ZonePortal.sol
0xalpharush: made diff smaller

klkvr (approved):

klkvr (dismissed):

legion2002:
+1

legion2002 (approved):

PR #1159, opened 2026-08-13
refactor(node): validate zones with Tempo consensus

klkvr:
## Summary
- reuse generic `TempoConsensus<ZoneChainSpec>` instead of noop consensus
- configure Zone consensus gas sections and domain-separated system transaction chain IDs
- update the Tempo dependency to the generic consensus implementation

## Validation
- `cargo +nightly clippy -p zone-node --all-targets -- -D warnings`

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -5,6 +5,7 @@ use std::{net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 use alloy_primitives::Address;
 use alloy_signer_local::PrivateKeySigner;
 use clap::{Args, CommandFactory, FromArgMatches};
+use reth_chainspec::EthChainSpec;
 use reth_ethereum::cli::Cli;
 use reth_tracing::tracing::info;
 use tempo_evm::consensus::TempoConsensus;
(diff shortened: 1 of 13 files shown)


klkvr:
cyclops audit

crates/node/src/node.rs
```diff
@@ -1621,6 +1623,53 @@ impl ZoneExecutorBuilder {
...
+        let l1_chain_id = match self.l1_state_provider_config.chain_id {
+            Some(chain_id) => chain_id,
+            None => {
+                let l1_provider = L1StateProvider::new(
+                    self.l1_state_provider_config,
+                    L1StateCache::new(),
+                    tokio::runtime::Handle::current(),
+                )
+                .await?;
+                l1_provider.chain_id().await?
```
0xrusowsky: nit: rather than initializing an `L1StateProvider` here and in the zone executor builder, we could share a lazily initialized provider between both builders

crates/node/src/cli.rs
```diff
@@ -96,8 +96,8 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
 
     let components = |spec: Arc<ZoneChainSpec>| {
         (
-            ZoneEvmConfig::new_without_l1(spec),
-            NoopConsensus::default(),
+            ZoneEvmConfig::new_without_l1(spec.clone()),
+            TempoConsensus::new(spec),
```
0xrusowsky, line 111: should this use `ZoneConsensusBuilder` to be L1 schedule-aware?

mattsse (dismissed):
LGTM

0xrusowsky (dismissed):

0xrusowsky (approved):

PR #1162, opened 2026-08-13
refactor(node): simplify peer block import

klkvr:
## Summary

Simplifies follower peer-block import by removing checks already enforced by the L1 subscriber or `advanceTempo` execution:

- the duplicate local-checkpoint transition check
- peer-driven insertion into the deposit queue
- retry-loop and helper indirection around anchor waiting

The importer still waits for the exact locally observed L1 anchor, compares `advanceTempo` deposits and enabled tokens with receipt-authenticated portal events, authenticates live senders after the anchor updates the schedule, and executes the block through the engine.

## Validation

- `cargo +nightly clippy -p zone-l1 -p zone-node --all-targets --all-features -- -D warnings`

### DIFF
--- a/crates/l1/src/event.rs
+++ b/crates/l1/src/event.rs
@@ -98,6 +98,43 @@ impl L1PortalEvents {
         }
     }
 
+    /// Validate that `advanceTempo` processes every deposit and token enable observed in this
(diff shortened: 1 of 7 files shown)


klkvr:
cyclops audit

PR #1163, opened 2026-08-13
fix: use current gas limits for withdrawal helpers

decofe:
## Summary

- raise the ZoneOutbox approval gas limit to 500,000
- raise the withdrawal request gas limit to 10,000,000, matching the current E2E test allowance
- stop immediately when the L2 withdrawal receipt failed instead of polling L1 forever

## Validation

- `git diff --check`
- extracted Justfile recipe bodies pass `bash -n`
- `just --list` was unavailable because `just` is not installed in the sandbox

Prompted by: @grandizzy


jenpaff (approved):

jenpaff:
cyclops audit fast

grandizzy (approved):

PR #1172, opened 2026-08-13
feat(prover): embed devnet genesis in CI image

rkrasiuk:
## Summary

Download the Tempo devnet genesis during CI image builds from a configurable URL, defaulting to `https://devnet-assets.tempoxyz.dev/tempo-devnet-nextfork.json`. The genesis is embedded in the prover image through the existing tempo-genesis build context.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,7 +14,17 @@ on:
         description: "Tag nightly"
         type: boolean
         default: false
+      tempo_genesis_url:
+        description: "Tempo genesis JSON URL to embed in the prover image"
+        type: string
+        default: "https://devnet-assets.tempoxyz.dev/tempo-devnet-nextfork.json"
   workflow_call:
+    inputs:
+      tempo_genesis_url:
+        description: "Tempo genesis JSON URL to embed in the prover image"
+        required: false
+        type: string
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -112,6 +123,22 @@ jobs:
...
+          mkdir -p .tempo-genesis
+          curl --proto '=https' --tlsv1.2 \
+            --fail-with-body --location \
+            --output .tempo-genesis/genesis.json \
+            "$TEMPO_DEVNET_GENESIS_URL"
+          jq -e '.config.chainId | type == "number"' \
+            .tempo-genesis/genesis.json >/dev/null
+
+          genesis_sha=$(sha256sum .tempo-genesis/genesis.json | cut -d' ' -f1)
+          echo "Embedding Tempo devnet genesis sha256:${genesis_sha}"
```
shekhirin, line 140: nit: this is not an actual genesis hash, I think it may be confusing to have it here

shekhirin (approved):

.github/workflows/docker.yml
rkrasiuk: it is prefixed with sha256 tho

PR #1178, opened 2026-08-13
fix(prover): pass Tempo genesis to enclave explicitly

rkrasiuk:
## Summary

Pass `--tempo-genesis /etc/tempo/genesis` explicitly in the prover enclave entrypoint.


rkrasiuk:
+1

shekhirin (approved):

PR #1179, opened 2026-08-13
fix(precompiles): restrict scoped reads to account owners

decofe:
## Summary
- remove the active-sequencer exception from account-scoped precompile reads
- restrict TIP-20, nonce, storage-credit, and account-keychain getters to owners/spenders named by the call
- remove the now-unused admission error path and add sequencer-denial regression coverage

## Validation
- `cargo test -p zone-precompiles`
- `cargo +nightly clippy -p zone-precompiles --all-targets -- -D warnings`

Prompted by: @0xalpharush


klkvr (dismissed):

0xalpharush:
cyclops audit super-fast

crates/precompiles/src/privacy.rs
```diff
@@ -5,24 +5,19 @@ use alloy_sol_types::SolError;
...
-    execution::{CallCheck, CallRuleError},
+    execution::CallCheck,
     storage::{L1State, L1StorageReader},
 };
 
-/// Allow a read when the immediate EVM caller owns any account named by the getter, or is an
-/// active sequencer at the transaction's anchored Tempo block.
-pub(crate) fn check_caller_or_sequencer<P: L1StorageReader>(
-    l1: &L1State<P>,
+/// Allow a read only when the immediate EVM caller owns an account named by the getter.
```
mattsse, line 9: **[P2] Align every normative mention with owner-only scoped reads**

This correctly removes the sequencer bypass, but these normative references still grant it:

- `specs/spec.md:841`: make the selected account the default/only reader; retain additional readers only where explicitly defined. In the same paragraph, specify `balanceOf(account)` as owner-only and `allowance(owner, spender)` as owner-or-spender. Keep `ZoneInbox.refunds` as the explicit owner-or-sequencer exception because this PR does not change that getter.
- `specs/spec.md:1486`: change “restricted to the account owner (or sequencer)” to “`balanceOf` is restricted to the account owner; `allowance` is restricted to its owner or spender.”
- `invariants.md:126`: remove “or the sequencer” and state the owner/spender rule explicitly.
- `invariants.md:127`: remove “or is an active sequencer.”

I also checked the other related references. Keep `specs/spec.md:762` and `:2025` plus `invariants.md:128` because `ZoneInbox.refunds` still admits sequencers. Keep `specs/spec.md:134`, `:967`, `:979`, and `:1055` plus `invariants.md:140–141` because they describe sequencer/operator visibility through the unchanged unrestricted RPC on port 8545, not an elevated authorization-token or account-getter bypass. GitHub cannot attach an applyable suggestion block to `specs/spec.md` until that unchanged file is added to this PR, so the exact required edits are listed here.

crates/node/tests/it/redacted_rpc_e2e.rs
```diff
@@ -853,15 +851,9 @@ async fn test_tip20_eth_call_privacy() -> eyre::Result<()> {
...
-    let sequencer_balance_bytes = hex::decode(
-        sequencer_balance["result"]
-            .as_str()
-            .expect("sequencer balanceOf should return hex")
-            .trim_start_matches("0x"),
-    )?;
-    assert_eq!(
-        PrecompileTip20::balanceOfCall::abi_decode_returns(&sequencer_balance_bytes)?,
-        expected_owner_balance
+    assert!(
```
mattsse, line 854: **[P3] Update the stale test documentation above this case**

Lines 766–767 still say that the configured sequencer retains access, contradicting these new assertions. Suggested wording: “`eth_call` against the zone TIP-20 enforces read privacy for `balanceOf` and `allowance` without a sequencer bypass.”

mattsse (commented):
The owner-only direction looks right. The remaining issue is documentation alignment: the normative privacy section, TIP-20 precompile section, security invariants, and one test comment still describe a sequencer bypass that this PR intentionally removes. I audited the other sequencer-read references; the refund getter and unrestricted operator-RPC statements should remain because those behaviors are unchanged.

mattsse (approved):
LGTM

PR #1180, opened 2026-08-13
feat(xtask): add sequencer key rotation admin commands

adityapk00:
Stacked on #1153.
Adds guarded admin commands and supporting configuration, snapshot, invariant, and secret-file handling for sequencer key rotation.

PR #1181, opened 2026-08-13
feat(xtask): add individual sequencer identity rotation

adityapk00:
## Summary

- add `admin identity prepare` for independent per-node P2P and secp256k1 keys
- add guarded `admin sequencer-set replace` dry-run and execute flows
- support safe mixed-manifest leadership handoffs during one-at-a-time rotation
- require both the target and relay to have loaded the finalized next manifest  
- 
## Stack

Depends on #1180. This PR is based on `xtask_seqkey_rotate` so it contains only the individual-identity rotation commits in its diff.


adityapk00:
cyclops audit super-fast

legion2002 (approved):

legion2002 (approved):

PR #1195, opened 2026-08-14
refactor(node): route batch submission through persistent actor

adityapk00:
## Summary

Cuts the node over to the persistent actor and removes the old inline monitor.

Stacked on #2 of this series.


adityapk00:
cyclops audit fast

crates/sequencer/src/lib.rs
```diff
@@ -222,6 +283,59 @@ pub async fn spawn_zone_sequencer<P: ZoneSequencerProvider>(
...
+) -> ZoneSequencerHandle {
+    let mut runtime = spawn_persistent_batch_submission(&config, signer, zone_provider.clone())
+        .await
+        .expect("valid persistent batch submission configuration");
+    runtime
+        .resources
+        .batch_submission
+        .set_role(BatchSubmissionRole::leader(1))
+        .await
+        .expect("persistent batch submission actor starts");
```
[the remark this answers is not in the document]
adityapk00: Fixed in #1194

crates/sequencer/src/batch_submission.rs
```diff
@@ -506,24 +684,37 @@ impl<B: BatchSubmissionBackend> BatchSubmissionActor<B> {
         shutdown: &CancellationToken,
     ) -> Result<ConfirmedBatchSubmission> {
         tokio::pin!(operation);
+        let crossed_send_boundary = Arc::new(AtomicBool::new(false));
+        let boundary = crossed_send_boundary.clone();
+        let tracked = std::future::poll_fn(move |context| {
+            boundary.store(true, Ordering::Release);
```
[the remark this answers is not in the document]
adityapk00: Fixed in #1194

crates/sequencer/src/candidate_monitor.rs
```diff
@@ -0,0 +1,180 @@
...
+                if current.zone_height >= candidate.to {
+                    self.latest_observed = candidate.to;
+                    info!(
+                        generation = self.generation,
+                        zone_height = candidate.to,
+                        "Observed confirmed batch submission progress"
+                    );
+                    break;
+                }
+                if current != anchor {
```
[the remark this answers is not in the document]
adityapk00: Fixed in #1194

crates/sequencer/src/batch_submission.rs
```diff
@@ -380,7 +539,12 @@ impl<B: BatchSubmissionBackend> BatchSubmissionActor<B> {
                 )
                 .await?
             {
-                Some(prepared) => prepared?,
+                Some(Ok(prepared)) => prepared,
+                Some(Err(error)) if self.attestation_store.is_some() => {
+                    warn!(%error, zone_height = candidate.to, "Rejected invalid settlement certificate");
+                    continue;
+                }
+                Some(Err(error)) => return Err(error),
```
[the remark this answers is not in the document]
adityapk00: Fixed in #1194 + the attestation prepare here.

crates/sequencer/src/lib.rs
```diff
@@ -222,6 +283,59 @@ pub async fn spawn_zone_sequencer<P: ZoneSequencerProvider>(
...
+    let handle = spawn_zone_sequencer_generation(
+        config,
+        zone_provider,
+        prover_config,
+        runtime.resources.clone(),
+        1,
+        shutdown.clone(),
+    )
+    .await;
+    let mut actor = runtime.actor_handle;
```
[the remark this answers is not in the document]
adityapk00: removed the legacy static single sequenccer, not needed anymore

crates/sequencer/src/batch_submission.rs
```diff
@@ -415,12 +583,22 @@ impl<B: BatchSubmissionBackend> BatchSubmissionActor<B> {
...
-                        return Err(error);
+                    if !self.role_matches(&candidate)
+                        || self
+                            .anchor
+                            .is_some_and(|anchor| anchor.zone_height >= candidate.to)
+                        || !self.candidate_extends_anchor(&candidate)
+                    {
+                        return Ok(());
+                    }
+                    attempt = 1;
```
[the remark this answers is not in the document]
adityapk00: Fixed, added a exponential backoff to 60s, so it doesn't drain L1 funds every 200ms

crates/sequencer/src/candidate_monitor.rs
```diff
@@ -0,0 +1,180 @@
...
+                _ = fallback.tick() => {}
+                notification = canonical.next() => {
+                    let Some(notification) = notification else {
+                        eyre::bail!("canonical zone state notification stream closed");
+                    };
+                    if notification.reverted().is_some() {
+                        eyre::bail!("canonical zone chain reorged while the sequencer was active");
+                    }
+                }
+            }
```
klkvr: why not have this parse the notifications explicitly? notification includes all of the receipts and we can act on it directly i believe?

crates/sequencer/src/candidate_monitor.rs
```diff
@@ -0,0 +1,180 @@
...
+struct CandidateMonitor<P: ZoneSequencerProvider> {
+    metrics: crate::metrics::ZoneMonitorMetrics,
+    config: ZoneMonitorConfig,
+    provider: P,
+    generation: u64,
+    candidate_tx: mpsc::Sender<BatchCandidate>,
+    progress_rx: watch::Receiver<Option<SubmissionAnchor>>,
+    latest_observed: u64,
+    shadow_prover: Option<ShadowProver>,
+}
```
klkvr: why does it have to be a separate actor? could all this logic be part of the batch submission actor instead?

crates/sequencer/src/candidate_monitor.rs
adityapk00: good call, updated to parse txns, receipts directly.

crates/sequencer/src/candidate_monitor.rs
adityapk00: folded it in. the main nuance is error recovery: source errors now reset its catchup state, and tried by the actor (with backoff)

crates/sequencer/src/candidate_monitor.rs
```diff
@@ -0,0 +1,289 @@
...
+        self.record_metrics(number, anchor.zone_height);
+        let candidate = self.candidate(generation, anchor, boundary, finalized, end);
+        self.candidate_pending = true;
+        self.enqueue_prover(&candidate);
+        Ok(Some(candidate))
+    }
+
+    async fn next(
+        &mut self,
+        generation: u64,
```
klkvr: seems like this u64 is passed here just to be returned back as `BatchCandidate::generation`? seems kind of redundant

crates/sequencer/src/candidate_monitor.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+impl<P: ZoneSequencerProvider> BatchCandidateSource for CandidateMonitor<P> {
+    fn next_candidate(
+        &mut self,
+        generation: u64,
+        anchor: SubmissionAnchor,
+    ) -> BoxFuture<'_, eyre::Result<BatchCandidate>> {
+        Box::pin(async move {
+            self.reconcile_anchor(anchor);
```
klkvr: this + logic at the top of `Self::next` seems quite confusing and looks like a consequence of `CandidateMonitor` being persistent while it could just be a temporary stream of batches?

i.e  once `BatchSubmissionActor` knows we are leading. it could create a provider-powered stream of batches from the anchor it started leading at

does this sound reasonable? i think this is kind of what i meant by merging candidate monitor into batch submission actor

crates/sequencer/src/lib.rs
```diff
@@ -185,40 +240,24 @@ pub async fn spawn_zone_sequencer<P: ZoneSequencerProvider>(
...
-        shutdown,
-    );
-
+    let batch_submission = resources.batch_submission;
+    let batch_submission_handle = tokio::spawn(async move {
+        tokio::select! {
+            () = shutdown.cancelled() => {}
+            () = batch_submission.closed() => {}
+        }
+    });
```
klkvr: this seems a bit odd? why not poll batch submission actor death on higer level vs spawning a watcher per generation

crates/sequencer/src/batch_submission.rs
```diff
@@ -94,16 +116,155 @@ pub trait BatchSubmissionBackend: Send + Sync + 'static {
...
+    fn resync(&self) -> BoxFuture<'_, Result<BatchResync>> {
+        Box::pin(async move {
+            let portal_anchor = resolve_portal_zone_anchor(
+                &self.provider,
+                self.portal_address,
+                self.submitter.l1_provider(),
+            )
+            .await
+            .wrap_err("failed to resolve portal-confirmed zone block during resync")?;
+            let previous_snapshot = self
```
klkvr, line 199: tbh i'm not really a fan of those abstractions because of how they make it trickier to navigate the code. i understand we need them for tests but i there must be ways to test things by ingecting custom behavior at lower level? e.g changing rpc responses

crates/sequencer/src/batch_submission.rs
klkvr: what would be the worst case if we just race entire `process_candidate` with `shutdown ` via a select?

it seems like the shutdown propagation pattern we've had before had just migrated into the persistent actor now?

klkvr: also it looks like this shutdown is crreated here https://github.com/tempoxyz/zones/blob/e1a5bf9b52e675bd0e7b7241bb04a0cdf51adcef/crates/sequencer/src/lib.rs#L203-L204

and is never actually fired?

PR #1198, opened 2026-08-14
chore: remove obsolete Solidity reference implementations

decofe:
## Summary

- remove the obsolete Solidity reference implementations while preserving the deployed `ZonePortal`, `ZoneMessenger`, and `Verifier` runtimes
- retain the interfaces, libraries, tests, storage-layout guard, and runtime-sync workflow required by those deployed contracts
- retain Rust-owned JSON fixtures under `crates/l1/testdata`
- remove tests and sources for the deleted implementations

## Validation

- `git diff --check`
- verified Tempo TIP-1091 and the runtime-sync workflow consume `ZonePortal`, `ZoneMessenger`, and `Verifier`
- local Forge cannot compile the T8 configuration because the installed Forge lacks Tempo hardfork support; CI uses the Tempo-capable toolchain

Prompted by: @shekhirin


legion2002 (commented):
SwapAndDepositRouter also we want to keep

crates/contracts/README.md
```diff
@@ -1,134 +1,12 @@
-# Tempo Specs
+# Zone runtime contracts
```
jenpaff, line 1: @0xKitsune @legion2002 @0xrusowsky now that we remove the spec reference impls, can we just move the remaining contracts to `/crates/contracts/src/runtime` ? Or any concerns with that?

legion2002: yes we should move, ref-impl is misleading now as the remaining contracts are deployed in prod

legion2002:
cyclops audit super-fast

.github/workflows/specs.yml
```diff
@@ -51,18 +51,18 @@ jobs:
         id: solidity-key
         run: |
           echo "forge-sha=$(forge --version | sed -n 's/^Commit SHA: //p')" >> "$GITHUB_OUTPUT"
-          echo "specs-tree=$(git rev-parse HEAD:specs/ref-impls)" >> "$GITHUB_OUTPUT"
+          echo "contracts-tree=$(git rev-parse HEAD:crates/contracts)" >> "$GITHUB_OUTPUT"
 
       - name: Restore Solidity artifacts
```
jenpaff: i think this can be removed right?

.github/workflows/contracts.yml
```diff
@@ -4,18 +4,18 @@ on:
   push:
```
jenpaff, line 4: can the specs workflow be removed? @0xrusowsky @legion2002

decofe: I think we should keep it: although the reference implementations are gone, this workflow still runs Forge build/fmt plus Solidity tests, symbolic checks, and coverage for the retained deployed ZonePortal, ZoneMessenger, Verifier, and SwapAndDepositRouter runtimes.

xtask/src/generate_zone_genesis.rs
```diff
@@ -82,7 +82,7 @@ pub(crate) struct GenerateZoneGenesis {
     #[arg(long)]
     pub(crate) sequencer: Option<Address>,
 
-    #[arg(long, default_value = "specs/ref-impls/out")]
+    #[arg(long, default_value = "crates/contracts/out")]
```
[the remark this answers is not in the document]
decofe: Fixed in 6ef63f77. Zone genesis now initializes the native ZoneInbox directly, commits the native initializer state before reading the cache, and no longer accepts or reads a Foundry artifact path. I also verified generate-zone-genesis succeeds from a clean output path and emits code at 0x1c00000000000000000000000000000000000001.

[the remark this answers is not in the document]
jenpaff: @0xrusowsky @legion2002 worth reviewing

.github/workflows/contracts.yml
jenpaff: should we at least rename it ? specs doesn't make sense anymore

decofe: Agreed. Renamed the workflow file from specs.yml to contracts.yml, changed the workflow display name to Contracts, and renamed the aggregate check to contracts success in 18f23ad3.

jenpaff (approved):

jenpaff (dismissed):

.github/workflows/contracts.yml
legion2002: yes we should keep it, becasue it updates ZonePortal, renaming makes sense

xtask/src/generate_zone_genesis.rs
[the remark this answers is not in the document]
legion2002: addressed in https://github.com/tempoxyz/zones/commit/5417e4d4aca7e80e5ab0f39ad0249f3c48dbd382

.github/workflows/contracts.yml
legion2002: addressed in https://github.com/tempoxyz/zones/commit/18f23ad3a04db6adf34fa9474b15be9706ccc202

legion2002 (approved):

0xrusowsky (approved):

legion2002 (approved):

0xrusowsky (approved):

sds (approved):

PR #1201, opened 2026-08-17
feat(prover): define Nitro batch attestation format

rkrasiuk:
## Summary

Define `ProofBundle` for verifier configuration and raw Nitro attestation documents. Add the canonical Nitro batch-attestation digest derived from SPF public inputs and batch output.


decofe:
cyclops audit fast

crates/prover/src/protocol.rs
```diff
@@ -1,9 +1,28 @@
...
+use zone_spf::{BatchOutput, BatchWitness, PublicInputs};
 
 /// Current version of the prover request and response wire format.
 pub const PROTOCOL_VERSION: u16 = 1;
 
+/// Canonical verifier configuration for the first Nitro-backed verifier policy.
+pub const NITRO_VERIFIER_CONFIG_V1: &[u8] = &[1];
+
+/// Type hash for the data placed in the Nitro attestation document's `user_data` field.
+pub const NITRO_BATCH_ATTESTATION_TYPE: &str = "NitroBatchAttestation(uint256 parentChainId,address verifier,address portal,uint32 zoneId,uint64 tempoBlockNumber,uint64 anchorBlockNumber,bytes32 anchorBlockHash,uint64 expectedWithdrawalBatchIndex,bytes32 prevBlockHash,bytes32 nextBlockHash,bytes32 prevProcessedHash,bytes32 nextProcessedHash,uint64 prevDepositNumber,uint64 nextDepositNumber,bytes32 withdrawalQueueHash,bytes32 verifierConfigHash)";
```
shekhirin: this should include `TokenEnablementTransition` fields https://github.com/tempoxyz/zones/blob/216eab4c4eb5949d2235ba2c2699fd34051b46e2/crates/contracts/src/runtime/interfaces/IZone.sol#L89-L93

crates/prover/src/protocol.rs
```diff
@@ -1,9 +1,28 @@
...
+
```
klkvr: i think we can use sol! macro for this?

PR #1205, opened 2026-08-17
feat(prover): support checkpoint-only zone blocks

shekhirin:
Ref: https://github.com/tempoxyz/tempo/pull/7164

Extends offline witness generation to extract and authenticate checkpoint-only Tempo header sequences, including the final imported header and bootstrap storage reads.


shekhirin:
cyclops audit

PR #1211, opened 2026-08-18
Backrun withdrawals with ordered 2D nonces

0xKitsune:
This PR uses an isolated 2D nonce lane to backrun `submitBatch` with ordered withdrawal processing in the same Tempo block. It falls back to the existing recovery path when the fast path is unavailable or uncertain.

### DIFF
--- a/crates/sequencer/src/settlement.rs
+++ b/crates/sequencer/src/settlement.rs
@@ -23,13 +23,21 @@
 //! configured direct window by falling back to ancestry mode — a recent anchor
 //! block plus a locally validated parent-hash header chain.
 
-use std::{collections::BTreeMap, fmt, sync::OnceLock, time::Duration};
+use std::{
+    collections::BTreeMap,
+    fmt,
+    sync::{
+        OnceLock,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::Duration,
+};
 
 use crate::{
     ZoneSequencerProvider,
     abi::{self, BlockTransition, DepositQueueTransition, IZoneInbox, IZoneOutbox, ZonePortal},
     attestation::{AttestationStore, SettlementAttestation, SettlementCertificate},
-    withdrawals::single_batch_gas_limit,
+    withdrawals::{WithdrawalBatch, build_withdrawal_batches},
 };
 use alloy_consensus::{Transaction, TxReceipt as _, transaction::TxHashRef as _};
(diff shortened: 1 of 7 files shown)


0xKitsune:
cyclops audit fast

crates/sequencer/src/settlement.rs
```diff
@@ -478,11 +506,85 @@ impl BatchSubmitter {
...
+            let pending_submission = submission
+                .send()
+                .await
+                .map_err(|error| BatchSubmitError::Other(error.into()))?;
+            let submit_tx_hash = *pending_submission.tx_hash();
+            let backrun = self
+                .portal
+                .processWithdrawals(withdrawals.to_vec(), B256::ZERO)
+                .from(submission_address)
+                .nonce_key(SUBMIT_BATCH_NONCE_KEY)
```
0xKitsune: Fixed in 822859899. The paired submitBatch and all gas-bounded withdrawal chunks now use dedicated key 4, while mandatory settlement remains on key 1 and recovery remains on key 2. We also skip the fast path while the portal is paused and permanently disable key 4 after any uncertain or reverted transaction, so a stale optional nonce cannot block later settlement.

0xKitsune:
cyclops audit fast

crates/sequencer/src/settlement.rs
```diff
@@ -478,14 +523,127 @@ impl BatchSubmitter {
...
+                                WITHDRAWAL_BACKRUN_CONFIRM_TIMEOUT,
+                                backrun.send_sync(),
+                            )
+                            .await;
+                            (batch, backrun_nonce, result)
+                        }
+                    });
+            let (submission_result, backrun_results) = tokio::join!(
+                tokio::time::timeout(Duration::from_secs(30), pending_submission.get_receipt()),
+                future::join_all(backruns),
```
0xKitsune, line 600: Fixed in 42502cabc. The monitor now passes the full WithdrawalBatchLimits into settlement. The fast path admits only a prefix bounded by max_in_flight_batches, 20M aggregate planned withdrawal gas (22M including submitBatch, leaving 8M of Tempo general-gas headroom), and the portal reserve of 20 worst-case bouncebacks. After a partial prefix confirms, the full payload is persisted and lane 2 reconciles the processed prefix before draining the suffix.

0xKitsune:
cyclops audit fast

PR #1212, opened 2026-08-18
feat(prover): produce Nitro NSM attestations

rkrasiuk:
## Summary

Validate the witness parent chain, execute SPF, and bind the canonical batch digest to Nitro attestation user data. Return the raw COSE/CBOR document as the proof bundle and report NSM failures explicitly.


decofe:
cyclops audit fast

shekhirin (dismissed):

PR #1214, opened 2026-08-18
fix(precompiles): preserve receive policy resolution errors

decofe:
## Summary

- propagate `TempoPrecompileError` from receive-policy recipient resolution
- reserve `Unauthorized` for successful non-matching resolutions
- cover out-of-gas propagation with a regression test

## Testing

- `cargo test -p zone-precompiles receive_policy_guard --no-fail-fast`
- `cargo +nightly clippy -p zone-precompiles --tests -- -D warnings`

[ZONE-701](https://linear.app/tempoxyz/issue/ZONE-701/receive-policy-guard-discards-state-read-errors-from-resolve-recipient)

Prompted by: @0xrusowsky


decofe:
Closing as a duplicate of #1204; the requested `Error(TempoPrecompileError)` simplification is now applied there.

PR #1216, opened 2026-08-18
feat: introduce checker & observe mode

grandizzy:
## Summary

Adds an observe-only Reth ExEx that independently verifies Zone bridge solvency without affecting execution or consensus.

## What changed

- Derives per-account TIP-20 balances and bridge liabilities from authenticated Tempo and Zone events.
- Verifies exact Zone balances and supply, Portal collateral, deposit mints, and withdrawal debits/burns after every block.
- Handles deposits, withdrawals, bounce-backs, refunds, sponsored fees, and reverted transaction fees.
- Automatically discovers and authenticates the Zone/Portal identity from genesis and Tempo history.
- Persists the current accounting state and active finding in bounded MDBX storage, resuming from the last verified block after restart.
- Retries transient acquisition failures and disables only the checker on terminal errors while continuing to drain ExEx notifications.
- Exposes verification, lag, retry, disabled, and divergence metrics plus structured activity logs.
- Reuses the node’s authenticated L1 observations when enabled, while preserving the existing node path when checker mode is off.
- Adds checker design documentation and unit/integration coverage.

Enable with:

```bash
tempo-zone node ... --checker.mode observe
```

The checker database is created automatically under `<node datadir>/checker`.


crates/checker/src/accounting/effects.rs
```diff
@@ -0,0 +1,388 @@
...
+        match *event {
+            L1PortalEvent::TokenEnabled { token } => {
+                effects.push(Effect::EnableToken { token });
+            }
+            L1PortalEvent::DepositMade {
+                token, net_amount, ..
+            } => effects.push(Effect::PendingDeposit {
+                token,
+                amount: U256::from(net_amount),
+                increase: true,
```
0xalpharush: it feels strange to have bool. should be credit/debit enum and not a second type/arg but captured by effect?

crates/checker/src/l1/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+/// Fetch every canonical Tempo block after `parent` through `tip`.
+pub(crate) async fn collect_l1_history(
+    provider: &DynProvider<TempoNetwork>,
+    portal: Address,
+    parent: BlockNumHash,
+    tip: BlockNumHash,
+) -> eyre::Result<Vec<L1BlockEvidence>> {
+    eyre::ensure!(tip.number >= parent.number, "Tempo anchor moved backwards");
+    let mut previous = parent;
+    let mut history = Vec::with_capacity((tip.number - parent.number) as usize);
```
0xrusowsky: trying to be wary of OOM issues, just flagging that if this were to run in sequencer nodes and somehow an attacker manages to provide a faulty number, it could kill the nodes.

we could pre-check capacity and error if the diff is too big

crates/checker/src/accounting/effects.rs
```diff
@@ -0,0 +1,388 @@
...
+    }
+    effects
+}
+
+/// Derive account and liability changes from authenticated Zone events.
+pub(crate) fn from_zone<'a>(
+    events: impl IntoIterator<Item = &'a L2BridgeEvent>,
+) -> Result<Vec<Effect>, EffectError> {
+    let events = events.into_iter().collect::<Vec<_>>();
+    authenticate_mints(&events)?;
```
0xalpharush: can we do this earlier and not pass events from other sources? if we just check when we create the type and make two args ZoneEvent, PortalEvent, it is not possible to be from wrong source

crates/checker/src/accounting/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// Token was authenticated through Portal creation or enablement.
+    pub(crate) enabled: bool,
+    /// Sum of all nonzero account entitlements.
+    pub(crate) account_total: U256,
+    /// Liability from deposits authenticated on Tempo but not yet reflected on the Zone.
+    pub(crate) pending_deposits: U256,
+    /// Liability from withdrawals accepted on the Zone but not yet settled on Tempo.
+    pub(crate) pending_withdrawals: U256,
+    /// Liability from refunds parked on Tempo but not yet claimed.
+    pub(crate) pending_refunds: U256,
```
0xalpharush: there are also refund on zone right?

crates/checker/src/l1/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+            .get_block_by_number(number.into())
+            .hashes()
+            .await?
+            .ok_or_else(|| eyre::eyre!("Tempo block {number} is unavailable"))?;
+        eyre::ensure!(
+            block.header().parent_hash() == previous.hash,
+            "Tempo history is not contiguous at block {number}"
+        );
+        let coordinate = BlockNumHash::new(number, block.header().hash());
+        history.push(collect_l1_block_at(provider, portal, coordinate).await?);
```
0xrusowsky: this calls the provider to get blocks by number and then by hash, but we already have the hash info when calling them by number, so we could save rpc calls to speed up sync

crates/checker/src/runtime.rs
```diff
@@ -0,0 +1,402 @@
...
+    let tempo_parent = BlockNumHash::from(prior.metadata.imported_tempo);
+    let history = collect_l1_history(l1, config.portal_address, tempo_parent, tempo)
+        .await
+        .map_err(BlockError::Retry)?;
+    if history.is_empty() {
+        return Err(fail(eyre::eyre!("Zone block did not advance Tempo")));
+    }
+    let mut block_effects = effects::from_tempo_history(&history);
+    block_effects
+        .extend(effects::from_zone(l2.bridge_events()).map_err(|error| fail(error.into()))?);
```
0xrusowsky: as a follow-up we could use fields such as `DepositMade.deposit_number`, `DepositOutcome.deposit_hash`, `TempoAdvanced.deposits_processed`, etc to harden checks beyond accounting aggregates

crates/checker/src/runtime.rs
```diff
@@ -0,0 +1,402 @@
...
+    let next = store
+        .apply(
+            prior,
+            zone,
+            BlockRef::new(number.saturating_sub(1), parent_hash),
+            tempo.into(),
+            prior.metadata.imported_tempo,
+            &block_effects,
+        )
+        .map_err(|error| BlockError::Retry(error.into()))?;
```
0xrusowsky: candidate already contains the verified post-block state, but `Store::apply()` clones the old ledger and applies the same effects again.

could we pass candidate and its `BlockDelta` into the store and persist them directly to avoid the full-ledger clone?

0xrusowsky: maybe we could have smth like:
```
   struct CandidateTransition {
       coordinates: TransitionCoordinates,
       state: State,
       delta: BlockDelta,
   }
```

crates/checker/src/accounting/effects.rs
grandizzy: changed in ca56f294

crates/checker/src/accounting/effects.rs
grandizzy: changed in 10cee45c

crates/checker/src/accounting/mod.rs
grandizzy: these were missing, added in ee11c133

crates/checker/src/runtime.rs
grandizzy: pls see d5c3fa56 did it a little bit different

crates/checker/src/runtime.rs
grandizzy: yep, agree, let's follow-up and keep the scope of the 1st PR narrow to be reviewed easier

crates/checker/src/l1/mod.rs
grandizzy: changed in a9512a7e

crates/checker/src/l1/mod.rs
grandizzy: pls see 6f305cef

crates/checker/src/runtime.rs
0xrusowsky: great! 

looking more into this, `CandidateTransition` duplicates most of `DeltaRecord` and `Store::apply` immediately converts between them. Let's directly:
```rs
   struct CandidateTransition {
       state: State,
       record: DeltaRecord,
   }
```
wdyt?

grandizzy: yes, makes sense, will bundle this change with perf improvement to avoid cloning the complete account ledger

crates/checker/src/accounting/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+    pub(crate) const fn new(token: Address, account: Address) -> Self {
+        Self { token, account }
+    }
+}
+
+/// Aggregate circulating and in-flight liabilities for one token.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) struct TokenState {
+    /// Token was authenticated through Portal creation or enablement.
+    pub(crate) enabled: bool,
```
0xrusowsky: imo we should make `State::tokens` represent authenticated token enablement. Right now account/liability mutations can create a default `TokenState`, so map membership and enabled flag are 2 different sources of truth.

since any zone token must first be enabled, so we can simplify the current model by:
- removing `TokenState::enabled` and `is_empty()`
- making `EnableToken` the only insertion path
- retaining zero-valued token rows (cause map membership means enabled)
- returning `UnknownToken` when the token is absent

crates/checker/src/l2/events.rs
```diff
@@ -0,0 +1,1417 @@
...
+    }
+
+    pub(crate) fn block_number(&self) -> u64 {
+        self.tempo_block_number
+    }
+}
+
+/// Semantic value of one recognized Zone Inbox or Outbox event.
+#[derive(Debug)]
+pub(crate) enum L2BridgeEvent {
```
0xrusowsky: this enum currently mixes block control data (`TempoAdvanced`), receipt-local authentication evidence (`TokenBurn`), and semantic activity consumed after authentication.

it also carries ABI-level representations into downstream consumers:
- `DepositOutcome` permits inconsistent `processed`/`recipient` combinations
- `WithdrawalBounceBack` similarly encodes its outcome with a boolean
- `WithdrawalRequested` is classified using `sender == Address::ZERO`, while the protocol identifies deposit bounce-backs by the canonical `sender == 0 && fallbackNonce == 0` combination. the checker currently drops `fallbackNonce`

i think we could simplify consumers if we normalize the evidence when promoting decoded receipt events into `L2BlockEvidence`:
- store the required Tempo anchor directly on `L2BlockEvidence`, making it non-optional and removing it from the event stream
- keep burns private to receipt authentication and discard them afterward
- expose canonical transfers separately from authenticated bridge actions
- replace correlated flags/options with semantic outcome enums:
```rs
enum DepositResult {
    Processed { recipient: Address },
    Failed,
}
```
- decode and validate withdrawal origin explicitly:
```rs
enum WithdrawalOrigin {
    User { sender: Address },
    DepositBounceBack,
}
```
where `(sender == 0, fallbackNonce == 0)` is a bounce-back, `(sender != 0, fallbackNonce != 0)` is a user withdrawal, and mixed combinations are rejected
- normalize bridge amounts to `U256` at this boundary

this should make invalid combinations unrepresentable and remove repeated flag/sentinel handling from authentication, accounting, and telemetry. it should also eliminate `anchor_index`, the redundant missing-anchor check, and the ignored `TempoAdvanced`/`TokenBurn` arms.

crates/checker/src/runtime.rs
grandizzy: added in e252696818fb88a173d9aaaca62b9a328163dfbb + avoid cloning the complete account ledger

crates/checker/src/l2/events.rs
```diff
@@ -0,0 +1,1417 @@
...
+            };
+            if let Some(event) = event {
+                if matches!(event, L2BridgeEvent::TempoAdvanced(_)) {
+                    eyre::ensure!(
+                        self.anchor_index.is_none(),
+                        "duplicate TempoAdvanced in block {block}"
+                    );
+                    self.anchor_index = Some(self.events.len());
+                }
+                self.events.push(L2EventEvidence {
```
0xrusowsky: since we already have the receipt boundary in `extract_receipt`, could we authenticate a receipt-local event buffer here before appending its events to the block?

currently we attach `transaction_index` to every event, flatten all receipts, and then reconstruct the receipt boundaries twice with `chunk_by` in `L2Events::{authenticate_mints, authenticate_withdrawals}`.

concretely, this could:

1. decode this receipt into a local `Vec<L2BridgeEvent>`
2. call the mint and withdrawal authenticators with the transaction hash and that slice
3. append the authenticated events to the block-level vector

that should let us remove `transaction_index`, the `usize as u32` cast, both `chunk_by` passes, and likely `L2EventEvidence` entirely since its transaction hash is only needed while producing authentication errors.

this seems like a small, self-contained first step toward the evidence normalization discussed in https://github.com/tempoxyz/zones/pull/1216#discussion_r3811287116.

crates/checker/src/accounting/mod.rs
grandizzy: accomodated in 5b87d04a7ded0e70281d6a986e72a6dc2eabd100

crates/checker/src/accounting/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+        match self {
+            Self::Credit(amount) => value.checked_add(amount).ok_or(AccountingError::Overflow),
+            Self::Debit(amount) => value.checked_sub(amount).ok_or(AccountingError::Underflow),
+        }
+    }
+}
+
+/// One independently authenticated accounting change.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) enum Effect {
```
0xrusowsky, line 88: wdyt about we collapsing `Effect` into a smaller accounting primitive?

```rs
enum LiabilityKind {
    Deposit,
    Withdrawal,
    TempoRefund,
    ZoneRefund,
}

enum Effect {
    EnableToken(Address),
    Account(AccountKey, BalanceChange),
    Liability(Address, LiabilityKind, BalanceChange),
}
```

since `State::apply()` already makes the whole effect list atomic, a transfer can simply emit debit + credit.

this would remove the separate credit/debit/transfer and four `Pending*` variants, simplify `apply_effect`, and allow effect derivation to happen in a single pass.

crates/checker/src/persistence/mod.rs
```diff
@@ -0,0 +1,521 @@
...
+            metadata.observed_zone
+        } else {
+            finding.zone
+        };
+        metadata.observed_zone = observed_through;
+        metadata.status = Status::Diverged {
+            first_unchecked: finding.zone,
+            observed_through,
+        };
+        tx.put::<Findings>(ACTIVE_FINDING, finding)?;
```
0xrusowsky: could we make `Metadata` the single source of truth for an active divergence? 

right now the same state is persisted in three places: `metadata.observed_zone` duplicates `Status::Diverged::observed_through`, `finding.zone` duplicates `Status::Diverged::first_unchecked`, and the full finding is stored separately in this singleton `Findings` row. `record_finding()` and `observe_diverged()` therefore have to keep these copies synchronized, while `Store::load()` reloads only `Metadata` and never reloads or validates this finding row.

could we instead embed the finding in the status, use `Metadata::observed_zone` as the sole observed-through coordinate, and use `finding.zone` as the sole first-unchecked coordinate? 
```rs
enum Status {
   Verifying
   Diverged { finding: Finding }
}
```

that should let us remove `observed_through`, `first_unchecked`, `ACTIVE_FINDING`, and potentially the entire `CheckerFindings` table plus its synchronization code

crates/checker/src/bootstrap.rs
```diff
@@ -0,0 +1,271 @@
...
+    );
+    eyre::ensure!(
+        event.initialToken == initial_token,
+        "Portal initial token does not match Zone genesis"
+    );
+    Ok(BlockNumHash::new(number, hash))
+}
+
+/// Require each receipt's provenance to match the block it was fetched from.
+fn authenticate_receipts(
```
0xrusowsky: this duplicates the receipt-provenance validation already implemented by `l1::validate_l1_receipts`.

could we expose and reuse that helper here so transaction count, block hash/number, transaction hash, and receipt index are validated in one place?

grandizzy: yep, cleaned up in 398ee118d747ae1fde67dc3e1c4fdaf71002497a

crates/checker/src/runtime.rs
```diff
@@ -0,0 +1,440 @@
...
+
+/// Result of processing one notification's blocks.
+enum Outcome {
+    Applied(Box<Snapshot>),
+    /// Reorg exceeded retained history; state must be rebuilt from the checkpoint.
+    Rebuild,
+}
+
+/// Failure processing one block.
+enum BlockError {
```
0xrusowsky, line 285: I wonder if we need a third `Fatal` variant. Right now, anything that is not an authenticated finding seems to become retryable, including errors that another attempt cannot resolve—for example, corrupted persistence state or an invalid delta chain. Bootstrap has a similar issue because configuration and genesis validation are also wrapped in `retry_forever`.

should we distinguish transient RPC/state-acquisition failures from unrecoverable checker errors, and let the latter escape `run_inner` so the checker is disabled and drained as documented?

crates/checker/src/l2/events.rs
grandizzy: did it in f0560a77d33f6c53372c5915fd1868a07e1fef22

crates/checker/src/l2/events.rs
grandizzy: 66d51ecf44cf88b5f7fde066e2f83ab22b34b967

contrib/checker-lab/checker-lab.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
```
legion2002: can we convert all these bash files into testable rust scripts?

crates/checker/src/accounting/mod.rs
grandizzy: added in 19c3c2598e8749b308ffb3a2a104f8759cf21c1c

crates/checker/src/persistence/mod.rs
grandizzy: added in 3f942d6413507816487f9f336ba231a164f73a38

crates/checker/src/runtime.rs
grandizzy: introduced in 4d54785c2b4ec1cbae52ec8b56175ccb3ba5a507 as Disabled

grandizzy:
cyclops audit note="Review this PR with extra focus on checker correctness against the Zones specification and precompile semantics, node availability under RPC, persistence, reorg and malformed-data failures, and performance, memory and I/O scalability. Prioritize concrete correctness and denial-of-service findings over style."

crates/checker/src/persistence/mod.rs
```diff
@@ -0,0 +1,474 @@
...
+            state: Arc::new(state),
+        })
+    }
+
+    /// Unwind verified state to one retained exact ancestor.
+    pub(crate) fn reorg(
+        &self,
+        prior: &Snapshot,
+        ancestor: BlockRef,
+    ) -> Result<Snapshot, PersistenceError> {
```
0xrusowsky: reorgs are not possible in zones afaik, because we only anchor to finalized L1 blocks

crates/checker/src/accounting/effects.rs
```diff
@@ -0,0 +1,411 @@
...
+                    kind: LiabilityKind::Deposit,
+                    change: BalanceChange::Debit(U256::from(amount) + U256::from(bounceback_fee)),
+                });
+                effects.push(Effect::Liability {
+                    token,
+                    kind: LiabilityKind::TempoRefund,
+                    change: BalanceChange::Credit(U256::from(amount)),
+                });
+            }
+            L1PortalEvent::RefundClaimed { token, amount, .. } => {
```
grandizzy, line 95: https://github.com/tempoxyz/zones/pull/1216/changes/93625a727c5ac6b47807aa7daabe0159ac915b13

crates/checker/src/l2/events.rs
```diff
@@ -0,0 +1,1270 @@
...
+        &mut self,
+        transaction: &T,
+        receipt: &R,
+        block: u64,
+    ) -> eyre::Result<()>
+    where
+        T: TxHashRef,
+        R: TxReceipt<Log = Log>,
+    {
+        if !receipt.status() {
```
grandizzy: https://github.com/tempoxyz/zones/pull/1216/changes/ad0e79317c33b1e1a740213619dc62a718edefe1

0xalpharush:
cyclops audit note="cross check against https://github.com/tempoxyz/zones/blob/main/xtask/src/verify_portal_backing.rs"

crates/checker/src/persistence/mod.rs
grandizzy: crosschecked that's the case and removed in https://github.com/tempoxyz/zones/pull/1216/changes/0859c1bb4e23c21bd97450e5a886b0d77c3a6441

crates/checker/src/l1/mod.rs
```diff
@@ -0,0 +1,354 @@
...
+            block.header().number()
+        )));
+    }
+    if block.header().parent_hash() != parent.hash {
+        return Err(unavailable(eyre::eyre!(
+            "Tempo history is not contiguous at block {number}"
+        )));
+    }
+    let coordinate = BlockNumHash::new(number, block.header().hash());
+    if expected.is_some_and(|expected| coordinate != expected) {
```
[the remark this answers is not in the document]
decofe: c02ebac2dc153ca65b4bb76d87a9c73358ef47f9

crates/checker/src/l2/state.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/// Read exact balances and supply for affected accounts at one Zone block.
+pub(crate) fn read_accounting_state<P: StateProviderFactory>(
+    provider: &P,
+    accounts: &BTreeMap<Address, BTreeSet<Address>>,
+    block: BlockNumHash,
+    spec: TempoHardfork,
+) -> Result<Vec<TokenAccountingEvidence>, AccountingStateError> {
+    let mut state = provider
```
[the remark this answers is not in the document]
decofe: 1679e83f446942aa7616f9035b97b7381914af89

crates/checker/src/runtime.rs
```diff
@@ -0,0 +1,476 @@
...
+    let (store, mut snapshot) = Store::open_or_create(&config.database_path, &checkpoint)?;
+    metrics.update(&snapshot);
+
+    ctx.catch_up_notifications_with_head(ExExHead::new(snapshot.metadata.verified_zone.into()))?;
+    ctx.send_finished_height(snapshot.metadata.verified_zone.into())?;
+
+    while let Some(notification) = ctx.notifications.try_next().await? {
+        let delivered_tip = notification_tip(&notification)
+            .ok_or_else(|| eyre::eyre!("received an empty ExEx notification"))?;
+        let recoverable_finding = match &snapshot.metadata.status {
```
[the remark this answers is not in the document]
decofe: N/A: Zones have instant finality, so offline canonical block replacement cannot occur.

crates/checker/src/bootstrap.rs
```diff
@@ -0,0 +1,307 @@
...
+    config: &CheckerConfig,
+    creation: BlockNumHash,
+    anchor: BlockNumHash,
+    initial_token: Address,
+) -> Result<State, AttemptError> {
+    let mut state = State::default();
+    state
+        .apply(&[crate::accounting::Effect::EnableToken(initial_token)])
+        .map_err(AttemptError::disable)?;
+    if anchor.number < creation.number {
```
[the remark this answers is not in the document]
decofe: N/A: already fixed on the current PR head.

contrib/checker-lab/checker-lab.sh
grandizzy: @legion2002 this is more for local usage like to start env and inject bugs / see if checker flags them, would it be fine to drop them altogether and follow up in order to keep scope PR narrow/easier to review?

grandizzy:
marking as rfr, there are some cyclops findings re exex edge cases but would like to avoid over optimization and fix them case by case in dev env

contrib/checker-lab/checker-lab.sh
legion2002: sgtm

crates/checker/DESIGN.md
```diff
@@ -0,0 +1,127 @@
...
+The checker stores an expected balance for each known `(token, account)` and
+five aggregate liability buckets per token:
+
+| Bucket | Meaning |
+| --- | --- |
+| `account_total` | Sum of circulating Zone account entitlements |
+| `pending_deposits` | Custody received on Tempo but not resolved on the Zone |
+| `pending_withdrawals` | Burned on the Zone but not resolved on Tempo |
+| `pending_tempo_refunds` | Deposit refunds parked on Tempo |
+| `pending_zone_refunds` | Withdrawal bounce-backs returning to or parked on the Zone |
```
legion2002, line 30: do we need a fee bucket here also? currently the fee is 0 so it doesn't matter that much, but if we turn on the sequencer fees for deposits, withdrawals or bouncebacks then we'll need it for the solvency checker?

crates/checker/src/accounting/effects.rs
```diff
@@ -0,0 +1,434 @@
...
+
+/// Derive liability changes from one authenticated Tempo block.
+pub(crate) fn from_tempo(block: &L1BlockEvidence) -> Vec<Effect> {
+    from_tempo_events(block.portal_events())
+}
+
+/// Derive liability changes from authenticated Tempo Portal events.
+fn from_tempo_events<'a>(events: impl IntoIterator<Item = &'a L1PortalEvent>) -> Vec<Effect> {
+    let mut effects = Vec::new();
+    for event in events {
```
legion2002, line 45: do we need any fee events here?

contrib/checker-lab/checker-lab.sh
grandizzy: removed in https://github.com/tempoxyz/zones/pull/1216/changes/b72a00d3360842720b18d14d975153962a69be9e

crates/checker/src/persistence/codec.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// Validate a dynamic value before passing it to MDBX's infallible codec interface.
+pub(super) fn validate<T: Serialize>(value: &T) -> Result<(), String> {
+    options()
+        .serialized_size(value)
+        .map(|_| ())
+        .map_err(|error| error.to_string())
+}
+
+fn encode<T: Serialize>(value: &T) -> Result<Vec<u8>, CodecError> {
```
legion2002, line 25: still looking into this, but do you know if we can switch to reth::compact for things, and use the macros instead of this manual codec?

crates/checker/src/l1/events.rs
```diff
@@ -0,0 +1,518 @@
...
+use alloy_primitives::{Address, Log};
+use alloy_sol_types::SolEvent;
+use tempo_alloy::rpc::TempoTransactionReceipt;
+use tempo_zone_contracts::ZonePortal;
+
+use crate::decode_event;
+
+/// Semantic value of one recognized Zone Portal event.
+#[derive(Debug)]
+pub(crate) enum L1PortalEvent {
```
legion2002, line 13: if this is intended to run as an exex, can we just reuse the L1 event detection and parsing from the existing zone<> tempo connection?

crates/checker/DESIGN.md
grandizzy: for the initial version we discussed to keep it narrow to solvency (`custody >= liabilities`) do you think should address right now or OK to follow-up for fees?

crates/checker/src/persistence/codec.rs
grandizzy: good point, will look into, the only hard requirement would be to gracefully handle errors so a fatal error would result in checker disabled / alert only instead halting the node

crates/checker/src/l1/events.rs
grandizzy: merged https://github.com/tempoxyz/zones/pull/1257

crates/checker/DESIGN.md
legion2002: yup can do as a follow up

crates/node/src/cli.rs
```diff
@@ -184,6 +186,13 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
         builder.config_mut().rpc.rpc_max_logs_per_response = MAX_LOGS_PER_RESPONSE.into();
         builder.config_mut().rpc.rpc_max_blocks_per_filter = MAX_BLOCKS_PER_FILTER.into();
 
+        // Clone L1 connection details for the checker before they are moved into ZoneNode.
+        let checker_l1_rpc_url = args.l1_rpc_url.clone();
```
legion2002: fine to keep as is, but pointing out that this means that a shared RPC means that a compromise could forge headers for both the zone and the checker

crates/checker/src/persistence/model.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub(crate) observed_zone: BlockRef,
+    pub(crate) status: Status,
+}
+
+impl Metadata {
+    /// Classify one block coordinate against the verified tip.
+    pub(crate) fn classify(&self, number: u64, hash: B256) -> AppliedStatus {
+        if number > self.verified_zone.number {
+            AppliedStatus::New
+        } else if number == self.verified_zone.number && hash != self.verified_zone.hash {
```
legion2002, line 89: this shouldn't happen in normal operations right?
should we be sending `BlockError::Disable` if this happens

crates/checker/src/persistence/codec.rs
```diff
@@ -0,0 +1,80 @@
...
+    let mut encoded = vec![VERSION];
+    options().serialize_into(&mut encoded, value).map_err(map)?;
+    Ok(encoded)
+}
+
+fn decode<T: DeserializeOwned>(encoded: &[u8]) -> Result<T, CodecError> {
+    let (&version, value) = encoded.split_first().ok_or(CodecError::Empty)?;
```
legion2002, line 32: do we need this codec versioning for every value? can't we just bump the schema version and recreate the DB

legion2002 (dismissed):

crates/checker/src/persistence/model.rs
grandizzy: good catch, fixed in https://github.com/tempoxyz/zones/pull/1216/changes/04c88c6f60cf50357bf3570b9d452bb7da2eecb3

crates/node/src/cli.rs
grandizzy: will extend in a follow-up

grandizzy:
cyclops audit fast

crates/checker/src/l1/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+            "Tempo RPC returned block {} for requested block {number}",
+            block.header().number()
+        )));
+    }
+    if block.header().parent_hash() != parent.hash {
+        return Err(finding(eyre::eyre!(
+            "Tempo history is not contiguous at block {number}"
+        )));
+    }
+    let coordinate = BlockNumHash::new(number, block.header().hash());
```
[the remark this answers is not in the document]
legion2002: think this is valid. it was the main reason to use the authenticated L1 node tracker as the default, but maybe we should also add this to the fallback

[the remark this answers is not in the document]
legion2002: we can fix as follow up

legion2002 (dismissed):

crates/checker/src/l1/mod.rs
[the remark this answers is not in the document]
grandizzy: took the opportunity of rebase needed and fixed in https://github.com/tempoxyz/zones/pull/1216/changes/f7208a3200d6d091eea088b72e0a770510c8d10b

grandizzy:
cyclops audit fast

legion2002 (approved):

PR #1217, opened 2026-08-18
feat(zone): TIP-1096

shekhirin:
- Adds the T12-gated `ZoneInbox::advanceTempoHeaders` operation with bounded, consecutive-header validation.
- Introduces an authenticated token-enablement cursor across `ZoneInbox`, `ZonePortal`, settlement attestations, and batch certificates.
- Bounds outstanding deposits and token enablements so deferred portal work cannot exceed proof or execution capacity, including reserved headroom for withdrawal bounce-backs.
- Emits and decodes the T12 `TempoAdvanced` and `BatchSubmitted` events carrying the processed-token cursor.
- Preserves T11 compatibility through legacy events, legacy token admission rules, and pre-T12 timestamp behavior.
- Selects the legacy or T12 `ZonePortal` submission ABI and attestation format from the live Tempo L1 hardfork, including batches spanning the activation boundary.
- Initializes the T12 token cursor from a complete authenticated historical prefix while preserving legacy T11 cursor semantics.

Follow-up PRs add checkpoint-only payload execution, proving, and coordinated production.


shekhirin:
cyclops audit fast

shekhirin:
cyclops audit fast note="use https://github.com/tempoxyz/tempo/pull/7235 for full tip-1096 spec"

shekhirin:
cyclops audit note="1096 from https://github.com/tempoxyz/tempo/pull/7254"

shekhirin:
cyclops audit

crates/precompiles/src/storage.rs
```diff
@@ -189,6 +189,26 @@ impl<P: L1StorageReader> L1State<P> {
         self.read_l1(select_slot(&portal))
     }
 
+    /// Reads the length of the portal's append-only `_enabledTokens` array (storage slot 7).
+    pub fn portal_enabled_token_count(&self) -> tempo_precompiles::Result<usize> {
```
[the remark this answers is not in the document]
shekhirin: addressed in https://github.com/tempoxyz/zones/pull/1217/commits/27426f8bca6aa8cdd80baca13a29115566d41d8f, needs https://github.com/tempoxyz/tempo/pull/7231 to get merged

crates/precompiles/src/tempo_state.rs
```diff
@@ -176,7 +208,7 @@ impl TempoState {
                 TempoStateAbi::TempoStateCalls {
                     tempoBlockHash(call) => view(call, |_| self.tempo_block_hash.read()),
                     tempoBlockNumber(call) => view(call, |_| self.tempo_block_number.read()),
-                    finalizeTempo(call) => self.apply_checkpoint(l1, msg_sender, call),
+                    finalizeTempo(call) => self.apply_checkpoint(l1, msg_sender, call, z1_active),
```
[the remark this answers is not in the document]
shekhirin: addressed in https://github.com/tempoxyz/zones/pull/1217/changes/e4feb8535883b641891c39702e629d9fd97f58c1

crates/contracts/src/runtime/tempo/ZonePortal.sol
```diff
@@ -1373,6 +1410,7 @@ contract ZonePortal is IZonePortal {
                 withdrawalBatchIndex + 1,
                 blockTransition,
                 depositQueueTransition,
+                tokenEnablementTransition,
```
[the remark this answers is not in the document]
shekhirin: will be addressed in https://github.com/tempoxyz/zones/pull/1244

crates/contracts/src/runtime/tempo/ZonePortal.sol
```diff
@@ -1294,6 +1311,7 @@ contract ZonePortal is IZonePortal {
         uint64 recentTempoBlockNumber,
         BlockTransition calldata blockTransition,
         DepositQueueTransition calldata depositQueueTransition,
+        TokenEnablementTransition calldata tokenEnablementTransition,
```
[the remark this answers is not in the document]
shekhirin: not an issue

shekhirin:
cyclops audit fast

shekhirin:
cyclops audit fast

crates/sequencer/src/settlement.rs
```diff
@@ -1524,39 +1623,75 @@ pub(crate) fn read_zone_block_snapshot<P: ZoneSequencerProvider>(
...
+                        event.newProcessedDepositQueueHash,
+                        event.lastProcessedDepositNumber,
+                        event.lastProcessedEnabledTokenCount,
+                    )
+                }
+                LegacyTempoAdvanced::SIGNATURE_HASH => {
+                    let event = LegacyTempoAdvanced::decode_log(log).map_err(|err| {
+                        eyre::eyre!("invalid legacy TempoAdvanced log in block {number}: {err}")
+                    })?;
+                    settlement_abi = Some(SettlementAbi::Legacy);
```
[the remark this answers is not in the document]
shekhirin: fixed in https://github.com/tempoxyz/zones/pull/1217/commits/edfcf0eab922e6d7dfbff8632be327405c53b705

crates/precompiles/src/inbox/mod.rs
```diff
@@ -173,17 +247,46 @@ impl ZoneInbox {
...
+                depositsProcessed: U256::from(deposit_count),
+                newProcessedDepositQueueHash: current_hash,
+                lastProcessedDepositNumber: processed_number,
+            })?;
+        }
 
         Ok(())
     }
 
+    /// Authenticate a bounded consecutive header range without observing Tempo state or applying
```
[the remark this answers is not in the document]
shekhirin: will be fixed in the next PR

shekhirin:
cyclops audit fast

crates/sequencer/src/settlement.rs
```diff
@@ -1166,6 +1245,34 @@ pub struct WithdrawalPage {
...
+pub enum SettlementAbi {
+    /// Pre-T12 selector and EIP-712 statement.
+    Legacy,
+    /// T12 selector and token-enablement-bound EIP-712 statement.
+    T12,
+}
+
+impl SettlementAbi {
+    /// Resolve the settlement selector and attestation format from the live Tempo L1 hardfork.
+    pub async fn from_l1(provider: &DynProvider<TempoNetwork>) -> Result<Self> {
```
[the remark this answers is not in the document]
shekhirin: I believe this is a no-fix:
- If a legacy pre-T12 ABI batch is submitted to T12 L1, then it would be rejected and we will retry the batch submission with new ABI
- If a T12 ABI batch is submitted to an L1 with latest block not transitioned into T12 and only transitioning on the next block, it would be accepted correctly in the next block that's going to be T12

crates/node/src/settlement_attestation.rs
```diff
@@ -134,19 +139,35 @@ where
     let mut tempo_block_number = None;
     let mut processed_deposit_hash = None;
     let mut processed_deposit_number = None;
+    let mut processed_token_count = None;
     let mut withdrawal = None;
 
     for receipt in receipts {
```
rkrasiuk, line 145: unrelated to this PR specifically, but it doesn't make sense to me that we are iterating over all receipts here. we should just check first and last

crates/node/src/settlement_attestation.rs
```diff
@@ -157,7 +178,13 @@ where
         }
     }
 
-    Ok(BlockCommitments {
+    // Checkpoint-only blocks have no BatchFinalized or TempoAdvanced event. They are not
+    // settlement boundaries and must remain transparent to the previous-boundary scan.
+    let Some(withdrawal) = withdrawal else {
+        return Ok(None);
+    };
```
rkrasiuk, line 185: there can a non-checkpoint block with TempoAdvanced, but no withdrawal

rkrasiuk (changes requested):

crates/node/src/settlement_attestation.rs
```diff
@@ -134,19 +139,35 @@ where
...
+                    }
+                    Some(LegacyTempoAdvanced::SIGNATURE_HASH) => {
+                        let event = LegacyTempoAdvanced::decode_log(log).wrap_err_with(|| {
+                            format!("invalid legacy TempoAdvanced log in block {number}")
+                        })?;
+                        anchor_hash = Some(event.tempoBlockHash);
+                        tempo_block_number = Some(event.tempoBlockNumber);
+                        processed_deposit_hash = Some(event.newProcessedDepositQueueHash);
+                        processed_deposit_number = Some(event.lastProcessedDepositNumber);
+                        processed_token_count = Some(0);
```
klkvr, line 167: setting 0 here seems footgunish, should we maube treat it as None and then properly assert that it must ne None pre-T12 and Some post-T112?

crates/precompiles/src/tempo_state.rs
```diff
@@ -177,10 +201,12 @@ impl TempoState {
...
                     #[schedule(until = T12)]
-                    finalizeTempo_0(call) => self.apply_checkpoint(l1, msg_sender, call),
+                    finalizeTempo_0(call) => {
+                        self.apply_checkpoints(l1, msg_sender, core::slice::from_ref(&call.header))
+                    },
                     #[schedule(since = T12)]
-                    finalizeTempo_1(_) => {
-                        tempo_precompiles::dispatch::unknown_selector_result(calldata)
+                    finalizeTempo_1(call) => {
+                        self.apply_checkpoints(l1, msg_sender, &call.headers)
```
klkvr, line 209: we can probably just fully disable it after T12 because nobody is actually invoking this overload after T12 with inbox being a native precompile

crates/node/src/settlement_attestation.rs
shekhirin: this PR doesn't change this behaviour, two places where we checked the `withdrawal` field are these:
1. https://github.com/tempoxyz/zones/blob/368cdf56cbbe0ff951e3e96f33acb5dd771c8402/crates/node/src/settlement_attestation.rs#L181-L193
2. https://github.com/tempoxyz/zones/blob/368cdf56cbbe0ff951e3e96f33acb5dd771c8402/crates/node/src/settlement_attestation.rs#L209-L213

and we short-circuited it in the same way

why is this correct: pre-1096, a batch boundary is either a block with withdrawals, or every 120th block. Both of them would emit a `BatchFinalized` event, even if there was no withdrawals like in 120th block case.

crates/node/src/settlement_attestation.rs
shekhirin: yeah, I think this is to avoid hardcoding an invariant that `TempoAdvanced` is always first and `BatchFinalized` is always last, but I agree that we should just add tests for that, and check first and last

crates/precompiles/src/inbox/mod.rs
```diff
@@ -94,26 +99,88 @@ impl ZoneInbox {
...
-            && l1.read_portal(|portal| &portal.token_enablement_hash)? != next_token_enablement_hash
-        {
-            return Err(ZoneInboxError::invalid_token_enablement_hash().into());
+        if !portal.is_zero() {
+            if !StorageCtx.spec().is_t12() {
+                if l1.read_portal(|portal| &portal.token_enablement_hash)?
+                    != next_token_enablement_hash
+                {
+                    return Err(ZoneInboxError::invalid_token_enablement_hash().into());
+                }
```
klkvr: should we just always compare the hash here? even if it's T12

crates/precompiles/src/inbox/mod.rs
```diff
@@ -94,26 +99,52 @@ impl ZoneInbox {
...
+        let portal_enabled_token_count = if !portal.is_zero() {
+            if l1.read_portal(|portal| &portal.token_enablement_hash)? != next_token_enablement_hash
+            {
+                return Err(ZoneInboxError::invalid_token_enablement_hash().into());
+            }
+            StorageCtx
+                .spec()
+                .is_t12()
+                .then(|| l1.read_portal_vec_len(|portal| &portal.enabled_tokens))
+                .transpose()?
```
klkvr: always reading portal's array length seems wasteful, if we already have non-zero `processed_enabled_token_count` we could just increment it by `call.enabledTokens.len()` i think?

klkvr (approved):

crates/sequencer/src/settlement.rs
```diff
@@ -337,7 +342,8 @@ impl BatchSubmitter {
         &self,
         batch: &BatchData,
         shutdown: &sync::CancellationToken,
-    ) -> std::result::Result<LegacyBatchSubmitted, BatchSubmitError> {
+    ) -> std::result::Result<BatchSubmitted, BatchSubmitError> {
+        let settlement_abi = SettlementAbi::from_l1(&self.l1_provider).await?;
```
adityapk00, line 346: I might be misunderstanding this, but this selects the ABI from the L1 block, but the zone execution selects T12 activation from the zone block timestamp (which can be higher than the L1 block's timestamp). So during cutover, a zone block can already emit a nonzero cursor, while the latest L1 is still T11. 

If we submitbatch during this block, the block gets accepted, but does not initialie the portal's cursor, right? the local monitor advances, and then when the L1 activates T12, the next submit batch uses that value, while the portal requires 0, and it might brick?

adityapk00 (approved):

crates/node/src/settlement_attestation.rs
rkrasiuk: we already have that invariant and multitude of checks for it in the executor

rkrasiuk (approved):

crates/node/src/settlement_attestation.rs
rkrasiuk: fair

crates/sequencer/src/settlement.rs
shekhirin: that's right, will be addressed in https://github.com/tempoxyz/zones/pull/1245 by waiting for L1 to advance to hardfork boundary block in case when `hardfork_at(SystemTime::now()) >= T12 && l1_tip.hardfork < T12`

PR #1220, opened 2026-08-18
refactor(precompiles): internalize crypto + zone tx ctx

0xrusowsky:
this PR:
- internalizes the crypto precompiles
- internalizes the zone tx ctx precompile
- registers all precompiles using `set_precompile_lookup` like upstream tempo does


0xrusowsky:
cyclops audit fast note="zones are not live yet and we will regenesis, so it doesn't matter that it is an STF breaking change"

0xKitsune (dismissed):

klkvr (dismissed):

crates/precompiles/src/lib.rs
```diff
@@ -134,45 +134,40 @@ pub fn extend_zone_precompiles<P>(
...
+
         if is_tip20_prefix(*address) {
             Some(create_tip20_precompile(*address, &env))
-        } else if *address == STABLECOIN_DEX_ADDRESS {
+        } else if *address == TIP20_FACTORY_ADDRESS
+            || *address == TIP_FEE_MANAGER_ADDRESS
+            || *address == TIP20_CHANNEL_RESERVE_ADDRESS
+            || *address == STABLECOIN_DEX_ADDRESS
+        {
             None
```
klkvr: can we just remove this branch and let it fallback to last `None` branch?

0xrusowsky: addressed in https://github.com/tempoxyz/zones/pull/1220/commits/44ecba82dc3b0e0a00acfdd9425854c054dd005a

0xrusowsky:
cyclops audit fast note="zones are not live yet, and we can simply regenesis. we don't expect any external consumers of the internalized precompiles. They were initially set as precompiles rather than internal impl details, because originally the `ZoneInbox` and the `ZoneOutbox` were contracts rather than native precompiles"

klkvr (approved):

crates/precompiles/src/lib.rs
```diff
@@ -134,45 +133,26 @@ pub fn extend_zone_precompiles<P>(
...
+        #[cfg(feature = "std")]
+        if *address == ZONE_OUTBOX_ADDRESS {
+            return Some(create_outbox_precompile(l1.clone(), &env));
+        }
+
         if is_tip20_prefix(*address) {
             Some(create_tip20_precompile(*address, &env))
-        } else if *address == STABLECOIN_DEX_ADDRESS {
-            None
+        } else if *address == TEMPO_STATE_ADDRESS {
```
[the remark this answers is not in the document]
0xrusowsky: this is on purpose to match upstream tempo

PR #1229, opened 2026-08-19
perf(sequencer): backrun withdrawals with ordered 2D nonces

0xKitsune:
This PR uses an isolated 2D nonce lane to backrun `submitBatch` with ordered withdrawal processing in the same Tempo block. It falls back to the existing recovery path when the fast path is unavailable or uncertain.


0xKitsune:
cyclops audit

PR #1242, opened 2026-08-21
feat(node): build checkpoint-only Tempo imports

shekhirin:
- Models Tempo imports as either a full operational import or a checkpoint-only header range.
- Builds `advanceTempoHeaders` system transactions from bounded, consecutive Tempo headers.
- Validates block-number and parent-hash continuity across imported header ranges.
- Keeps checkpoint-only Zone blocks system-only by excluding user transactions and withdrawal finalization.
- Aggregates deferred deposits and token enablements for the next full import while enforcing protocol capacity limits.
- Finalizes the withdrawal batch on the first full import following checkpoint-only blocks.
- Preserves normal `advanceTempo` and withdrawal-batch behavior when the Zone follows Tempo without a gap.

### DIFF
--- a/crates/l1/src/block.rs
+++ b/crates/l1/src/block.rs
@@ -10,6 +10,36 @@ pub struct L1BlockDeposits {
 }
 
 impl L1BlockDeposits {
+    /// Prepare portal work accumulated across checkpoint-only blocks for one full import.
+    pub async fn prepare_many(
+        blocks: Vec<Self>,
+        encryption_keys: &EncryptionKeyRing,
+        portal_address: Address,
+    ) -> eyre::Result<PreparedL1Block> {
+        let mut blocks = blocks.into_iter();
+        let first = blocks
+            .next()
+            .ok_or_else(|| eyre::eyre!("cannot prepare an empty L1 range"))?;
+        let mut prepared = first.prepare(encryption_keys, portal_address).await?;
+        for block in blocks {
+            let next = block.prepare(encryption_keys, portal_address).await?;
+            prepared.header = next.header;
+            prepared.queued_deposits.extend(next.queued_deposits);
+            prepared.decryptions.extend(next.decryptions);
+            prepared.enabled_tokens.extend(next.enabled_tokens);
+        }
+        eyre::ensure!(
+            prepared.queued_deposits.len() <= zone_primitives::constants::MAX_UNPROCESSED_DEPOSITS,
+            "outstanding deposit suffix exceeds protocol capacity"
+        );
+        eyre::ensure!(
+            prepared.enabled_tokens.len()
+                <= zone_primitives::constants::MAX_UNPROCESSED_TOKEN_ENABLEMENTS,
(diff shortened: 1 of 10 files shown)


shekhirin:
cyclops audit fast

shekhirin:
cyclops audit fast

crates/payload/src/builder.rs
```diff
@@ -172,19 +172,45 @@ where
...
+            TempoImport::CheckpointOnly(_) => false,
+        };
+        let total_deposits = match tempo_import {
+            TempoImport::Full(prepared) => prepared.queued_deposits.len(),
+            TempoImport::CheckpointOnly(_) => 0,
+        };
+        let enabled_tokens = match tempo_import {
+            TempoImport::Full(prepared) => prepared.enabled_tokens.len(),
+            TempoImport::CheckpointOnly(_) => 0,
+        };
```
rkrasiuk: nit: could be methods on `TempoImport`

crates/payload/src/builder.rs
```diff
@@ -235,51 +261,60 @@ where
         })?;
 
         // Execute advanceTempo system transaction — exactly one per zone block.
+        let opening_tx = match tempo_import {
+            TempoImport::Full(prepared) => build_advance_tempo_tx(prepared, chain_id),
+            TempoImport::CheckpointOnly(headers) => {
+                build_advance_tempo_headers_tx(headers, chain_id)?
+            }
+        };
```
rkrasiuk, line 260: same

crates/l1/src/block.rs
```diff
@@ -166,4 +199,8 @@ pub struct PreparedL1Block {
     /// Tokens newly enabled for bridging in this block.
     #[serde(skip)]
     pub enabled_tokens: Vec<abi::EnabledToken>,
+    /// Whether this is the first full import following checkpoint-only Zone blocks.
+    /// Such a block closes the settlement batch containing that prefix.
+    #[serde(skip)]
+    pub follows_checkpoint_prefix: bool,
```
rkrasiuk: nit: follows_checkpoint_blocks. prefix is weird here. not strongly opinionated tho

crates/l1/src/block.rs
```diff
@@ -166,4 +199,8 @@ pub struct PreparedL1Block {
     /// Tokens newly enabled for bridging in this block.
     #[serde(skip)]
     pub enabled_tokens: Vec<abi::EnabledToken>,
+    /// Whether this is the first full import following checkpoint-only Zone blocks.
+    /// Such a block closes the settlement batch containing that prefix.
```
adityapk00: maybe explicitly say ",and signals a batch boundary" ?

adityapk00 (approved):

rkrasiuk (approved):

crates/payload/src/builder.rs
shekhirin: this I'd rather keep as-is, don't want to move `build_*` functions outside of this file

PR #1245, opened 2026-08-21
feat(node): coordinate deferred Tempo imports

shekhirin:
- Queues finalized L1 blocks until they are imported by the Zone chain.
- Produces bounded checkpoint-only blocks while catching up after T12 activation, without crossing the activation boundary.
- Uses the finalized L1 target during catch-up.
- Uses Tempo T12 activation for checkpoint eligibility and Zone block timestamp behavior.
- Retires queued L1 blocks only after their Zone imports are confirmed.
- Replicates deferred imports consistently to follower nodes.
- Recovers cursors from both legacy and T12 `TempoAdvanced` events.
- Keeps checkpoint-only blocks leader-neutral so operational leadership changes only take effect in full blocks.

### DIFF
--- a/crates/node/src/engine.rs
+++ b/crates/node/src/engine.rs
@@ -50,20 +50,21 @@ use std::{
     sync::Arc,
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
+use tempo_chainspec::spec::TempoHardforks as _;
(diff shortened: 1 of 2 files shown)


shekhirin:
cyclops audit fast

PR #1255, opened 2026-08-24
refactor(checker): streamline startup and retries

legion2002:
Stacked on #1216.

## What changed

This keeps the checker behavior the same while avoiding repeated work:

- On a normal restart, reuse the accounting state already stored in the database. Rebuild the initial state only when creating or resetting the database.
- When an RPC or local state read briefly fails, retry that read in place instead of reloading the database and processing the whole notification again.
- Fetch the ZoneFactory creation logs a few pages at a time instead of waiting for every page one by one. The results are still sorted and checked in the same way.
- Use the L1 event list directly and share the existing helper for events that are decoded but intentionally ignored.

## Follow-up hardening

Review follow-ups added on top of the initial refactor:

- Bound transient RPC and local-state acquisition to 30 attempts. Exhaustion disables verification while the ExEx continues draining notifications.
- Reuse the persisted creation coordinate on restart, avoiding historical `ZoneCreated` discovery and initial-state replay unless the database is created or reset.
- Consume concurrent `ZoneCreated` log pages incrementally and retain only unique creation coordinates.
- Preserve the authenticated L1 observation reuse introduced by #1257.

Validation now passes 57 checker tests.

## Validation

- `cargo test -p zone-checker` (57 passed)
- `cargo clippy -p zone-checker --all-targets`
- `git diff --check`

### DIFF
--- a/crates/checker/README.md
+++ b/crates/checker/README.md
@@ -96,15 +96,22 @@ height.
 ## Failure behavior
 
 Temporary Tempo RPC or local-state acquisition failures retry without advancing
-or acknowledging the block. A deterministic mismatch records one durable
-finding, freezes the verified tip, and continues acknowledging subsequent
-notifications while recording how far the unchecked range extends. A finding
-remains active until the checker is rebuilt from authenticated genesis.
+or acknowledging the block. Each retry budget is bounded. Tempo retries use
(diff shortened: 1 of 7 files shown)


grandizzy:
cyclops audit fast

PR #1256, opened 2026-08-24
refactor(checker): make metadata encoding fallible

legion2002:
## Summary

- store checker metadata as raw Reth `Bytes` and encode it fallibly before MDBX writes
- preserve the existing versioned bincode representation and strict decoding behavior
- keep schema-version and metadata rows behind type-specific helpers
- cover oversized divergence findings without a panic or partial write

## Why

The previous `MetaValue` database codec relied on callers validating dynamic metadata before Reth's infallible `Compress` implementation called `expect`. This moves the fallible boundary ahead of `put`, so serialization failures become ordinary persistence errors.

## Stack

Depends on #1216.

## Testing

- `cargo test -p zone-checker --no-fail-fast` (57 passed)
- `cargo clippy -p zone-checker --all-targets`

Strict clippy with `-D warnings` remains blocked by pre-existing dead-code warnings in `zone-precompiles`.

### DIFF
--- a/crates/checker/src/persistence/codec.rs
+++ b/crates/checker/src/persistence/codec.rs
@@ -14,21 +14,13 @@ fn options() -> impl bincode::Options {
         .with_limit(MAX_VALUE_SIZE - 1)
 }
 
-/// Validate a dynamic value before passing it to MDBX's infallible codec interface.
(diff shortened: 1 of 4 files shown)


legion2002:
doesnt' seem worth it

PR #1257, opened 2026-08-24
refactor(checker): reuse authenticated L1 observations

legion2002:
Stacked on #1216.

## Summary
- retain receipt-root-authenticated Portal logs in the node L1 tracker for asynchronous checker consumption
- have live checker verification reuse those exact block/hash/parent-bound observations instead of refetching Tempo receipts
- keep the archival RPC fallback for bootstrap and recovery beyond the bounded recent cache
- reuse the node L1 receipt-root/log-bloom verifier for fallback and bootstrap receipt authentication

## Behavior
Checker accounting, strict Portal decoding, parent/hash validation, persistence, findings, and observe-only behavior remain unchanged. The tracker retains the latest 256 consumed Tempo observations; older recovery ranges fall back to the existing archival provider.

## Validation
- `cargo test -p zone-l1 -p zone-checker` (132 passed)
- `cargo test -p zone-node --lib --features cli checker_mode` (3 passed)
- `cargo clippy -p zone-l1 -p zone-checker -p zone-node --features zone-node/cli --all-targets -- -D warnings`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13757,6 +13757,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+ "zone-l1",
  "zone-precompiles",
  "zone-primitives",
 ]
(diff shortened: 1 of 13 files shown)


grandizzy (approved):
makes sense!

grandizzy:
cyclops audit fast

PR #1259, opened 2026-08-24
fix(seq): Filter out failed receipts

adityapk00:
Update the monitor-side withdrawal reconstruction logic in fetch_slot_withdrawals to verify the transaction receipt status of each fetched WithdrawalRequested log.

  - fetch_finalized_batch ignores failed receipts.
  - fetch_slot_withdrawals is protected through its filtered boundary scan and fetch_finalized_batch call.
  - Added receipt-status filtering to:
      - withdrawal detection  executor.rs:239
      - zone snapshot reconstruction in  settlement.rs:1528
      - finalized-boundary scanning in  settlement.rs:1713

  - Added regression tests covering reverted withdrawal, TempoAdvanced, and BatchFinalized logs.

### DIFF
--- a/crates/evm/src/executor.rs
+++ b/crates/evm/src/executor.rs
@@ -279,15 +279,27 @@ mod tests {
     };
 
     use alloy_consensus::{Header, Signed, TxLegacy};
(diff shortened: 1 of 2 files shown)


adityapk00:
cyclops audit fast

PR #1261, opened 2026-08-24
tests(node): Chaos tests for L1 disconnect/catchup and P2P network

adityapk00:
Adds end-to-end network failure tests for multi-sequencer using a controllable TCP proxy between Zone nodes and L1.

  This tests reconnection and backfill behavior.

  ## What this covers

  - All three sequencers losing their L1 connections, reconnecting, backfilling missed L1 blocks, and converging
  - Asymmetric outages affecting:
      - Both followers
      - The leader
      - A single follower

  - Continued block production and 2-of-3 settlement when a healthy quorum remains
  - Deposits and withdrawals submitted during outages being processed after recovery
  - Batch submission resuming after connectivity is restored

  ## Supporting changes

  - Adds a TCP chaos proxy capable of disconnecting active connections and restoring the same endpoint
  - Adds cluster fixtures supporting shared and per-node L1 proxies

### DIFF
--- a/crates/node/tests/it/network_chaos_e2e.rs
+++ b/crates/node/tests/it/network_chaos_e2e.rs
@@ -182,7 +182,7 @@ async fn test_three_sequencers_reconnect_and_catch_up_after_l1_network_outage() 
         .wait_for_connections_after(0, 3, NETWORK_TIMEOUT)
         .await?;
     let portal = ZonePortal::new(cluster.portal_address, cluster.l1.provider());
-    let batches_before_outage = poll_until(
+    poll_until(
         NETWORK_TIMEOUT,
         POLL_INTERVAL,
         "an initial settled batch before the network outage",
@@ -232,7 +232,6 @@ async fn test_three_sequencers_reconnect_and_catch_up_after_l1_network_outage() 
             "node {index} reached outage target {outage_target} without an L1 connection"
         );
     }
-    let batches_at_resume = batch_count(&portal).await?.max(batches_before_outage);
 
     // Restoring the listeners must result in new TCP connections, not reuse of the sockets that
     // were alive before the fault. Catch-up then proves those connections carry valid RPC traffic.
@@ -267,8 +266,8 @@ async fn test_three_sequencers_reconnect_and_catch_up_after_l1_network_outage() 
         || {
             let portal = &portal;
             async move {
-                let count = batch_count(portal).await?;
-                Ok((count > batches_at_resume).then_some(count))
+                let settled_height = portal.zoneHeight().call().await?;
+                Ok((settled_height >= U256::from(recovered_height)).then_some(settled_height))
             }
         },
     )
@@ -372,8 +371,6 @@ async fn test_asymmetric_l1_outages_recover_followers_then_leader() -> eyre::Res
             index + 1
         );
     }
-    let batches_at_resume = batch_count(&portal).await?;
-
     for proxy in [&follower_one_proxy, &follower_two_proxy] {
         proxy.resume();
     }
@@ -408,8 +405,8 @@ async fn test_asymmetric_l1_outages_recover_followers_then_leader() -> eyre::Res
         || {
             let portal = &portal;
             async move {
-                let count = batch_count(portal).await?;
-                Ok((count > batches_at_resume).then_some(count))
+                let settled_height = portal.zoneHeight().call().await?;
+                Ok((settled_height >= U256::from(recovered_height)).then_some(settled_height))
             }
         },
     )
@@ -466,8 +463,6 @@ async fn test_asymmetric_l1_outages_recover_followers_then_leader() -> eyre::Res
             "node {index} advanced zone state without a connected leader"
         );
     }
-    let batches_before_leader_resume = batch_count(&portal).await?;
-
     leader_proxy.resume();
     leader_proxy
         .wait_for_connections_after(leader_accepted_before_outage, 1, NETWORK_TIMEOUT)
@@ -496,8 +491,9 @@ async fn test_asymmetric_l1_outages_recover_followers_then_leader() -> eyre::Res
         || {
             let portal = &portal;
             async move {
-                let count = batch_count(portal).await?;
-                Ok((count > batches_before_leader_resume).then_some(count))
+                let settled_height = portal.zoneHeight().call().await?;
+                Ok((settled_height >= U256::from(phase_two_recovered_height))
+                    .then_some(settled_height))
             }
         },
     )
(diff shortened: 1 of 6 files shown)


adityapk00:
cyclops audit super-fast

crates/node/tests/it/network_chaos_p2p_e2e.rs
```diff
@@ -0,0 +1,228 @@
...
+async fn test_follower_keeps_up_with_high_l1_latency() -> eyre::Result<()> {
+    reth_tracing::init_test_tracing();
+
+    let (cluster, _network, [_leader_l1, follower_l1, _other_follower_l1]) =
+        start_cluster().await?;
+    synchronized_head(&cluster).await?;
+    follower_l1.set_client_to_upstream_latency(L1_LATENCY);
+    follower_l1.set_upstream_to_client_latency(L1_LATENCY);
+
+    let target = cluster.nodes[0].provider().get_block_number().await? + RECOVERY_GAP;
```
0xKitsune: This only checks that the follower eventually catches up but it looks like it doesn't verify that the follower waits to see the L1 anchors before importing the blocks.

crates/node/tests/it/network_chaos_e2e.rs
```diff
@@ -0,0 +1,581 @@
...
+        .await?;
+    eyre::ensure!(
+        cluster.nodes[1].tempo_block_number().await? < single_follower_outage_target,
+        "isolated follower reached outage target without an L1 connection"
+    );
+    eyre::ensure!(
+        leader_proxy.active_connections() > 0 && follower_two_proxy.active_connections() > 0,
+        "healthy quorum member lost L1 connectivity during the single-follower outage"
+    );
+    poll_until(
```
0xKitsune, line 540: This count could increase from a batch that was already in flight before the follower disconnected so it looks like this could pass without proving the other two nodes settled a new batch during the outage.

crates/node/tests/it/utils/network.rs
```diff
@@ -0,0 +1,529 @@
...
+    cancel: CancellationToken,
+    bytes: &AtomicU64,
+) -> std::io::Result<()>
+where
+    R: AsyncRead + Unpin,
+    W: AsyncWrite + Unpin,
+{
+    let mut buffer = vec![0_u8; 16 * 1024];
+    loop {
+        while conditions.borrow().paused {
```
0xKitsune, line 400: If pause is enabled while this read is already waiting it looks like the next chunk can still be forwarded because the paused state is not checked again before the write.

crates/node/tests/it/network_chaos_p2p_e2e.rs
adityapk00: fixed in https://github.com/tempoxyz/zones/pull/1261/commits/e1ceb63142c13e1c325d526b05a91d19426e8c5d

crates/node/tests/it/network_chaos_e2e.rs
adityapk00: fixed in https://github.com/tempoxyz/zones/commit/e1ceb63142c13e1c325d526b05a91d19426e8c5d

crates/node/tests/it/utils/network.rs
adityapk00: fixed in https://github.com/tempoxyz/zones/commit/e1ceb63142c13e1c325d526b05a91d19426e8c5d

crates/node/tests/it/network_chaos_e2e.rs
```diff
@@ -0,0 +1,576 @@
...
+        NETWORK_TIMEOUT,
+        POLL_INTERVAL,
+        "2-of-3 settlement to continue with one follower disconnected from L1",
+        || {
+            let portal = &portal;
+            async move {
+                let settled_height = portal.zoneHeight().call().await?;
+                Ok((settled_height > U256::from(phase_three_baseline)).then_some(settled_height))
```
0xKitsune, line 548: Consider asserting that the portal settled the leader height captured after single_follower_outage_target. Checking that zoneHeight moved past phase_three_baseline can still pass on a batch from before the follower fell behind.

crates/node/tests/it/network_chaos_e2e.rs
```diff
@@ -0,0 +1,576 @@
...
+
+    Ok(())
+}
+
+/// Exercise three asymmetric L1 outages: both followers, the leader, then one follower. Each
+/// restored side must backfill its missed anchors, process a deposit and withdrawal submitted
+/// during the outage, and converge before the next phase. The final phase also proves that one
+/// healthy follower preserves 2-of-3 settlement.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn test_asymmetric_l1_outages_recover_followers_then_leader() -> eyre::Result<()> {
```
0xKitsune, line 284: nit: consider extracting the three outage phases into small helpers while keeping the shared cluster setup. This test is pretty long and the phase helpers would make each scenario easier to follow.

crates/node/tests/it/network_chaos_e2e.rs
```diff
@@ -0,0 +1,576 @@
...
+    eyre::ensure!(
+        recovered_height > baseline_height,
+        "nodes did not advance beyond their pre-outage zone height"
+    );
+    cluster
+        .wait_all_at(recovered_height, NETWORK_TIMEOUT)
+        .await?;
+    cluster.assert_same_block(recovered_height).await?;
+
+    poll_until(
```
0xKitsune, line 262: nit: consider extracting a wait_for_settled_height helper since this polling block is repeated for each outage phase.

crates/node/tests/it/utils/network.rs
```diff
@@ -0,0 +1,563 @@
...
+            _ = cancel.cancelled() => return Ok(()),
+            result = reader.read(&mut buffer) => result?,
+        };
+        if read == 0 {
+            return writer.shutdown().await;
+        }
+
+        // `paused` may change while `read` is pending. Recheck it before forwarding the bytes so
+        // enabling a pause cannot leak the chunk that wakes the read.
+        while conditions.borrow().paused {
```
0xKitsune, line 416: nit: consider extracting a wait_until_resumed helper since the same pause loop is repeated three times.

0xKitsune (approved):

0xKitsune:
Approved with a few comments and mostly nits. I wonder if there is any way to make this a bit more succinct but the tests in general look to make sense.

adityapk00:
Fixed all comments in #1278

PR #1262, opened 2026-08-24
ci: verify reproducible candidate image

Rjected:
Adds a dispatch-only reproducible candidate verification mode to Docker Build, using the shared workflow from https://github.com/tempoxyz/gh-actions/pull/188.

It builds and pushes a `profile=reproducible` candidate through Depot, then compares its extracted `tempo-zone` checksum with an independent rebuild of the same commit without Docker cache. Zones supplies the pinned build recipes; the shared workflow resolves inputs, publishes run-specific `tempo-zone-repro` candidates, extracts by digest, and uploads a seven-day comparison manifest.

Dispatch from `main` with `reproducible_verify=true` and an optional source `ref`. Verification covers the candidate binary only. See `docs/REPRODUCIBLE_BUILDS.md` for dispatch and result-inspection instructions.


.github/workflows/docker.yml
```diff
@@ -37,13 +46,291 @@ env:
...
+          CANDIDATE_TAG: run-${{ github.run_id }}-${{ needs.reproducible-verify-resolve.outputs.short_sha }}
+        run: |
+          set -euo pipefail
+          echo "repository=$CANDIDATE_REPOSITORY" >> "$GITHUB_OUTPUT"
+          echo "tag=$CANDIDATE_TAG" >> "$GITHUB_OUTPUT"
+
+      - name: Build and push reproducible candidate image
+        uses: depot/bake-action@1d58c2668346981089b088b7ef36755b206b20e9 # v1.13.0
+        with:
+          files: docker/docker-bake.hcl
```
grandizzy: both builds use this commit’s Dockerfile.reproducible, Bake configuration, and build script. That verifies self-consistency, but a malicious commit could modify both build paths, should the verifier use build definitions pinned to a trusted SHA maybe? (or maintained externally)

.github/workflows/docker.yml
```diff
@@ -37,13 +46,291 @@ env:
...
+              docker rm -f "$container_id" >/dev/null 2>&1 || true
+            fi
+            write_manifest
+            exit "$status"
+          }
+          trap cleanup EXIT
+
+          tagged_image="$CANDIDATE_REPOSITORY:$CANDIDATE_TAG"
+          for attempt in {1..12}; do
+            image_digest="$(docker buildx imagetools inspect "$tagged_image" \
```
grandizzy: ideally, we should retrieve the image digest from Depot and use it directly here. Should avoid the small window where the tag could be overwritten and verification could inspect a different image from the one created by this run.

grandizzy (commented):
ty! overall lgtm, left 2 comments

.github/workflows/docker.yml
Rjected: I think a trusted SHA makes sense for now, although we should probably set something up that is less ad-hoc. Maintaining externally seems better long-term, especially if reusable. Let's pin a SHA from main once this PR is merged, and then figure out how best to do this long term?

grandizzy:
cyclops audit

.github/workflows/docker.yml
```diff
@@ -37,13 +46,313 @@ env:
...
+    outputs:
+      commit_sha: ${{ steps.commit.outputs.commit_sha }}
+      short_sha: ${{ steps.commit.outputs.short_sha }}
+      source_date_epoch: ${{ steps.commit.outputs.source_date_epoch }}
+      version: ${{ steps.commit.outputs.version }}
+    steps:
+      - name: Check out requested ref
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ inputs.ref || github.sha }}
```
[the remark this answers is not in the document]
Rjected: same concern as https://github.com/tempoxyz/zones/pull/1262#discussion_r3850784878 will fix after merge

docker/Dockerfile.reproducible
```diff
@@ -47,3 +47,17 @@ RUN cargo build --locked --bin tempo-zone \
...
+FROM debian:bookworm-slim@sha256:4724b8cc51e33e398f0e2e15e18d5ec2851ff0c2280647e1310bc1642182655d AS reproducible-runtime-base
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /data
+
+FROM reproducible-runtime-base AS tempo-zone-reproducible
+COPY --from=builder /app/target/x86_64-unknown-linux-gnu/reproducible/tempo-zone /usr/local/bin/tempo-zone
+ENTRYPOINT ["/usr/local/bin/tempo-zone"]
```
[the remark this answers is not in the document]
Rjected: fixed in #1264

grandizzy (dismissed):
lgtm, thanks! would like to have one more review before merging, @sds or @0xalpharush ptal

grandizzy:
also think could be now be part of tempoxyz/gh-actions repo?

PR #1264, opened 2026-08-24
ci: ship reproducible zone binaries

Rjected:
Promotes the Debian `profile=reproducible` recipe to the canonical Linux x86_64 `tempo-zone` contract.

The normal image and Linux x86_64 archive now carry the same binary. Main and tag image builds independently rebuild the resolved commit, extract the binary from the pushed image by immutable digest, and publish verification manifests; draft creation waits for the equivalent raw-binary comparison.

ARM Linux and macOS archives continue to use the maxperf path and are not covered by this contract. The manual candidate-image verifier remains separate.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -43,6 +43,7 @@ permissions:
 
 env:
   REGISTRY: ghcr.io/tempoxyz
+  STAGING_REPOSITORY: ghcr.io/tempoxyz/tempo-zone-staging
   TEMPO_DEVNET_GENESIS_URL: ${{ inputs.tempo_genesis_url || 'https://devnet-assets.tempoxyz.dev/tempo-devnet-nextfork.json' }}
 
 jobs:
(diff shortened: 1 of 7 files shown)


docker/docker-bake.hcl
```diff
@@ -65,13 +65,20 @@ target "_common" {
 }
 
 target "tempo-zone" {
-  inherits = ["_common", "docker-metadata"]
-  target = "tempo-zone"
+  inherits = ["docker-metadata"]
+  dockerfile = "docker/Dockerfile.reproducible"
```
grandizzy, line 69: this builds with reproducible which aborts on panic https://github.com/tempoxyz/zones/blob/3f62972b263eac467ed6dffe4c30986ac3d79832/Cargo.toml#L98
I think worth double checking if ok or if we should unwind

.github/workflows/docker.yml
grandizzy, line 620: for main and tag runs, this publishes the shared Docker tags before the new reproducibility verification job completes, should we stage the image and promote it after verification maybe?

grandizzy (commented):
ty! left couple of comments, please check

docker/docker-bake.hcl
Rjected: ah I think we should make this unwind, more readable

.github/workflows/docker.yml
Rjected: makes sense, should only publish the tag until after verification succeeds

.github/workflows/release.yml
```diff
@@ -31,158 +23,384 @@ permissions:
...
+      commit_sha: ${{ steps.release.outputs.commit_sha }}
+      short_sha: ${{ steps.release.outputs.short_sha }}
+      source_date_epoch: ${{ steps.release.outputs.source_date_epoch }}
     steps:
-      - name: Get version from tag
-        id: get_version
+      - name: Check out release tag or ref
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: ${{ inputs.tag || github.ref }}
```
grandizzy: I think we should restrict this for tags only as it used to be, and drop ref

grandizzy (commented):
looks good, left one comment, think was added for testing?

grandizzy:
cyclops audit

PR #1265, opened 2026-08-24
refactor(node): extract P2P startup logic

0xKitsune:
This PR extracts P2P startup into a dedicated function.

PR #1266, opened 2026-08-24
refactor: extract P2P startup

0xKitsune:
This PR extracts P2P startup into a dedicated function.


0xKitsune:
Folded into #1265.

PR #1268, opened 2026-08-24
refactor(node): prepare L1 state

0xKitsune:
This PR refactors Zone L1 state seeding to a dedicated function that initializes enabled tokens and encryption keys.


0xKitsune:
Closing: this extraction only wrapped two existing calls and does not create a useful boundary.

PR #1273, opened 2026-08-25
fix(prover-utils): handle zero portal hash for first batch

rkrasiuk:
## Summary

 Treat the portal's zero block hash as the pre-genesis sentinel and use the canonical Zone genesis header as the parent of block 1.


shekhirin (approved):

shekhirin:
+1

PR #1284, opened 2026-08-25
test(p2p): matrix leadership handoff faults

legion2002:
## What

- replace duplicated leadership-fault setup and assertions with a declarative scenario definition and shared runner
- keep every original scenario as an independently runnable test and add two distinct cases: B+C unavailable before handoff, and outgoing A isolated before it consumes the activation boundary
- deterministically hold A below the recorded activation block, prove B waits for A boundary tip, then verify recovery and settlement under B
- keep the withdrawal/settlement-boundary test separate because it exercises a different invariant

## Why

This keeps coverage focused on behaviorally different failures while making another case a small declaration instead of another copied end-to-end test. Individual test names preserve CI sharding, targeted retries, and clear failures.

## Validation

- `git diff --check`
- `cargo test -p zone-node --test it --no-run`
- `cargo clippy -p zone-node --test it -- -D warnings`
- all 6 leadership-fault scenarios passed as independently runnable tests
- deterministic pre-activation outgoing-leader case passed focused stress runs
- `cargo test -p zone-node --test it test_handoff_recovers_across_settlement_boundary -- --nocapture`

### DIFF
--- a/crates/node/tests/it/network_chaos_p2p_e2e.rs
+++ b/crates/node/tests/it/network_chaos_p2p_e2e.rs
@@ -24,10 +24,46 @@ const RECOVERY_GAP: u64 = 6;
 const INITIAL_DEPOSIT: u128 = 10_000_000;
 const WITHDRAWAL_AMOUNT: u128 = 250_000;
 
-#[derive(Clone, Copy)]
-struct LeadershipDisconnect {
-    l1: bool,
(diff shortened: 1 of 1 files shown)


legion2002:
cyclops audit super-fast

adityapk00 (approved):

PR #1286, opened 2026-08-26
fix(evm): support L1-backed Zone re-execution

kuyziss:
## Summary
- let Zone CLI utility commands use `L1_HTTP_RPC_URL` for L1 storage cache misses
- preserve the offline fallback when the variable is unset
- warn and fall back on an invalid configured URL

## Validation
- `cargo +1.95.0 check -p zone-evm`
- `cargo +1.95.0 fmt --all -- --check`

Required by the dedicated Zone re-execute workflow: Zone blocks can contain authenticated L1 storage reads, and the generic Reth re-execute component otherwise intentionally uses the offline `127.0.0.1:1` fallback.

### DIFF
--- a/.github/workflows/zones-benchmark.yml
+++ b/.github/workflows/zones-benchmark.yml
@@ -85,7 +85,7 @@ on:
         description: L1 callback gas requested by composable withdrawals (maximum 10000000).
         type: string
         required: true
-        default: "10000000"
+        default: "5000000"
       swap-liquidity:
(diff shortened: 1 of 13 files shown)


klkvr:
cyclops audit fast

klkvr (approved):

klkvr (approved):

klkvr (approved):

PR #1302, opened 2026-08-26
fix(node): pace zone block timestamps

decofe:
## Summary

- wait until the next strictly monotonic Zone timestamp is no longer ahead of wall-clock time
- preserve full-speed steady-state production while limiting rapid catch-up to one block per millisecond
- add a focused test covering future-timestamp pacing

## Tests

- `cargo test -p zone-node engine::tests --lib`
- `cargo +nightly clippy -p zone-node --lib --tests`

Prompted by: @adityapk00

### DIFF
--- a/crates/node/src/engine.rs
+++ b/crates/node/src/engine.rs
@@ -446,31 +446,44 @@ fn zone_timestamp_millis(
         .max(parent_timestamp_millis.saturating_add(1))
 }
 
-/// Wait until the next valid Zone timestamp is no longer ahead of wall-clock time.
+/// Wait until the next valid Zone timestamp is not in the future, then return it.
 async fn paced_zone_timestamp_millis(
     l1_timestamp_millis: u64,
     parent_timestamp_millis: u64,
 ) -> eyre::Result<u64> {
-    loop {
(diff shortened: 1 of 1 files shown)


adityapk00:
cyclops audit fast

crates/node/src/engine.rs
```diff
@@ -450,6 +446,33 @@ fn zone_timestamp_millis(
...
+            l1_timestamp_millis,
+            parent_timestamp_millis,
+            wall_clock_timestamp_millis,
+        );
+
+        if timestamp_millis == wall_clock_timestamp_millis {
+            return Ok(timestamp_millis);
+        }
+
+        tokio::time::sleep(Duration::from_millis(
```
0xKitsune, line 469: IIUC, if the L1 timestamp is in the future or the clock moves backward, this sleep could run past the 40s generation shutdown timeout. 

Since the wait does not observe cancellation, this would then time out and would affect the role controller. We have not sent FCU yet, so can we select on the stop token here and recheck the production permit after waking?

adityapk00: I pushed a fix above which caps the wait at 1 sec.

klkvr (approved):

PR #1303, opened 2026-08-26
fix(prover): harden Nitro enclave bootstrap

0xalpharush:
## Summary

Stacked on #1301.

- build the kernel, resolved config, cmdline, NSM driver, init, and linuxkit from the same pinned AWS Nitro bootstrap source
- pin the Nix builder image by digest and publish SHA-384 hashes for the matched bootstrap artifacts
- apply a focused patch to AWS's pinned kernel expression rather than vendoring it
- apply a small Zones Kconfig fragment over AWS's x86_64 baseline and assert the resolved values after olddefconfig
- enable KASLR, kernel-memory ASLR, hardened usercopy, slab-freelist randomization, and dmesg restrictions
- remove unused enclave interfaces: debugfs, Magic SysRq, kexec, io_uring, BPF/JIT, userfaultfd, user/network namespaces, 32-bit compatibility, modify_ldt, /dev/mem, and /proc/kcore

These are actual changes from AWS's baseline rather than restatements of its defaults. CONFIG_EXPERT is enabled only so hidden default-on facilities can be disabled; it does not add a runtime interface.

## Validation

- [x] resolved Kconfig assertions pass
- [x] native CI compiles the patched Linux 6.6.79 kernel and matching NSM module with Nix's syscall filter enabled
- [x] pinned init and linuxkit builds/tests exercised
- [x] Docker Bake graph parses
- [x] git diff --check
- [x] existing Docker Build workflow completes the full EIF build
- [ ] boot on real Nitro hardware
- [ ] successful prover request on real Nitro hardware

Because this is a compatibility-sensitive reduction of AWS's baseline, the PR must pass a full EIF build, real-Nitro boot, and prover request before merge.

## Deferred follow-ups

- PERF_EVENTS and kprobes remain selected by other AWS baseline functionality. Disabling them should be a focused dependency review rather than expanding this PR's scope.
- INIT_ON_ALLOC_DEFAULT_ON and PAGE_TABLE_CHECK_ENFORCED should be reconsidered only with same-hardware prover memory and throughput benchmarks. INIT_ON_FREE_DEFAULT_ON is excluded because its runtime cost is disproportionate for this memory-heavy prover.
- AWS's cmdline uses nomodules, while Linux documents nomodule. Changing it directly would prevent the pinned init from loading nsm.ko; NSM should first be built in or init should disable further module loading after startup.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -195,6 +195,14 @@ jobs:
 
           docker run --rm \
             --platform linux/amd64 \
+            --entrypoint /bin/cp \
+            --volume "$PWD/$EIF_OUTPUT_DIR:/output" \
+            "$EIF_BUILDER_IMAGE" \
+            /usr/share/nitro_enclaves/components.sha384 \
+            /output/components.sha384
+
+          docker run --rm \
+            --platform linux/amd64 \
             --volume /var/run/docker.sock:/var/run/docker.sock \
             --volume "$PWD/$EIF_OUTPUT_DIR:/output" \
             "$EIF_BUILDER_IMAGE" \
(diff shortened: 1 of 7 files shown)


0xalpharush:
cyclops audit fast

decofe:
Closing per author cleanup. The Nitro kernel/bootstrap refresh landed in #1301, while reproducible image work is tracked in #1262 and #1264. No single PR carries this PR's full hardening delta.

PR #1304, opened 2026-08-27
refactor(dev): launch dev node directly

0xKitsune:
This PR gives `tempo-zone dev` its own Zone launch path instead of re-entering normal node startup through `--sequencer`. Normal `tempo-zone node` behavior is unchanged.


0xKitsune:
Closing to replace this draft with a clean PR and unpolluted history.

PR #1306, opened 2026-08-27
refactor(node): model chain runtime explicitly

0xKitsune:
This PR models the node runtime as one explicit configuration instead of three independent optional fields. Existing builder methods remain supported.


0xKitsune:
Closing because this runtime abstraction was based on the earlier standalone-versus-networked model and is no longer part of the intended stack.

PR #1312, opened 2026-08-27
feat(node): build checkpoint-only blocks and coordinate deferred Tempo imports

shekhirin:
Continuously catches the Zone up to finalized Tempo by producing bounded checkpoint-only blocks, then carrying deferred portal work into the next full import. Keeps this state consistent across T12 activation, follower replication, and restarts. Verifies the checkpoint blocks leader according to https://github.com/tempoxyz/tempo/pull/7366.

### DIFF
--- a/crates/l1/src/queue.rs
+++ b/crates/l1/src/queue.rs
@@ -10,6 +10,8 @@ use std::collections::VecDeque;
 pub(crate) struct PendingDeposits {
     /// Pending L1 blocks with their portal events, not yet processed by the Zone.
     pending: VecDeque<L1BlockDeposits>,
+    /// Portal work crossed by checkpoint-only Zone blocks and owed by the next full block.
+    deferred: Vec<L1BlockDeposits>,
     /// Highest L1 block ever enqueued (number + hash). Survives `confirm` /
     /// `drain` so that reconnecting subscribers know where the queue left off,
     /// even if the engine has already consumed the blocks.
@@ -83,6 +85,14 @@ impl PendingDeposits {
         self.pending.front()
     }
 
+    pub(crate) fn peek_headers(&self, maximum: usize) -> Vec<SealedHeader<TempoHeader>> {
+        self.pending
+            .iter()
+            .take(maximum)
+            .map(|block| block.header.clone())
+            .collect()
+    }
+
     /// Confirm the next pending L1 block was successfully processed and remove it.
     ///
     /// The caller must pass the [`NumHash`] returned by [`Self::peek`]. A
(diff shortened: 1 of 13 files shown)


shekhirin:
cyclops audit fast

shekhirin:
cyclops audit fast

crates/l1/src/subscriber.rs
```diff
@@ -880,6 +960,61 @@ impl L1Subscriber {
...
+        if self.deposit_queue.last_enqueued().is_some() {
+            return Ok(());
+        }
+        let checkpoint = self.local_state.latest_tempo_checkpoint()?;
+        if from > checkpoint.number {
+            return Ok(());
+        }
+
+        let mut deferred: Option<DeferredPortalWork> = None;
+        for block_number in from..=checkpoint.number {
```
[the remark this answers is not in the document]
shekhirin: 1. Unbounded — we do not collect all headers in a vec, instead we only collect missed deposits and token enablements
2. Sequential — can improve later
3. All-or-nothing — we need it for correctness
4. Blocks L1 ingestion — this happens on startup, and is expected

shekhirin:
cyclops audit fast

crates/node/src/replication.rs
```diff
@@ -1149,63 +1157,85 @@ where
...
+    for (header, observed) in headers
+        .iter()
+        .cloned()
+        .zip(observed_headers.iter().cloned())
+    {
+        deposit_queue
+            .try_enqueue_sealed(header, observed)
+            .wrap_err_with(|| format!("cannot queue an imported header of block {block_number}"))?;
+    }
+
```
klkvr: i've always found this a bit weird

it feels like it'd be cleaner if `wait_for_peer_anchor` success already guaranteed that deposits queue is caught up to the header we've awaited

that way we wouldn't have to queue again here and could just ask queue for whatever events this block range has vs fethcing all of them ourselves

shekhirin: I think it already does, and this loop is a sort of a sanity check, but yeah it's weird

in `backfill`, we first update the deposit queue, and only then emit a notification that `wait_for_peer_anchor` wakes up on https://github.com/tempoxyz/zones/blob/4e4f98e061d706497c750d4a31a077fb43db6704/crates/l1/src/subscriber.rs#L892-L910

crates/l1/src/queue.rs
```diff
@@ -225,12 +426,58 @@ impl DepositQueue {
...
+        self.inner.lock().confirm_operational(expected)
+    }
+
+    /// Idempotently reconcile a canonical full import and release the deferred work it consumed.
+    pub fn confirm_operational_through(&self, expected: NumHash) -> eyre::Result<()> {
+        self.inner.lock().confirm_operational_through(expected)
+    }
+
     /// Advance the queue past a canonical follower anchor.
     pub fn confirm_through(&self, expected: NumHash) -> eyre::Result<()> {
```
klkvr: looks like `confirm_through` is only used in tests now?

crates/node/src/replication.rs
klkvr: yeah so maybe we could refactor this here right away? given that it causes extra complexity

crates/l1/src/queue.rs
```diff
@@ -1,6 +1,92 @@
...
+            .event_blocks
+            .iter()
+            .map(|block| block.events.deposits.len())
+            .sum();
+        self.enabled_token_count = self
+            .event_blocks
+            .iter()
+            .map(|block| block.events.enabled_tokens.len())
+            .sum();
+    }
```
klkvr: this looks a bit odd? iiuc we're always consuming the entire deferred range so partially draining deferred work shouldn't really be possible?

crates/l1/src/subscriber.rs
```diff
@@ -880,6 +960,61 @@ impl L1Subscriber {
         self.follow_finalized(&provider, triggers).await
     }
 
+    async fn recover_deferred_work(
```
klkvr, line 975: not sure if there's a way but would be nice to unify this with backfill somehow

crates/node/src/engine.rs
```diff
@@ -421,24 +430,147 @@ impl ZoneEngine {
...
+        .len()
+        .saturating_sub(usize::from(reserve_for_full))
+        .min(zone_primitives::constants::MAX_TEMPO_HEADERS_PER_ZONE_BLOCK)
+        // Never reserve a queue front from an older hardfork for a full import under newer Zone
+        // rules. This also closes the small race where the hardfork successor is appended between
+        // the bounded queue snapshot and tip read.
+        .max(usize::from(first_l1_hardfork < zone_hardfork));
+    if checkpoint_count == 0 {
+        if target_is_visible && finalized_target.is_some_and(|target| !target.ready) {
+            TempoImportDecision::WaitForFinalizedTarget
```
klkvr, line 567: smth to note is that right now if we're waiting for target to be ready we don't actually get notified when it becomes ready here https://github.com/tempoxyz/zones/blob/4e4f98e061d706497c750d4a31a077fb43db6704/crates/node/src/engine.rs#L247-L265

this still gets woken up by fcu interval tho so not critical

crates/node/src/replication.rs
```diff
@@ -1114,13 +1115,20 @@ where
...
         let existing = provider.sealed_header(block_number)?.ok_or_else(|| {
             eyre::eyre!("missing local canonical header at height {block_number}")
         })?;
         if existing.hash() == hash {
+            // This path bypasses `new_payload`, so verify the peer-supplied body against the
+            // canonical header before deriving queue mutations from it.
+            block.ensure_transaction_root_valid()?;
+            reconcile_canonical_import(deposit_queue, &tempo_import).wrap_err_with(|| {
+                format!("cannot reconcile duplicate canonical peer block {block_number}")
+            })?;
```
klkvr, line 1133: not sure i understand why this is relevant for the PR

klkvr (dismissed):

adityapk00 (dismissed):

crates/node/src/replication.rs
shekhirin: addresses this finding https://github.com/tempoxyz/zones/pull/1312#discussion_r3881726788

I guess this it not an issue if we don't expect peers to be malicious, can get rid of it to simplify

shekhirin:
cyclops audit fast

crates/l1/src/subscriber.rs
shekhirin: can do in a follow-up

adityapk00 (approved):

PR #1315, opened 2026-08-27
fix(spf): authenticate bootstrap batch boundaries

legion2002:
## Summary

- authenticate a block-1 witness against the exact genesis header in verifier-selected `SpfConfig`
- require the first imported Tempo state to prove that the configured Portal has an initialized sequencer set
- reject finalization in intermediate blocks and require it in the final block, including empty withdrawal batches
- make the witness generators include the bootstrap Portal proof and stop automatic ranges at the first batch boundary

## Compatibility

This reads and authenticates the existing Zone genesis and Portal state. It does not mutate or regenerate genesis and does not change ZonePortal ABI or storage. Ordinary batches after the first submitted batch retain their existing parent-hash flow.

## Stack

- 1/2: #1314
- 2/2: this PR

## Validation

- `cargo test -p zone-spf --lib` (25 passed)
- `cargo test -p zone-node --test it spf_ -- --nocapture` (3 passed)
- `cargo check -p zone-sequencer -p tempo-zone-prover-utils`
- `cargo clippy -p zone-spf -p zone-sequencer -p tempo-zone-prover-utils --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14169,6 +14169,7 @@ dependencies = [
  "serde",
  "tempo-chainspec",
  "tempo-evm",
+ "tempo-precompiles",
  "tempo-primitives",
  "tempo-revm",
  "tempo-zone-contracts",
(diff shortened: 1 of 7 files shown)


legion2002:
cyclops audit fast

PR #1324, opened 2026-08-28
feat: shadow prover support for RPC followers

rkrasiuk:
## Summary

Enables RPC-only follower nodes to run the detached shadow prover without requiring sequencer or settlement keys. The follower observes finalized submitBatch transactions on L1, decodes the accepted quorum-certificate inputs, derives the exact anchor used by settlement, and waits for the corresponding Zone block range to become canonical locally before enqueueing the proof.

### DIFF
--- a/crates/l1/src/subscriber.rs
+++ b/crates/l1/src/subscriber.rs
@@ -375,14 +375,6 @@ pub trait LeadershipSink: Send + Sync + std::fmt::Debug {
     fn apply_leader_transition(&self, transition: &crate::LeaderTransition) -> eyre::Result<()>;
 }
 
-/// Observes accepted batch submissions decoded from verified finalized receipts.
-///
-/// Unlike consensus-critical sinks, this is observational and cannot fence L1 ingestion.
-pub trait FinalizedBatchSubmissionSink: Send + Sync + std::fmt::Debug {
-    /// Observe one finalized Portal submission.
-    fn observe_finalized_batch(&self, submission: FinalizedBatchSubmission);
-}
-
 /// Configuration for the L1 subscriber.
 #[derive(Debug, Clone)]
 pub struct L1SubscriberConfig {
@@ -414,8 +406,9 @@ pub struct L1Subscriber<P> {
     pub(crate) block_tracker: L1BlockTracker,
     /// Optional sink for leadership transitions.
     pub(crate) leadership_sink: Option<Arc<dyn LeadershipSink>>,
-    /// Optional observational sink for finalized accepted batch submissions.
-    pub(crate) finalized_batch_submission_sink: Option<Arc<dyn FinalizedBatchSubmissionSink>>,
+    /// Optional observational channel for finalized accepted batch submissions.
+    pub(crate) finalized_batch_submissions:
+        Option<tokio::sync::mpsc::UnboundedSender<FinalizedBatchSubmission>>,
     /// Private encryption keys bound by finalized Portal rotation events.
     pub(crate) encryption_keys: Option<EncryptionKeyRing>,
     /// L1 subscriber metrics for connection health, backfill, and event ingestion.
@@ -467,7 +460,9 @@ where
         l1_state_cache: L1StateCache,
         block_tracker: L1BlockTracker,
         leadership_sink: Option<Arc<dyn LeadershipSink>>,
-        finalized_batch_submission_sink: Option<Arc<dyn FinalizedBatchSubmissionSink>>,
+        finalized_batch_submissions: Option<
+            tokio::sync::mpsc::UnboundedSender<FinalizedBatchSubmission>,
+        >,
         encryption_keys: Option<EncryptionKeyRing>,
     ) -> Self {
         Self {
@@ -478,7 +473,7 @@ where
             l1_state_cache,
             block_tracker,
             leadership_sink,
-            finalized_batch_submission_sink,
+            finalized_batch_submissions,
             encryption_keys,
             subscriber_metrics: Default::default(),
         }
@@ -789,9 +784,15 @@ where
                 self.block_tracker
                     .record_with_portal_events(anchor, events.clone())?;
             }
-            if let Some(sink) = &self.finalized_batch_submission_sink {
+            if let Some(sender) = &self.finalized_batch_submissions {
                 for submission in finalized_batches {
-                    sink.observe_finalized_batch(submission);
+                    if sender.send(submission).is_err() {
+                        warn!(
+                            target: "zone::l1::subscriber",
+                            "Finalized batch submission observer is unavailable"
+                        );
+                        break;
+                    }
                 }
             }
             // Publish derived L1 state only after the header has been admitted to every
@@ -891,26 +892,33 @@ where
                     if retain_receipt_logs && let Some(logs) = &mut portal_logs {
                         logs.push(log.inner.clone());
                     }
-                    if self.finalized_batch_submission_sink.is_some()
+                    if self.finalized_batch_submissions.is_some()
                         && log.topic0() == Some(&ZonePortal::BatchSubmitted::SIGNATURE_HASH)
                     {
-                        let event = ZonePortal::BatchSubmitted::decode_log(&log.inner)
-                            .wrap_err_with(|| {
-                                format!(
-                                    "failed to decode BatchSubmitted in L1 block {block_number}"
-                                )
-                            })?
-                            .data;
-                        finalized_batches.push(FinalizedBatchSubmission {
-                            block_number,
-                            transaction_hash: receipt.transaction_hash(),
-                            log_index: log.log_index.ok_or_else(|| {
-                                eyre::eyre!(
-                                    "BatchSubmitted in L1 block {block_number} has no log index"
-                                )
-                            })?,
-                            event,
-                        });
+                        match (
+                            ZonePortal::BatchSubmitted::decode_log(&log.inner),
+                            log.log_index,
+                        ) {
+                            (Ok(event), Some(log_index)) => {
+                                finalized_batches.push(FinalizedBatchSubmission {
+                                    block_number,
+                                    transaction_hash: receipt.transaction_hash(),
+                                    log_index,
+                                    event: event.data,
+                                });
+                            }
+                            (Err(err), _) => warn!(
+                                target: "zone::l1::subscriber",
+                                block_number,
+                                error = ?err,
+                                "Could not decode finalized batch submission for observer"
+                            ),
+                            (_, None) => warn!(
+                                target: "zone::l1::subscriber",
+                                block_number,
+                                "Finalized batch submission has no log index"
+                            ),
+                        }
                     }
                     invalidated.insert(address);
                     if let Some(address) =
--- a/crates/l1/src/subscriber.rs
+++ b/crates/l1/src/subscriber.rs
@@ -443,7 +443,7 @@ pub struct L1Subscriber<P> {
     pub(crate) leadership_sink: Option<Arc<dyn LeadershipSink>>,
     /// Optional observational channel for finalized accepted batch submissions.
     pub(crate) finalized_batch_submissions:
-        Option<tokio::sync::mpsc::UnboundedSender<FinalizedBatchSubmission>>,
+        Option<tokio::sync::mpsc::Sender<FinalizedBatchSubmission>>,
     /// Private encryption keys bound by finalized Portal rotation events.
     pub(crate) encryption_keys: Option<EncryptionKeyRing>,
     /// L1 subscriber metrics for connection health, backfill, and event ingestion.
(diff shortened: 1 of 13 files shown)


rkrasiuk:
cyclops audit fast

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    async fn fetch_submit_batch_calls(
+        &self,
+        tx_hash: B256,
+    ) -> Result<Vec<ZonePortal::submitBatchCall>> {
+        let tx = self
+            .l1_provider
+            .client()
+            .request::<_, serde_json::Value>("eth_getTransactionByHash", (tx_hash,))
```
shekhirin: can we use alloy provider method here?

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,386 @@
...
+        let this = self.clone();
+        tokio::spawn(async move {
+            loop {
+                if recovery_sender.is_closed() {
+                    return;
+                }
+                match this.recover_recent_submissions().await {
+                    Ok(recovered) => {
+                        for submission in recovered {
+                            if recovery_sender.send(submission).is_err() {
```
shekhirin: can we send to this channel directly from `recover_recent_submissions`, so that we don't accumulate them all in memory first

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,386 @@
...
+        anchor: ShadowProofAnchor,
+    ) -> Result<()> {
+        let from = loop {
+            if batch.prev_block_hash.is_zero() {
+                break 1;
+            }
+            if let Some(previous) = self.zone_provider.block_number(batch.prev_block_hash)? {
+                break previous.saturating_add(1);
+            }
+            tokio::time::sleep(POLL_INTERVAL).await;
```
shekhirin, line 268: do we expect this to be hit due to a race between new block processing and proving?

rkrasiuk: not exactly with proving, but rather the local state of the follower node. it might lag behind the quorum participating zone nodes

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,386 @@
...
+            }
+        });
+    }
+
+    async fn process_submission(&self, submission: &FinalizedBatchSubmission) -> Result<()> {
+        let call = self
+            .fetch_submit_batch_calls(submission.transaction_hash)
+            .await?
+            .into_iter()
+            .find(|call| call_matches_event(call, &submission.event))
```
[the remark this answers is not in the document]
shekhirin: not a concern, batch submission txs always have one `submitBatch` call

shekhirin (dismissed):

crates/l1/src/subscriber.rs
```diff
@@ -379,6 +394,9 @@ pub struct L1SubscriberConfig {
     /// Optional sink that receives leadership transitions before the block that
     /// recorded them is enqueued.
     pub leadership_sink: Option<Arc<dyn LeadershipSink>>,
+    /// Optional observational channel for finalized accepted batch submissions.
+    pub finalized_batch_submissions:
+        Option<tokio::sync::mpsc::UnboundedSender<FinalizedBatchSubmission>>,
```
0xKitsune: Not blocking but lets plan to move this to the `L1Subscriber` struct pending https://github.com/tempoxyz/zones/pull/1356 landing.

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        info!(
+            target: "zone::node::shadow_prover",
+            zone_from = from,
+            zone_to = to,
+            anchor_number = anchor.number,
+            anchor_hash = %anchor.hash,
+            "Queueing finalized quorum-certified batch for shadow proving"
+        );
+        self.prover
```
0xKitsune: It looks like a batch that does not end in `finalizeWithdrawalBatch` can still pass here because the SPF accepts open snapshots and copies the expected batch index.

crates/l1/src/subscriber.rs
```diff
@@ -903,6 +933,34 @@ impl L1Subscriber {
...
+                        && log.topic0() == Some(&ZonePortal::BatchSubmitted::SIGNATURE_HASH)
+                    {
+                        match (
+                            ZonePortal::BatchSubmitted::decode_log(&log.inner),
+                            log.log_index,
+                        ) {
+                            (Ok(event), Some(log_index)) => {
+                                finalized_batches.push(FinalizedBatchSubmission {
+                                    block_number,
+                                    transaction_hash: receipt.transaction_hash(),
```
0xKitsune: Does the fetched transaction needs to be authenticated against the finalized block since receipt root verification does not cover `transaction_hash` or `log_index` RPC metadata.

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,360 @@
...
+            }
+        });
+    }
+
+    async fn process_submission(&self, submission: &FinalizedBatchSubmission) -> Result<()> {
+        let call = self
+            .fetch_submit_batch_calls(submission.transaction_hash)
+            .await?
+            .into_iter()
+            .find(|call| call_matches_event(call, &submission.event))
```
0xKitsune, line 107: Maybe a nit but IIUC a multicall tx can map multiple logs to the first matching `submitBatch` call because call order and the log index are ignored.

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,360 @@
...
+        mut submissions: UnboundedReceiver<FinalizedBatchSubmission>,
+        recovery_sender: UnboundedSender<FinalizedBatchSubmission>,
+    ) {
+        self.spawn_recovery(recovery_sender);
+
+        let mut observed = HashSet::new();
+        while let Some(submission) = submissions.recv().await {
+            let observation_id = (submission.transaction_hash, submission.log_index);
+            if observed.insert(observation_id) {
+                self.spawn_submission_retry(submission);
```
0xKitsune: It looks like this can bypass the queue bound because every observation is retained forever and gets a task that can also wait forever.

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,360 @@
...
+        return Ok(tempo_block_number);
+    }
+    ensure!(
+        recent_tempo_block_number > tempo_block_number,
+        "submitted ancestry anchor does not follow its Tempo checkpoint"
+    );
+    Ok(recent_tempo_block_number)
+}
+
+#[cfg(test)]
```
0xKitsune, line 351: Looks like we could also use tests covering recovery, calldata binding and rejection of non boundary submissions.

crates/l1/src/subscriber.rs
```diff
@@ -828,6 +846,17 @@ impl L1Subscriber {
                     .block_tracker
                     .record_with_portal_events(anchor, events.clone())?;
             }
+            if let Some(sender) = &self.config.finalized_batch_submissions {
+                for submission in finalized_batches {
+                    if sender.send(submission).is_err() {
```
0xKitsune: It looks like a failed send means the observer is permanently dead, but we only warn and silently drop every later submission.

crates/l1/src/tests.rs
```diff
@@ -1652,6 +1653,87 @@ fn corrupt_recognized_portal_log(portal: Address) -> Log {
...
+    assert_eq!(submissions.len(), 1);
+    assert_eq!(submissions[0].block_number, 10);
+    assert_eq!(submissions[0].transaction_hash, B256::with_last_byte(0xaa));
+    assert_eq!(submissions[0].log_index, 4);
+    assert_eq!(submissions[0].event.nextBlockHash, B256::repeat_byte(0x22));
+}
+
+#[test]
+fn finalized_batch_observer_errors_do_not_fence_ingestion() {
+    use alloy_sol_types::SolEvent as _;
```
0xKitsune: nit: can we import this at the top of the tests mod once rather than in each test function?

crates/node/src/node.rs
```diff
@@ -198,6 +199,17 @@ pub struct ZoneSequencerAddOnsConfig {
...
 
+/// Configuration for detached shadow proving that does not require sequencer keys.
+#[derive(Debug, Clone)]
+pub struct ZoneShadowProverAddOnsConfig {
+    /// Zone ID bound into SPF public inputs.
+    pub zone_id: u32,
+    /// EIP-2935 history settings used to recover recently finalized submissions.
+    pub batch_anchor_config: BatchAnchorConfig,
+    /// Remote prover TCP address. When absent, execute the SPF in-process.
+    pub prover_address: Option<String>,
```
0xKitsune: minor nit: might be nice to make this an enum. Something like:

```rust
pub enum ProverRuntime {
    InProcess,
    Remote(IPAddr)
}
```

Feel free to ignore this if preferred.

crates/l1/src/subscriber.rs
rkrasiuk: reasonable, let's do in the follow up

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        info!(
+            target: "zone::node::shadow_prover",
+            zone_from = from,
+            zone_to = to,
+            anchor_number = anchor.number,
+            anchor_hash = %anchor.hash,
+            "Queueing finalized quorum-certified batch for shadow proving"
+        );
+        self.prover
```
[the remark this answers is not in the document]
rkrasiuk: fixed in https://github.com/tempoxyz/zones/pull/1324/changes/7a9dd64b10da3dbd6cb287feea5884baace9806b, finalized-settlement jobs now call `validate_settlement_boundary` before proving.

crates/node/src/shadow_prover.rs
rkrasiuk: addressed in https://github.com/tempoxyz/zones/pull/1324/changes/7a9dd64b10da3dbd6cb287feea5884baace9806b

crates/l1/src/subscriber.rs
rkrasiuk: addressed in https://github.com/tempoxyz/zones/pull/1324/changes/adf2005b50487d2edbfeab322d3a867c09cc043f

crates/l1/src/subscriber.rs
rkrasiuk: addressed in https://github.com/tempoxyz/zones/pull/1324/commits/d9c2b9752adfd94ef625894a3635554affefbc2a

crates/l1/src/subscriber.rs
```diff
@@ -36,6 +37,54 @@ pub struct AuthenticatedPortalLogs {
...
+                    target: "zone::l1::subscriber",
+                    block_number = block.number,
+                    error = ?err,
+                    "Could not decode finalized batch submission for observer"
+                ),
+            }
+        }
+    }
+    submissions
+}
```
shekhirin: definitions are already way too out of order in this file, but can we move this function below where free functions are

rkrasiuk:
cyclops audit fast

shekhirin (dismissed):

shekhirin (dismissed):

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,400 @@
...
+                        error!(
+                            target: "zone::node::shadow_prover",
+                            l1_block_hash = %submission.block.hash,
+                            transaction_index = submission.transaction_index,
+                            log_index = submission.log_index,
+                            error = ?err,
+                            attempts = attempt,
+                            "Failed to process finalized batch submission; retry budget exhausted"
+                        );
+                        return;
```
0xKitsune, line 71: It looks like retry exhaustion or a temporarily missing block permanently drops the finalized submission because there is no persisted cursor or startup replay.

crates/node/src/shadow_prover.rs
```diff
@@ -0,0 +1,400 @@
...
+            anchor_number = anchor.number,
+            anchor_hash = %anchor.hash,
+            "Queueing finalized quorum-certified batch for shadow proving"
+        );
+        self.prover
+            .enqueue_with_anchor(from, to, batch, anchor)
+            .await
+    }
+
+    async fn wait_for_local_range(&self, to: u64, batch: &BatchData) -> Result<u64> {
```
0xKitsune, line 260: nit: can we add some brief/succinct docs to these fns?

0xKitsune (dismissed):

crates/node/src/shadow_prover.rs
rkrasiuk: i intentionally removed replay for the shadow prover. i think we should get an alert on failure and investigate it if it failed beyond retries. adding state tracking specifically for this task is an overkill imo (which batches were validate or not)

0xKitsune:
cyclops audit fast

0xKitsune (approved):

PR #1331, opened 2026-08-31
feat(checker): verify Portal balance conservation

grandizzy:
## Summary

Adds receipt-level Portal custody reconciliation and exact balance conservation checks.

The existing solvency check verifies that Portal custody covers bridge liabilities. However, excess collateral could hide an unauthorized loss while the Portal still appears solvent. This change additionally verifies that every balance change is explained by authenticated custody movements.

## Checks

For each successful Tempo receipt:

- Observed outflows must exactly match protocol-authorized outflows.
- Observed inflows must cover protocol-required inflows.
- Additional inbound transfers are treated as unattributed surplus and cannot authorize an unrelated outflow.

For every imported Tempo block and enabled token:

```text
Portal balance(tip)
  = Portal balance(parent)
  + observed inflows
  - authorized outflows
```

This detects missing deposits, incorrect withdrawals or refunds, and otherwise unexplained custody losses.

## Fees

Fee amounts are included in the expected custody movements. Fees are currently always zero, so this does not change current accounting results, but the invariant remains valid when nonzero fees are enabled.

No checker database migration or historical backfill is required.

## Performance

Verified tip balances are cached and reused as the next block’s parent balances. This avoids additional steady-state L1 queries; full parent reads are required only on startup, recovery, cache mismatch, or for newly enabled tokens.

### DIFF
--- a/crates/checker/DESIGN.md
+++ b/crates/checker/DESIGN.md
@@ -81,7 +81,25 @@ ghost account_total(token) == sum of ghost account balances(token)
 
 ### Portal collateral
 
-For every enabled token at the exact imported Tempo block:
+For every successful Tempo receipt, Portal-affecting TIP-20 transfers must
+satisfy:
+
+```text
+observed outflow == protocol-authorized outflow
+observed inflow  >= protocol-required inflow
+```
+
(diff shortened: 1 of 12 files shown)


grandizzy:
cyclops audit

PR #1341, opened 2026-09-01
 fix(prover): model Tempo imports by call variant

rkrasiuk:
## Summary

Represent Zone Tempo imports with a typed `TempoImport` enum. Distinguish full `advanceTempo` imports from checkpoint-only `advanceTempoHeaders` imports. Validate T12 batch shape using the import variant.

### DIFF
--- a/crates/spf/src/lib.rs
+++ b/crates/spf/src/lib.rs
@@ -4,11 +4,9 @@
 //! execution. It is presently a normal Rust verifier rather than a `no_std`
 //! proving guest.
 
-use alloy_consensus::{BlockHeader as _, Sealable as _, Transaction as _};
-use alloy_eips::eip2718::Decodable2718 as _;
+use alloy_consensus::{BlockHeader as _, Sealable as _};
 use alloy_primitives::{B256, U256, keccak256};
 use alloy_rlp::Decodable as _;
-use alloy_sol_types::SolInterface as _;
 use reth_chainspec::EthChainSpec as _;
 use reth_evm::execute::BlockAssemblerInput;
 use reth_primitives_traits::SealedHeader;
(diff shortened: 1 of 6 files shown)


crates/spf/src/lib.rs
```diff
@@ -441,24 +428,41 @@ fn validate_tempo_anchor(
     Ok(())
 }
 
+fn validate_t12_block_shape(block: &ZoneBlock, is_last: bool) -> Result<(), Error> {
+    let valid = match &block.tempo_import {
+        TempoImport::CheckpointOnly { .. } => !is_last,
+        TempoImport::Full { .. } => is_last && block.finalize_withdrawal_batch_count.is_some(),
```
shekhirin, line 434: is this right? Previous condition allowed to not have `finalize_withdrawal_batch_count` https://github.com/tempoxyz/zones/blob/180352c3c22d8b7e82287141440127ba73ffd4f1/crates/spf/src/lib.rs#L253-L258

crates/spf/src/lib.rs
```diff
@@ -441,24 +428,41 @@ fn validate_tempo_anchor(
...
+            ..
+        } => {
+            if deposits.len() > MAX_UNPROCESSED_DEPOSITS
+                || enabled_tokens.len() > MAX_UNPROCESSED_TOKEN_ENABLEMENTS
+            {
+                return Err(Error::PortalWorkCapacityExceeded { block_index: index });
+            }
+        }
+        TempoImport::CheckpointOnly { .. } => {
+            if !block.transactions.is_empty()
```
shekhirin, line 459: we should also check deposits here, as they can be non-zero and `block.transactions.is_empty() == true`

crates/spf/src/lib.rs
shekhirin: yeah I think this is right, all batch boundary `advanceTempo` blocks should have it https://github.com/tempoxyz/zones/blob/4e4f98e061d706497c750d4a31a077fb43db6704/crates/payload/src/builder.rs#L588-L595

shekhirin:
cyclops audit fast

crates/spf/src/lib.rs
rkrasiuk: deposits are non-existent in this variant

PR #1344, opened 2026-09-01
fix(node): validate peer block body commitments

0xalpharush:
## Summary

- decode replicated peer blocks with `decode_sealed`, preserving the exact received header hash
- validate transaction, ommer, and withdrawal commitments before submitting the block to the Engine API
- add a regression test for two body encodings that share a header hash but disagree on withdrawals

## Why

A peer block can retain the same hashed header while omitting the withdrawals body field. Passing that mismatched body into the Engine API can mark the shared header hash invalid, causing a later valid representation to be rejected. Validate the body/header relationship at the replication boundary before it can reach that cache.

## Testing

- `cargo test -p zone-node replication::tests --lib`
- `cargo fmt --check --all`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13882,6 +13882,7 @@ dependencies = [
  "reqwest",
  "reth-chain-state",
  "reth-chainspec",
+ "reth-consensus-common",
  "reth-eth-wire-types",
  "reth-ethereum",
  "reth-evm",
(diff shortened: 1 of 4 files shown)


0xalpharush:
cyclops audit

adityapk00 (approved):

decofe:
Closing per author cleanup. No one-to-one replacement landed; the active peer-block decoding hardening is tracked in #1336, with overlapping #1337.

PR #1356, opened 2026-09-02
refactor(l1): simplify subscriber state ownership

0xKitsune:
This PR simplifies the L1 subscriber's state ownership and makes the ingestion path easier to follow.

Previously, `L1SubscriberConfig` mixed durable settings with runtime objects such as caches, key state, and the block tracker. The subscriber also erased its local state provider behind a custom checkpoint-reader trait. This obscured where the persisted Zone checkpoint came from and made configuration responsible for mutable runtime state.

The subscriber is now generic over the existing `StateProviderFactory` and reads the latest persisted checkpoint directly from that provider.

```rust
pub struct L1Subscriber<P> {
    config: L1SubscriberConfig,
    provider: P,
    // --snip--
}
```

Runtime dependencies are now explicit subscriber fields, while `L1SubscriberConfig` contains only actual settings.

```rust
pub struct L1SubscriberConfig {
    pub l1_rpc_url: String,
    pub portal_address: Address,
    pub l1_fetch_concurrency: usize,
    pub retry_connection_interval: Duration,
    pub retain_portal_evidence: bool,
}
```

### DIFF
--- a/crates/l1/src/subscriber.rs
+++ b/crates/l1/src/subscriber.rs
@@ -1,5 +1,5 @@
 use super::*;
-use eyre::WrapErr as _;
+use eyre::{OptionExt as _, WrapErr as _};
 use std::collections::HashSet;
 use tempo_contracts::precompiles::{ITIP20::TransferPolicyUpdate, TIP403_REGISTRY_ADDRESS};
 use tempo_primitives::is_tip20_prefix;
@@ -562,12 +562,12 @@ impl L1Subscriber {
         &self,
         l1_provider: &impl Provider<TempoNetwork>,
     ) -> Result<u64, L1SubscriberError> {
-        l1_provider
+        Ok(l1_provider
             .get_header_by_number(BlockNumberOrTag::Finalized)
             .await
             .inspect_err(|_| self.subscriber_metrics.fetch_failures.increment(1))?
             .map(|header| header.number())
-            .ok_or_else(|| eyre::eyre!("L1 finalized block is not available").into())
+            .ok_or_eyre("L1 finalized block is not available")?)
     }
 
     /// Synchronize all missing blocks through the current finalized L1 head.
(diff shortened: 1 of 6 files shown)


0xKitsune:
cyclops audit fast

crates/l1/src/subscriber.rs
```diff
@@ -364,54 +367,34 @@ pub struct L1SubscriberConfig {
...
-    }
-}
-
 /// L1 chain subscriber that listens for new blocks and extracts deposit events.
 #[derive(Clone)]
-pub struct L1Subscriber {
+pub struct L1Subscriber<P> {
     pub(crate) config: L1SubscriberConfig,
-    pub(crate) local_state: Arc<dyn LocalTempoCheckpointReader>,
+    pub(crate) provider: P,
```
adityapk00: nit: I got confused that this was the L1 provider (`l1_provider`) that's passed to follow_finalized below, before realizing this is the zone's DB provider. Maybe `zone_provider`? no big deal, though

crates/node/src/node.rs
```diff
@@ -592,6 +605,7 @@ where
         // snapshot at the local Tempo anchor, and install the transition sink before
         // the subscriber starts so no block is ever consumed ahead of its
         // leadership transition.
+        let mut leadership_sink: Option<Arc<dyn LeadershipSink>> = None;
```
adityapk00: nit: leadership_sink doesn't need to be `mut`, right? like 
```
let leadership_sink =
      if let Some(p2p) = self.p2p_config.as_ref() {
          // ...
          Some(Arc::new(ScheduleLeadershipSink { /* ... */ }))
      } else {
          None
      };
```

adityapk00 (dismissed):

crates/node/src/node.rs
0xKitsune: Addressed in https://github.com/tempoxyz/zones/pull/1356/changes/256fc0449637a97f6a1337c03d3ec3f7633d5aa0

crates/l1/src/subscriber.rs
0xKitsune: Addressed in https://github.com/tempoxyz/zones/pull/1356/changes/256fc0449637a97f6a1337c03d3ec3f7633d5aa0

adityapk00 (approved):

PR #1357, opened 2026-09-02
fix(l1): wake consumer when finalized target is ready

shekhirin:
The subscriber handles a finalized target in two stages:
1. It announces target N as `ready = false`.
2. It downloads and verifies the range.
3. It rereads L1 to ensure the finalized target did not advance.
4. It marks N as `ready = true`.

While N is not ready, the engine intentionally refuses to process it as a full advanceTempo import. This prevents it from treating a partially downloaded range as complete.

The fix sends a queue notification immediately after N becomes ready.


shekhirin:
cyclops audit fast

klkvr (approved):

PR #1366, opened 2026-09-03
fix(ci): use enterprise-owned Rust actions

decofe:
The existing `setup-rust-build` composite expands to `dtolnay/rust-toolchain`, `rui314/setup-mold`, and `mozilla-actions/sccache-action`, which the enterprise action policy now rejects before jobs start. Replace every remaining use with the equivalent enterprise-owned vendored actions while preserving disabled mold/sccache settings and the macOS mold guard.

Validated by parsing all workflow YAML, `zizmor --min-severity high`, and `git diff --check`.

Prompted by: @grandizzy

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -39,9 +39,13 @@ jobs:
           submodules: recursive
           persist-credentials: false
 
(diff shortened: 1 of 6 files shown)


grandizzy:
cyclops audit fast

grandizzy (approved):

mattsse (approved):

PR #1369, opened 2026-09-03
feat(sequencer): collect block proofs before canonicalization

shekhirin:
Ref: https://linear.app/tempoxyz/issue/ZONE-766/collect-zone-and-l1-proofs-proactively

Collect and persist Zone and L1 witnesses before the historical state needed to generate them disappears. The collector must retain witnesses for the entire unfinalized tail: every canonical Zone block above the portal's finalized `zoneHeight()`. Witnesses are pruned only after settlement is finalized.

Locally produced blocks become canonical only after their witnesses are synced to disk. One collector runs across leadership changes and collects missing witnesses while following or after a restart. Catch-up can temporarily leave gaps; storage write or pruning failures shut down the node.

The shadow prover reuses retained witnesses. They currently remain in both RAM and on disk, so usage grows if settlement stalls. P2P witness propagation is follow-up work.

Storage format v2 requires regenerating v1 files. Formatting and scoped Clippy passed; tests were not run.


klkvr:
cyclops audit fast

crates/sequencer/src/prover.rs
```diff
@@ -292,17 +293,31 @@ pub fn spawn_shadow_prover<P: ZoneSequencerProvider>(
...
 
-    ShadowProver { sender }
+    ShadowProver { sender, proofs }
 }
 
 impl ShadowProver {
     /// Queue a candidate without waiting for validation or queue capacity.
-    pub(crate) fn try_enqueue(&self, from: u64, to: u64, batch: BatchData) {
+    pub(crate) async fn try_enqueue(&self, from: u64, to: u64, batch: BatchData) {
+        let proofs = match self.proofs.wait_for_range(from, to).await {
```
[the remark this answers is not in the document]
klkvr: fixed in 3963b12

crates/node/src/node.rs
```diff
@@ -824,17 +825,44 @@ where
                 Ok(())
             })
             .await?;
+        let proof_collector_config = ProofCollectorConfig {
+            directory: proof_directory,
+            debug_api: Arc::new(NodeZoneDebugApi::new(
+                handle.eth_handlers().api.clone(),
+                l1_provider.clone(),
+            )),
+            settlement: self
```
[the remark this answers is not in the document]
klkvr: fixed in 3963b12

crates/node/src/engine.rs
```diff
@@ -399,23 +418,33 @@ impl ZoneEngine {
             eyre::bail!("Invalid payload for block {block_number}");
         }
 
-        // newPayload succeeded — remove the exact finalized L1 block that
-        // produced it. A mismatch indicates an internal consumer-ordering bug.
-        self.deposit_queue.confirm(l1_num_hash)?;
-        self.l1_block_tracker.prune_through(l1_num_hash.number);
-        if let Some(permit) = &self.production_permit {
-            permit.record_applied_anchor(l1_num_hash.number);
+        if let Some(collector) = &self.proof_collector {
```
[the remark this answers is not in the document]
klkvr: this is by design, though @shekhirin i'm not sure if we're fine with this behavior for shadow prover? do we enable it on any of the sequencers?

klkvr:
cyclops audit fast

klkvr:
cyclops audit fast

crates/sequencer/src/proofs.rs
```diff
@@ -0,0 +1,692 @@
...
+            drop(removed);
+        }
+        sync_directory(&self.directory)
+    }
+}
+
+#[derive(Debug)]
+struct ProofStoreState {
+    pruned_through: u64,
+    proofs: BTreeMap<u64, Arc<StoredBlockProof>>,
```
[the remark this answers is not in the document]
klkvr: not sure if worth addressing, even if settlement somehow stalls for 24 hrs the memory footprint should be around 2gib at most, and all of the witnesses are anyway going to have to fit into memory to construct prover inputs

crates/sequencer/src/proofs.rs
```diff
@@ -0,0 +1,692 @@
...
+        let start = self.store.state.read().pruned_through.saturating_add(1);
+        for number in start..=head {
+            let block_hash = self
+                .provider
+                .block_hash(number)?
+                .ok_or_eyre(format!("canonical Zone block {number} has no hash"))?;
+            if !self.store.contains(number, block_hash) {
+                self.collect_and_persist(block_hash).await?;
+            }
+        }
```
0xrusowsky, line 162: rather than aborting the whole pass on a collection failure, should we log the gap and continue?
otherwise i think one expired witness blocks the entire tail (allowing potentially recoverable witnesses to expire)

0xrusowsky: i think overstated, but codex thinks that:
> this pass processes the entire backlog before returning to the request queue, so a node promoted during catch-up can stall block production until historical collection finishes.

klkvr: we really want to avoid gaps here because we need witnesses for entire unproved tail to actually prove the batch, so even if we let it proceed further this is pretty much unrecoverable state

crates/node/src/cli.rs
```diff
@@ -585,6 +587,8 @@ pub struct ZoneArgs {
 
     /// Require Nitro-attested SPF validation for settlement, or run observational SPF validation
     /// on an rpc_only follower.
+    ///
+    /// Implies enable_proof_persistence.
```
shekhirin: we don't have `enable_proof_persistence` on `ZoneArgs`
```suggestion
```

crates/node/src/engine.rs
```diff
@@ -406,8 +420,25 @@ impl ZoneEngine {
...
+        }
+
+        // Canonicalize the new head — FCU-with-attrs above only set the
+        // *previous* head as canonical; this bare FCU makes the just-built
+        // block the EL's canonical head.
+        let forkchoice = ForkchoiceState::same_hash(header.hash());
+        let result = self.to_engine.fork_choice_updated(forkchoice, None).await?;
+        if !result.is_valid() {
+            eyre::bail!("Invalid post-newPayload fork choice update {forkchoice:?}: {result:?}");
+        }
```
shekhirin: is not using `update_forkchoice_state` a drive-by fix here?

crates/node/src/node.rs
```diff
@@ -837,6 +848,43 @@ where
...
+                        handle.eth_handlers().api.clone(),
+                        l1_provider.clone(),
+                    )),
+                    portal_address: self.portal_address,
+                    l1_provider: l1_provider.clone(),
+                };
+                let (collector, collector_task) = spawn_proof_collector(
+                    proof_collector_config,
+                    provider.clone(),
+                    tokio_util::sync::CancellationToken::new(),
```
shekhirin: should this be a shared cancellation token with sequencer task?

crates/sequencer/src/proofs.rs
```diff
@@ -0,0 +1,798 @@
...
+use tempo_alloy::TempoNetwork;
+use tempo_zone_contracts::ZonePortal;
+use tokio::sync::{mpsc, oneshot};
+use tokio_util::sync::CancellationToken;
+use tracing::{error, info};
+use zone_rpc::{ZoneDebugApi, types::ZoneExecutionWitness};
+
+use crate::ZoneSequencerProvider;
+
+const FORMAT_VERSION: u32 = 2;
```
shekhirin: should be 1, but generally unsure if we need versioning?

klkvr: i think it doesn't hurt, changed to 1

crates/node/src/node.rs
klkvr: changed to just spawn the future directly onto node's runtime, that way it should respect the global shutdown

crates/node/src/engine.rs
klkvr: hmm yeah and i don't think it really matters much, reverted

crates/node/src/cli.rs
klkvr: fixed

shekhirin:
+1

PR #1372, opened 2026-09-03
feat(evm): align Zones with Tempo T11 precompile decoding

legion2002:
## Summary

Updates the Zones EVM integration to Tempo PR [#7444](https://github.com/tempoxyz/tempo/pull/7444) and makes Zone precompile admission follow the active Tempo hardfork.

- aligns Tempo, Reth, Alloy, Commonware, and lockfile revisions
- passes the active TempoHardfork into precompile input-gas calculation
- uses the Tempo fork-aware ABI decoder configuration in every Zone admission wrapper
- retains permissive decoding before T11 and enables strict decoding at T11

## Why

Tempo precompile input cost and ABI decoding are now hardfork-dependent. Zones wraps Tempo precompiles with its own authorization layer, so decoding calldata with a different configuration before forwarding can produce caller-dependent results for malformed input.

Using the same Tempo configuration in both layers ensures historical pre-T11 blocks retain their old decoding and gas behavior, while T11 blocks use the increased input cost, strict ABI validation, and the allocation limit selected by Tempo.

Zone chain specs already inherit Tempo fork activation times, and ZoneEvmConfig selects the Tempo spec from each block timestamp. This change carries that selected spec through the wrapper admission paths.

## Reexecution

- pre-T11: 6 gas per calldata word and permissive ABI decoding
- T11+: 30 gas per calldata word and strict ABI decoding
- historical blocks derive behavior from their original block timestamp

## Validation

- cargo test -p zone-precompiles --lib --no-fail-fast: 152 passed, 1 existing ignored
- cargo check -p zone-evm
- T10/T11 input-gas boundary regression
- pre-T11 noncanonical-address privacy regression
- T11 owner/outsider malformed-calldata consistency regression
- existing timestamp-based Tempo fork-selection coverage

## Stack note

This is the base of a two-PR stack.

The dependency alignment includes a newer Reth revision that removed the old HashedStorage.wiped and sparse-trie wipe APIs. As a result, the full workspace check stops at the expected zone-spf compatibility errors on this PR alone. The stacked child adapts stateless proof execution and the node APIs to the new Reth revision, and restores the full workspace build.

This split keeps the Tempo/T11 behavior change independently reviewable.

### DIFF
--- a/crates/node/src/replication.rs
+++ b/crates/node/src/replication.rs
@@ -18,6 +18,8 @@ use std::{
     time::Duration,
 };
 use tempo_alloy::TempoNetwork;
+use tempo_chainspec::hardfork::TempoHardfork;
+use tempo_precompiles::dispatch::abi_decoder_config_for_spec;
 use tempo_primitives::{Block, TempoHeader, TempoTxEnvelope};
 use tokio::sync::{mpsc, oneshot};
 use tokio_util::sync;
@@ -1375,8 +1377,11 @@ fn decode_advance_tempo(
     if signed.tx().to != ZONE_INBOX_ADDRESS.into() {
         eyre::bail!("first Tempo system transaction is not sent to IZoneInbox")
     }
-    let call = IZoneInbox::advanceTempoCall::abi_decode(signed.tx().input.as_ref())
-        .map_err(|err| eyre::eyre!("first transaction does not decode as advanceTempo: {err}"))?;
+    let call = IZoneInbox::advanceTempoCall::abi_decode_with_config(
+        signed.tx().input.as_ref(),
+        abi_decoder_config_for_spec(TempoHardfork::latest()),
+    )
+    .map_err(|err| eyre::eyre!("first transaction does not decode as advanceTempo: {err}"))?;
 
     // 3. the system tx is valid.
     let mut header_rlp = call.header.as_ref();
--- a/crates/precompiles/src/account_keychain.rs
+++ b/crates/precompiles/src/account_keychain.rs
@@ -1,7 +1,7 @@
 //! Zone read-privacy rules for the upstream Tempo AccountKeychain precompile.
 
 use alloy_primitives::Address;
-use alloy_sol_types::SolInterface;
+use alloy_sol_types::{SolCall, SolInterface};
 use tempo_contracts::precompiles::IAccountKeychain;
 use tempo_precompiles::dispatch::abi_decoder_config_for_spec;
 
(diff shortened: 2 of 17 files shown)


crates/precompiles/src/tip403_proxy/mod.rs
```diff
@@ -25,7 +25,12 @@ alloy_sol_types::sol! {
 pub(crate) struct Tip403Rules;
 
 impl CallRules for Tip403Rules {
-    fn admit(&self, _data: &[u8], caller: Address) -> CallCheck {
+    fn admit(
+        &self,
+        _data: &[u8],
+        caller: Address,
+        _spec: tempo_chainspec::hardfork::TempoHardfork,
```
klkvr: we have spec available via `StorageCtx` so extra arg should not be necessary?

crates/precompiles/src/account_keychain.rs
0xalpharush, line 83: we can only check these selectors using `.get(..4)` after T11

crates/precompiles/src/tip403_proxy/mod.rs
legion2002: addressed in https://github.com/tempoxyz/zones/pull/1372/commits/961e6640a8a292c854bb2045a851733fb373aba2

klkvr (dismissed):

klkvr (dismissed):

legion2002:
cyclops audit fast

crates/node/src/replication.rs
```diff
@@ -1375,8 +1377,11 @@ fn decode_advance_tempo(
     if signed.tx().to != ZONE_INBOX_ADDRESS.into() {
         eyre::bail!("first Tempo system transaction is not sent to IZoneInbox")
     }
-    let call = IZoneInbox::advanceTempoCall::abi_decode(signed.tx().input.as_ref())
-        .map_err(|err| eyre::eyre!("first transaction does not decode as advanceTempo: {err}"))?;
+    let call = IZoneInbox::advanceTempoCall::abi_decode_with_config(
+        signed.tx().input.as_ref(),
+        abi_decoder_config_for_spec(TempoHardfork::latest()),
```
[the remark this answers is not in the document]
legion2002: the leader encodes without trailing bytes in our current setup.

crates/spf/src/mpt.rs
```diff
@@ -130,10 +130,9 @@ impl StatelessSparseTrie {
             for (hashed_address, storage) in storage_updates {
                 let current_account = self.trie_account(hashed_address)?;
                 let has_revealed_storage = self.inner.storage_trie_ref(&hashed_address).is_some();
-                if !storage.wiped
-                    && current_account
-                        .as_ref()
-                        .is_some_and(|account| account.storage_root != EMPTY_ROOT_HASH)
+                if current_account
```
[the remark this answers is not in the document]
legion2002: there won't be any storage wipes on zones

jenpaff (approved):

PR #1378, opened 2026-09-03
refactor(node): encapsulate follower block sync

adityapk00:
## Summary
- introduce a generation-owned follower block sync object
- encapsulate pending peer-block buffering in PendingBlocks

### DIFF
--- a/crates/node/src/replication.rs
+++ b/crates/node/src/replication.rs
@@ -1119,7 +1119,7 @@ where
 
         // 3. Require the block to advance the local Tempo checkpoint by exactly
         // one independently observed L1 block.
-        let (l1_header, portal_inputs) = Self::decode_advance_tempo(&block)?;
+        let (l1_header, portal_inputs) = decode_advance_tempo(&block)?;
         let local = self
             .provider
             .state_by_block_hash(parent.hash())?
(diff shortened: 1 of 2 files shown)


adityapk00:
cyclops audit fast

crates/node/src/replication.rs
```diff
@@ -732,569 +765,508 @@ pub(crate) async fn run_follower_block_sync<P>(
...
-    block_number: u64,
-) -> eyre::Result<()> {
-    let Some(sender) = live_sender else {
-        return Ok(());
-    };
-    match schedule.leader_for(anchor_number) {
-        Some(record) if &record.leader == sender => Ok(()),
-        Some(record) => eyre::bail!(
-            "live block {block_number} for anchor {anchor_number} was broadcast by {sender}, but \
+impl<P> FollowerBlockSync<P> {
```
0xrusowsky: `validate_live_block_sender` and `wait_for_validated_peer_anchor` don't require neither `self` not the provider, so they should be free fns imo

crates/node/src/replication.rs
```diff
@@ -1347,54 +1319,57 @@ fn validate_l1_checkpoint_transition(
...
 }
 
-/// Decode the L1 header embedded in the first `IZoneInbox.advanceTempo` system transaction.
-#[cfg(test)]
-fn decode_advance_tempo_header(
-    block: &SealedBlock<Block>,
-) -> eyre::Result<SealedHeader<TempoHeader>> {
-    decode_advance_tempo(block).map(|(header, _)| header)
-}
+impl<P> FollowerBlockSync<P> {
```
0xrusowsky: same reasoning here, i think `decode_advance_tempo` is better suited as a free fn.

also `decode_advance_tempo_header` is a kinda useless test helper that could be inlined?

0xrusowsky (dismissed):
lgtm, but i'd personally apply the suggested nits

crates/node/src/replication.rs

crates/node/src/replication.rs

0xrusowsky (approved):

PR #1379, opened 2026-09-03
fix: bound ABI decoding in Zone transaction paths

0xalpharush:
## Summary

- skip duplicate Zone decoding for AccountKeychain selectors that have no Zone-specific privacy policy
- strictly decode nested Inbox deposit payloads with the shared 16 MiB decoder limit
- classify finalizeWithdrawalBatch system transactions by destination and selector, matching advanceTempo
- strictly decode replicated advanceTempo calls with the shared 16 MiB limit

## Safety

This PR is stacked on #1372.

For unrestricted AccountKeychain selectors, the Zone wrapper forwards the original calldata and caller without decoding them first. The upstream Tempo precompile remains authoritative: it charges input gas, enforces selector scheduling, performs bounded ABI decoding, and applies AccountKeychain authorization and state checks. Zone privacy-sensitive getters still decode before applying their caller restrictions.

System-transaction classification now performs only constant-time destination and selector checks for both advanceTempo and finalizeWithdrawalBatch. Classification only proposes the next block phase. The phase commits after successful authoritative precompile execution, so calldata rejected during execution cannot advance it.

Nested Inbox decoding uses strict mode and the shared ABI_DECODER_MEMORY_LIMIT. Strict mode replaces the previous decode-and-reencode canonicality check. Replication applies the same bound before processing peer-supplied advanceTempo data.

## Testing

- cargo test -p zone-evm --lib: 36 passed
- cargo test -p zone-precompiles --lib: 153 passed, 1 ignored
- cargo test -p zone-precompiles inbox --lib: 23 passed
- cargo test -p zone-precompiles account_keychain --lib: 4 passed
- cargo fmt --check
- git diff --check

A full zone-node check remains blocked by an unrelated existing zone-spf API mismatch around HashedStorage.wiped/wipe.

### DIFF
--- a/crates/evm/src/executor.rs
+++ b/crates/evm/src/executor.rs
@@ -13,7 +13,7 @@ use alloy_evm::{
     },
     eth::{EthBlockExecutor, EthTxResult},
 };
-use alloy_sol_types::SolEvent as _;
+use alloy_sol_types::{SolCall as _, SolEvent as _};
 use reth_evm::block::StateDB;
 use reth_revm::{Inspector, context::result::ResultAndState};
 use tempo_evm::{TempoBlockExecutionCtx, TempoReceiptBuilder};
@@ -22,9 +22,7 @@ use tempo_revm::evm::TempoContext;
 use tempo_zone_contracts::IZoneOutbox;
 use zone_chainspec::ZoneChainSpec;
 use zone_l1::state::L1StateProvider;
-use zone_precompiles::{
-    ADVANCE_TEMPO_SELECTOR, L1StorageReader, is_finalize_withdrawal_batch_calldata,
-};
+use zone_precompiles::{ADVANCE_TEMPO_SELECTOR, L1StorageReader};
 use zone_primitives::constants::{ZONE_INBOX_ADDRESS, ZONE_OUTBOX_ADDRESS};
 
 use crate::{L1OverlayDB, ZoneEvm};
@@ -106,7 +104,8 @@ impl ZoneTransactionKind {
         }
 
         if tx.calls().any(|(kind, input)| {
-            kind.to() == Some(&ZONE_OUTBOX_ADDRESS) && is_finalize_withdrawal_batch_calldata(input)
+            kind.to() == Some(&ZONE_OUTBOX_ADDRESS)
+                && input.starts_with(&IZoneOutbox::finalizeWithdrawalBatchCall::SELECTOR)
         }) {
             return Self::FinalizeWithdrawalBatch;
         }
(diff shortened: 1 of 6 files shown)


0xalpharush:
cyclops audit

crates/node/src/replication.rs
```diff
@@ -1375,8 +1376,13 @@ fn decode_advance_tempo(
     if signed.tx().to != ZONE_INBOX_ADDRESS.into() {
         eyre::bail!("first Tempo system transaction is not sent to IZoneInbox")
     }
-    let call = IZoneInbox::advanceTempoCall::abi_decode(signed.tx().input.as_ref())
-        .map_err(|err| eyre::eyre!("first transaction does not decode as advanceTempo: {err}"))?;
+    let call = IZoneInbox::advanceTempoCall::abi_decode_with_config(
+        signed.tx().input.as_ref(),
+        AbiDecoderConfig::new()
+            .memory_limit(ABI_DECODER_MEMORY_LIMIT)
+            .strict(true),
```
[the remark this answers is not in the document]
0xalpharush: leader is normal ABI encoding

crates/precompiles/src/outbox/mod.rs
```diff
@@ -34,10 +34,11 @@ const WITHDRAWAL_BASE_GAS: u64 = 50_000;
 
 /// Returns whether `calldata` is a canonical `finalizeWithdrawalBatch` call.
 pub fn is_finalize_withdrawal_batch_calldata(calldata: &[u8]) -> bool {
-    let Ok(call) = IZoneOutbox::finalizeWithdrawalBatchCall::abi_decode(calldata) else {
-        return false;
-    };
-    call.abi_encode() == calldata
+    IZoneOutbox::finalizeWithdrawalBatchCall::abi_decode_with_config(
+        calldata,
+        AbiDecoderConfig::new().strict(true),
```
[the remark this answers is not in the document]
0xalpharush: fixed

crates/evm/src/executor.rs
```diff
@@ -106,7 +110,19 @@ impl ZoneTransactionKind {
         }
 
         if tx.calls().any(|(kind, input)| {
-            kind.to() == Some(&ZONE_OUTBOX_ADDRESS) && is_finalize_withdrawal_batch_calldata(input)
+            kind.to() == Some(&ZONE_OUTBOX_ADDRESS)
+                && if spec.is_t11() {
```
0xalpharush: tbh i think it does not need to be gated

legion2002 (dismissed):

crates/evm/src/executor.rs
legion2002: yeah we can remove. the encoding is kinda guaranteed to be correct rn

0xalpharush: removed

klkvr (approved):

PR #1384, opened 2026-09-03
refactor(node): extract follower sync module

adityapk00:
## Summary
- move follower block sync, buffering, and validation into follower.rs
- keep leader replication and role-neutral backfill serving in replication.rs


adityapk00:
cyclops audit fast

0xrusowsky (dismissed):
nit: should `AttestationContext` be moved into `settlement_attestation.rs` so both leader replication and follower sync don't have to pull it via `replication.rs`?

0xrusowsky (approved):

PR #1385, opened 2026-09-03
perf(precompiles): dedupe secp256k1 work in the ECIES path

mattsse:
Chaum-Pedersen verification computed `s*G - c*pubSeq` and `s*ephemeralPub - c*sharedSecret` as four separate scalar multiplications and re-encoded the generator on every challenge hash. The ECIES side recomputed `privSeq * G` per deposit and `eph * G` twice per authenticated withdrawal.

Each verification equation is now one two-term `lincomb` (Straus with the GLV endomorphism, the routine k256 uses for its own ECDSA verify), the generator encoding is a constant pinned by a test, and generator multiplications use `mul_by_generator` so they hit k256's precomputed tables where that feature is enabled, as it is for the node binary. A new `PreparedDecryptionKey` derives the sequencer public point once per key; `compute_ecdh_proof` and `decrypt_deposit` remain as thin wrappers. Adopting it in the L1 deposit loop is a follow-up.

All outputs are byte-identical, and the proof scalars and the deterministic withdrawal ciphertext are now pinned by fixed vectors captured from main. Measured with a throwaway interleaved A/B harness on the release profile: `verify` 172 → 137 µs, `challenge_hash` 5.6 → 0.7 µs, `compute_ecdh_proof` 173 → 132 µs (110 µs with a reused key), `decrypt_deposit` 177 → 136 µs (114 µs reused), `encrypt_authenticated_withdrawal_deterministic` 134 → 76 µs.

### DIFF
--- a/crates/precompiles/src/chaum_pedersen.rs
+++ b/crates/precompiles/src/chaum_pedersen.rs
@@ -9,7 +9,7 @@ use alloy_primitives::Keccak256;
 use k256::{
     AffinePoint, FieldBytes, ProjectivePoint, Scalar,
     elliptic_curve::{
-        ops::Reduce,
+        ops::{LinearCombination, Reduce},
         sec1::{FromEncodedPoint, ToEncodedPoint},
     },
 };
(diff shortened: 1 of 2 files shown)


mattsse:
not worth it

PR #1387, opened 2026-09-03
perf(sequencer): scan batch boundaries from receipts only

mattsse:
The zone monitor rescans for `BatchFinalized` boundaries once per zone block, and the scan read the full body of every block in the range through `block_by_number`, which reth serves as a deep clone of the body. The body was only used for a transaction hash and a count check; the events live in the receipts and the recorded transaction index is enough to find the `finalizeWithdrawalBatch` transaction later.

The scan now reads receipts only and resolves that transaction by index. The boundary block is read once per batch and shared between batch reconstruction and the commitment snapshot, which previously each re-read it, and the count check moved to that single read. Against the mock provider, a 64-block scan plus one batch drops from 66 to 1 `block_by_number` calls and from 66 to 65 `receipts_by_block` calls.

The two per-block monitor lines reporting the scanned range are now `debug!`.

### DIFF
--- a/crates/sequencer/src/monitor.rs
+++ b/crates/sequencer/src/monitor.rs
@@ -42,8 +42,8 @@ use crate::{
     resolve_portal_zone_anchor,
     settlement::{
         BatchAnchorConfig, BatchData, BatchSubmitError, BatchSubmitter, FinalizedBatchLog,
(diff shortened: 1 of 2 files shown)


mattsse:
not worth it

PR #1389, opened 2026-09-03
perf(l1): decrypt L1 deposits in parallel off the engine task

mattsse:
`L1BlockDeposits::prepare` ran the full ECIES pipeline (ECDH, Chaum-Pedersen proof, AES-GCM) inline for every encrypted deposit, and the leader awaits it from `ZoneEngine::advance` before block production starts. A block near the portal's cap of 230 deposits stalled the engine's tokio worker for the whole duration, and the key ring lock was taken and a `SecretKey` cloned per deposit.

Each deposit's result depends only on its key, its payload and the portal address, so keys are now resolved once per distinct index and the per-deposit work runs on rayon inside `spawn_blocking`, collected through an indexed pipeline so order and contents are unchanged. A missing key index fails with the same error, now before any crypto runs, and blocks without encrypted deposits stay inline. Per-deposit log lines may now interleave.

Timed with a throwaway harness over 230 encrypted deposits on a loaded 12-core machine, dev build with the crypto crates at opt-level 3: median wall time per call drops from roughly 200-225 ms to 32-37 ms, the minimum from 117 ms to 21 ms. For a single deposit the thread hop adds about 50 µs against roughly 260 µs of crypto it moves off the engine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13828,6 +13828,7 @@ dependencies = [
  "k256",
  "metrics",
  "parking_lot",
+ "rayon",
  "reth-metrics",
  "reth-primitives-traits",
  "reth-provider",
--- a/crates/l1/Cargo.toml
+++ b/crates/l1/Cargo.toml
@@ -45,6 +45,7 @@ futures.workspace = true
 k256.workspace = true
 metrics.workspace = true
 parking_lot.workspace = true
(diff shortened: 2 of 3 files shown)


mattsse (commented):
not worth doing since we dont see a lot of deposits in the same block

PR #1391, opened 2026-09-03
perf(node): serve redacted blocks from the sealed header

mattsse:
`eth_getBlockByNumber` and `eth_getBlockByHash` on the redacted RPC loaded the recovered block, collected every transaction hash and RLP-encoded the whole block for `size`, then discarded all of it: `redact_block` emptied the transaction list and `redact_header` zeroed `size`. Header-only requests for old blocks also evicted live traffic from the block cache.

The response is now built from the header alone: the block id is resolved to a hash, the header is taken from the cached block when the eth state cache holds it and otherwise read from the provider, then it is converted with the same `RpcConvert` as before with size 0, and paired with empty uncles, an empty transaction list and a withdrawals field derived from the header's withdrawals root. A block has withdrawals exactly when its header has that root, so the serialized JSON is unchanged for every block, including `latest`, hash lookups and unknown blocks. The one behaviour change is that a header whose body was pruned now returns the redacted header instead of null, which cannot occur on zone nodes since they retain bodies.

`eth_coinbase` took the same full-block path and now reads the beneficiary from `latest_header`.

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -33,10 +33,10 @@ use reth_rpc_api::Web3ApiServer;
 use reth_rpc_builder::EthHandlers;
 use reth_rpc_eth_api::{
     EthApiTypes, EthFilterApiServer, RpcConvert,
-    helpers::{EthApiSpec, EthBlocks, EthCall, EthFees, EthState, EthTransactions, FullEthApi},
+    helpers::{EthApiSpec, EthCall, EthFees, EthState, EthTransactions, FullEthApi},
 };
 use reth_rpc_eth_types::{EthApiError, logs_utils};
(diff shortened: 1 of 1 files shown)


crates/node/src/rpc.rs
```diff
@@ -739,19 +739,40 @@ impl<Api> ZoneRpc<Api>
...
-            let Some(mut block) = block else {
+            let Some(header) = self
+                .eth
+                .api
+                .provider()
+                .sealed_header_by_id(id)
+                .map_err(internal)?
+            else {
                 return Ok(raw_null());
             };
```
mattsse: it is likely that the block is cache so we should use the get_maybe_block and only on cache miss load the header

PR #1392, opened 2026-09-03
perf(node): stop rebuilding settlement attestations on every retry

mattsse:
While a batch boundary waits for L1 confirmation the leader re-proposed it every 500 ms, rebuilding the attestation each time: `previous_batch` walks a whole batch of blocks reading receipts, followed by a portal multicall and three more L1 reads. Every re-broadcast made each follower repeat that work inline on its block-import loop, and each returned signature made the leader repeat it again to verify a statement it had signed itself.

The leader now verifies a follower signature against its own stored proposal at `(height, digest)`; the digest commits to the whole statement, so this is equivalent, and `collect_follower_settlement_signatures` no longer needs the chain provider. The `previous_batch` walk is memoized per boundary height on `AttestationContext`, which is sound given instant finality. A follower remembers the last statement it signed and answers re-broadcasts from memory. The retry backs off from 500 ms to at most 5 s while the same boundary stays pending, resets when a new boundary appears, no longer fires immediately on entering the loop, and is skipped once every quorum member has signed.

Measured with the mock-based unit tests added here at the default 120-block interval: a cold build reads 121 receipt sets, 2 headers and makes 5 L1 requests; a retry at the same boundary reads 1 receipt set and 1 header; a re-broadcast or a follower-signature verification reads nothing and makes no L1 request. Moving proposal validation off the follower's import loop is a follow-up.

### DIFF
--- a/crates/node/src/replication.rs
+++ b/crates/node/src/replication.rs
@@ -15,6 +15,7 @@ use reth_provider::HeaderProvider;
 use reth_storage_api::{BlockNumReader, BlockReader, ReceiptProvider, StateProviderFactory};
 use std::{
     collections::{BTreeMap, HashMap},
+    sync::{Arc, Mutex},
     time::Duration,
 };
 use tempo_alloy::TempoNetwork;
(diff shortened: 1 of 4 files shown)


mattsse:
this is too complex

PR #1396, opened 2026-09-04
chore(deps): bump Tempo v1.14.0 and allow equal timestamps

legion2002:
## Summary

- pin Tempo to `release/v1.14.0` commit `a843c91a42a273ccff44e0b6afbee83f9b1021d3`
- align direct Reth dependencies with the release branch at `5f02aa3`
- configure both Zone consensus construction paths to allow equal millisecond timestamps and up to 100 ms of future clock drift
- allow consecutive Zone blocks to reuse their parent millisecond timestamp
- remove the former timestamp pacing sleep while keeping timestamp regressions invalid

## Why

Tempo PR #7331 keeps strict timestamp validation as the upstream default and adds an explicit opt-in for Zones. Tempo PR #7475 makes future timestamp tolerance configurable. Zones can use nondecreasing millisecond timestamps and tolerate ordinary clock skew, so catch-up no longer needs to manufacture a one-millisecond increment or wait for wall-clock time before producing each block.

The unstable-nextfork historical reexecution mismatch is intentionally out of scope for this PR. This PR does not add a Z1 or precompile compatibility gate.

The GitHub target is currently a mutable release branch rather than an immutable `v1.14.0` tag, so this PR records the exact tested commit.

## Rollout note

Upgrade followers and other validating nodes before sequencers begin using the relaxed timestamp policy. Older strict nodes reject equal parent timestamps and timestamps ahead of their local clocks.

## Validation

- `cargo test -p zone-precompiles --lib --no-fail-fast` (153 passed, 1 ignored)
- `cargo test -p zone-node engine::tests --lib` (7 passed)
- `git diff --check`

Upstream: https://github.com/tempoxyz/tempo/pull/7331, https://github.com/tempoxyz/tempo/pull/7475

### DIFF
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -124,7 +124,10 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
 
     let components = move |spec: Arc<ZoneChainSpec>| {
         let evm_config = cli_evm_config(spec.clone(), l1_config.clone());
-        (evm_config, TempoConsensus::new(spec))
+        (
+            evm_config,
+            TempoConsensus::new(spec).with_allow_equal_timestamps(true),
+        )
     };
 
     cli.run_with_components::<ZoneNode>(components, async move |mut builder, args| {
--- a/crates/node/src/cli.rs
+++ b/crates/node/src/cli.rs
@@ -126,7 +126,9 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
         let evm_config = cli_evm_config(spec.clone(), l1_config.clone());
         (
             evm_config,
-            TempoConsensus::new(spec).with_allow_equal_timestamps(true),
+            TempoConsensus::new(spec)
+                .with_allow_equal_timestamps(true)
+                .with_allowed_future_block_time_millis(100),
         )
     };
 
--- a/crates/node/src/engine.rs
+++ b/crates/node/src/engine.rs
@@ -335,11 +335,15 @@ impl ZoneEngine {
         // The L1 timestamp is a lower bound so a Zone block anchored after an L1 timestamp-based
         // fork cannot predate it. Use wall-clock time to avoid backdating transactions during
         // catch-up, while allowing multiple blocks in the same millisecond.
-        let timestamp_millis = paced_zone_timestamp_millis(
+        let wall_clock_timestamp_millis = SystemTime::now()
+            .duration_since(UNIX_EPOCH)?
+            .as_millis()
+            .try_into()?;
+        let timestamp_millis = zone_timestamp_millis(
             l1_block.header.timestamp_millis(),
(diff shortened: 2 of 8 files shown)


legion2002:
cyclops audit super-fast

crates/node/src/engine.rs
```diff
@@ -461,12 +461,9 @@ async fn paced_zone_timestamp_millis(
...
 
-    // If the next Zone timestamp is ahead of wall-clock time, wait until wall-clock catches up
-    // since validation doesn't allow future timestamps.
     if timestamp_millis > wall_clock_timestamp_millis {
-        // We'll sleep for a max of 1 second. If the delay is > 1 second, something more serious is wrong
-        // (eg. clock skew, system time jump, etc.). In this case, we'll return the timestamp anyway, and
-        // the engine will fail to produce a block (the validation will fail), which is the correct behavior.
+        // Wait for at most one second. A larger gap likely indicates clock skew or a system-time
+        // jump; return the timestamp and let engine validation surface the error in that case.
         tokio::time::sleep(
```
klkvr: we need to remove that sleep now that equal timestamps are allowed

crates/node/src/cli.rs
```diff
@@ -124,7 +124,10 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
 
     let components = move |spec: Arc<ZoneChainSpec>| {
         let evm_config = cli_evm_config(spec.clone(), l1_config.clone());
-        (evm_config, TempoConsensus::new(spec))
+        (
+            evm_config,
+            TempoConsensus::new(spec).with_allow_equal_timestamps(true),
```
klkvr: this also needs `with_allowed_future_block_time_millis(100)`

crates/node/src/node.rs
```diff
@@ -1677,7 +1677,7 @@ where
     type Consensus = TempoConsensus<ZoneChainSpec>;
 
     async fn build_consensus(self, ctx: &BuilderContext<Node>) -> eyre::Result<Self::Consensus> {
-        Ok(TempoConsensus::new(ctx.chain_spec()))
+        Ok(TempoConsensus::new(ctx.chain_spec()).with_allow_equal_timestamps(true))
```
klkvr: same here

crates/node/src/engine.rs
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/2cdbb4b1d52b04027143adeebf9dd45b836bfabe

crates/node/src/cli.rs
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/f1954dc125a846886036a0b264e678a2cd151a10

crates/node/src/node.rs
legion2002: Addressed in https://github.com/tempoxyz/zones/commit/f1954dc125a846886036a0b264e678a2cd151a10

klkvr (approved):

PR #1400, opened 2026-09-04
refactor(l1): simplify finalized block ingestion

0xKitsune:
This PR separates L1 block fetching from application, making the sync loop easier to follow: fetch a block, apply its events, then advance the cursor. It preserves receipt verification and finalized-batch observer delivery.

Stacked on #1380.

### DIFF
--- a/crates/l1/src/subscriber.rs
+++ b/crates/l1/src/subscriber.rs
@@ -4,7 +4,9 @@ use crate::{
     state::EnabledTokenRegistry,
 };
 use eyre::{OptionExt as _, WrapErr as _};
+use futures::stream;
 use std::collections::HashSet;
+use tempo_alloy::rpc::TempoTransactionReceipt;
 use tempo_contracts::precompiles::{ITIP20::TransferPolicyUpdate, TIP403_REGISTRY_ADDRESS};
 use tempo_primitives::is_tip20_prefix;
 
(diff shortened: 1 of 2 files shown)


crates/l1/src/subscriber.rs
```diff
@@ -628,200 +630,161 @@ where
...
+    ) -> Result<(TempoHeader, Vec<TempoTransactionReceipt>), L1SubscriberError> {
+        self.block_tracker.wait_for_capacity(block_number).await?;
+        let start = std::time::Instant::now();
+        let fetch_failures = &self.subscriber_metrics.fetch_failures;
+        let header_resp = async {
+            let header = l1_provider
+                .get_header_by_number(block_number.into())
+                .await?;
+            Ok::<_, L1SubscriberError>(
+                header.ok_or_eyre(format!("L1 header not found for block {block_number}"))?,
```
0xrusowsky, line 673: should we also verify `header_resp.number() == block_number` immediately after fetching the header? otherwise a provider response for a different height can technically bypass receipt verification and be applied under the "manipulated" height

crates/node/src/cli.rs
```diff
@@ -528,7 +528,8 @@ pub struct ZoneArgs {
     #[arg(
         long = "l1.fetch-concurrency",
         env = "L1_FETCH_CONCURRENCY",
-        default_value_t = 4
+        default_value_t = 4,
+        value_parser = clap::builder::RangedU64ValueParser::<usize>::new().range(1..)
```
0xrusowsky, line 532: note: this is technically a breaking cahnge since `L1_FETCH_CONCURRENCY=0` was previously accepted and normalized with `.max(1)`

PR #1410, opened 2026-09-07
fix: activate TIP-1096 at T13

shekhirin:
Ref: https://tips.sh/1096
Ref: https://github.com/tempoxyz/tempo/pull/7521

Move TIP-1096 activation from T12 to T13 across Tempo imports, token cursors and events, recovery production, and settlement ABI selection. The prover launches together with TIP-1096, so its batch-shape and outstanding-work checks apply unconditionally.

Pin Tempo to main revision `346c22eb`, which includes the T13 plumbing and matching L1 runtime activation. Stacked on `prover` (#1320).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11414,7 +11414,7 @@ dependencies = [
 [[package]]
 name = "tempo-alloy"
 version = "1.11.0"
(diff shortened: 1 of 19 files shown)


shekhirin:
cyclops audit fast

klkvr (approved):

PR #1412, opened 2026-09-07
fix(node): enforce first-header leader for recovery blocks

shekhirin:
Ref: https://tips.sh/1096

Header-only recovery blocks bypassed the production permit and live P2P sender check, allowing different sequencers to build competing catch-up chains. Select the leader at the first imported Tempo header for both checks, even when the range crosses a leadership change. Each subsequent block derives its leader independently from its own first header using the existing schedule and forced-recovery override.


shekhirin:
cyclops audit fast

klkvr (approved):

PR #1416, opened 2026-09-08
refactor(contracts): replace placeholder verifier with test mock

decofe:
Remove the always-accepting `Verifier.sol` runtime and use the existing `MockVerifier` in Solidity tests. Initialize the mock after `vm.etch` so fixtures continue to accept proofs, and rename the verifier unit test to match.

Remove the deleted artifact from the runtime-sync workflow, benchmark cache inputs, and contract documentation. The benchmark genesis helper now obtains its verifier bytecode from the pinned Tempo dependency; `Cargo.lock` remains part of the benchmark cache key.

Validation:
- `MockVerifierTest`: 1 passed with `FOUNDRY_HARDFORK=cancun` (Solc 0.8.35).
- Changed Solidity files pass `forge fmt --check` with the same config override.
- `cargo +nightly fmt --all -- --check`, shell syntax validation, and `git diff --check` pass.
- Full Tempo contract tests are blocked locally: installed Forge and the downloadable nightly do not recognize `tempo:T8`.
- The focused xtask Rust test was started but stopped during the initial dependency build; it has not been validated by execution.

Prompted by: @0xalpharush


xtask/src/install_reference_zone_factory.rs
```diff
@@ -142,7 +142,8 @@ fn install_eip2935_history_storage(genesis: &mut Genesis) -> eyre::Result<()> {
 fn load_native_artifacts(specs_out: &Path) -> eyre::Result<NativeArtifacts> {
     Ok(NativeArtifacts {
         portal: load_runtime(specs_out, "ZonePortal")?,
-        verifier: load_runtime(specs_out, "Verifier")?,
+        // The verifier runtime is supplied by Tempo rather than a local Solidity artifact.
+        verifier: TEMPO_ZONE_VERIFIER_RUNTIME,
```
0xKitsune: Looks like this uses the legacy verifier runtime, which doesn't accept the new `TokenEnablementTransition` argument, so batch submissions will revert. should this use `T13_ZONE_VERIFIER_RUNTIME`?

0xKitsune (approved):

0xKitsune:
cyclops audit fast

PR #1417, opened 2026-09-08
feat(bench): select Tempo hardfork != `latest`

0xrusowsky:
## Why

Zone benchmarks currently inherit every Tempo localnet hardfork at genesis. That makes it impossible to benchmark a Zones revision against an earlier compatible protocol lane while newer Zone runtime changes are still staged elsewhere.

## What changes

- Add an optional `tempo-hardfork` workflow-dispatch input, defaulting to `latest`.
- Discover the pinned Tempo revision's supported hardforks from `generate-localnet --help`, so adding a future hardfork requires no Zones workflow update.
- Explicitly activate every fork through the selected fork at genesis and schedule later forks for 2100.
- Include the resolved hardfork in the persistent L1 snapshot configuration/cache key.
- Record the resolved hardfork in the workflow summary, run metadata, and ClickHouse metadata.
- Keep the dispatch API within GitHub's 25-input limit by fixing the ordinary Zone activity amount to its existing default of `1` instead of exposing it as a manual input.

With the current Tempo pin, `latest` resolves to `t12`; callers can select `t11` while Zones `main` remains on the pre-TIP-1096 protocol.

## Validation

- Parsed `.github/workflows/zones-benchmark.yml` as YAML and verified it has exactly 25 dispatch inputs.
- `actionlint 1.7.12 .github/workflows/zones-benchmark.yml`
- `bash -n contrib/bench/l1-snapshot.sh contrib/bench/scenario-reporting.sh`
- `git diff --check`
- Exercised resolution against the exact pinned Tempo `a843c91a` xtask:
  - `latest` resolved to `t12`.
  - `T11` normalized to `t11`.
  - unsupported and malformed forks were rejected.
- Generated a localnet genesis for the `t11` lane and verified `t10Time=0`, `t11Time=0`, and `t12Time=4102444800`.
- [End-to-end benchmark run](https://github.com/tempoxyz/zones/actions/runs/34243708214): the `t11` snapshot rebuilt successfully and the encrypted-deposit workload completed 1/1 journeys with no failures.

### DIFF
--- a/.github/workflows/zones-benchmark.yml
+++ b/.github/workflows/zones-benchmark.yml
@@ -141,6 +141,11 @@ on:
         type: string
         required: false
         default: ""
+      tempo-hardfork:
+        description: Latest Tempo hardfork active during the benchmark, or `latest`.
+        type: string
+        required: false
+        default: latest
       txgen-ref:
         description: Exact txgen commit SHA; empty resolves the current txgen HEAD.
         type: string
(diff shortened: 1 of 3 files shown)


0xrusowsky:
cyclops audit super-fast

Rjected (approved):

PR #1427, opened 2026-09-09
fix(bench): align contract runtimes across Tempo forks

decofe:
The benchmark installs the current Zones contracts, but T12 replaces them with Tempo's bundled bytecode and fails the post-block runtime check. Embed the benchmark artifacts in every Tempo runtime set before building, so genesis and fork upgrades agree for the selected fork; keep both validators' bytecode checks intact.

Build the patched Tempo source outside the upstream binary cache and include its runtime hash in snapshot metadata. The genesis installer accepts matching benchmark code while continuing to reject conflicting account state. This measures current Zones contracts under the selected execution fork, not historical mainnet contract deployments.

Validation: all five runtime preparation tests (including full and partial synthetic T13 upgrades) and all three genesis-installer tests pass. Compiled Solidity artifacts match all six patched constants in Tempo's source byte for byte. Nightly Rust formatting, Clippy, shell syntax, and the GitHub Actions scan pass.

Both full-journey workflows passed, including SPF validation, and their published results were verified in ClickHouse:

- [T11](https://github.com/tempoxyz/zones/actions/runs/34402476910): 100/100 journeys, zero failures/timeouts; 1.84 journeys/sec, 6.00s median, 6.55s p95.
- [T12](https://github.com/tempoxyz/zones/actions/runs/34400146527): 100/100 journeys, zero failures/timeouts; 1.74 journeys/sec, 6.45s median, 7.06s p95.

T11 ran the final revision. T12 ran before the partial-upgrade handling refinement; the final script produces byte-for-byte identical Tempo source for that pinned revision. The runs used different runners and seeds, so these timings are validation results rather than a controlled fork performance comparison.

Prompted by: @Rjected

### DIFF
--- a/.github/workflows/zones-benchmark.yml
+++ b/.github/workflows/zones-benchmark.yml
@@ -614,6 +614,11 @@ jobs:
           git -C "$TEMPO_ROOT" apply --check "$GITHUB_WORKSPACE/$patch"
           git -C "$TEMPO_ROOT" apply "$GITHUB_WORKSPACE/$patch"
 
+      - uses: actions/setup-node@249970729cb0ef3589644e2896645e5dc5ba9c38 # v6
+        with:
+          node-version: "24"
+          package-manager-cache: false
+
       - name: Embed benchmark contracts in Tempo fork runtimes
         run: |
           set -euo pipefail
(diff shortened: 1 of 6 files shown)


0xKitsune (approved):

Rjected:
cyclops audit fast

PR #1428, opened 2026-09-09
fix(node): derive startup portal from genesis

decofe:
After regenesis, nodes can load the new genesis while retaining the previous `L1_PORTAL_ADDRESS` and `ZONE_ID` from infrastructure configuration. Derive both identities from the loaded genesis so restarting from zone 7 to zone 8 selects zone 8's native TIP-1091 portal without per-reset Argo overrides.

Use the derived portal for node startup, the checker, and CLI replay with `L1_HTTP_RPC_URL`. Accept legacy identity flags and environment values as deprecated, ignored compatibility inputs. Existing on-chain portal zone-ID, parent chain-ID, and initialization checks remain in place. Custom portal-address overrides are no longer supported by these CLI paths.

Extends [#1121](https://github.com/tempoxyz/zones/pull/1121), implements the [regenesis discussion](https://tempoxyz.slack.com/archives/C09K7873V4N/p1788984790095159), and supersedes the per-reset overrides in [workflows #1040](https://github.com/tempoxyz/workflows/pull/1040). Includes the replay derivation covered by [#1405](https://github.com/tempoxyz/zones/pull/1405), which overlaps this change.

Validation:
- All 22 CLI unit tests pass with `cargo +stable test -p zone-node --features cli --lib cli::tests --locked`.
- Startup regression covers consecutive zone IDs across mainnet, Moderato, and a custom development parent, with absent, matching, and stale legacy settings.
- Replay regression verifies the portal wired into the EVM's L1 state reader without network calls.
- Nightly formatting and `git diff --check` pass; conflicts with current main are resolved.
- No deployment or live regenesis was run. Infrastructure values are unchanged.

Prompted by: @rkrasiuk


0xKitsune:
cyclops audit fast

0xKitsune (approved):

PR #1430, opened 2026-09-10
feat(rpc): include L1 proofs in zone execution witnesses

klkvr:
Replaces `debug_zoneExecutionWitness`'s `tempo_reads` list with deduplicated L1 account/storage proof nodes, and adds the initial Tempo checkpoint header and Zone block identity. Proofs are fetched at the checkpoints actually read during replay, and the prover CLI and shadow prover consume the response directly instead of separately collecting L1 proofs. This gives #1369 a shared witness payload to persist; the node's configured L1 endpoint must support historical `eth_getMultiProof` requests.

Validation: formatting, whitespace checks, and scoped library Clippy (`zone-node`, `zone-evm`) pass, with an allowance for the existing `semicolon_in_expressions_from_non_local_macros` lint. The broader all-targets check was killed by signal 9 while compiling the node integration-test target. Tests were not run.


klkvr:
cyclops audit fast

shekhirin (approved):

PR #1433, opened 2026-09-11
feat(fees): enable dynamic Zone fees at Z1

0xKitsune:
This PR enables dynamic execution fees on Zones behind the `Z1` hardfork and makes the RPC fee surface follow the same policy.

Before `Z1`, Zones continue to produce blocks with a zero base fee and retain the existing public RPC defaults. Once `Z1` is active, Zones use Tempo's `T7` controller to derive the next base fee from the parent block's base fee and gas usage, including the `T7` floor and cap.

```rust
let fork = self.zone_hardfork_at(target_timestamp);
if fork.is_z1() {
    let parent_base_fee = parent
        .inner
        .base_fee_per_gas
        .expect("Zone blocks are expected to have a base fee");
    Some(tempo_t7_next_block_base_fee(
        parent_base_fee,
        parent.inner.gas_used,
    ))
} else {
    Some(0)
}
```

At `Z1`, `eth_gasPrice` returns only the latest base fee rather than adding Reth's sampled tip. `eth_fillTransaction` gives dynamic transactions base-fee headroom while defaulting an omitted priority fee to zero, consistent with `eth_maxPriorityFeePerGas`. Explicit priority fees are preserved.

```rust
let (gas_price, max_fee_per_gas) = if fork.is_z1() {
    (base_fee, base_fee.saturating_mul(2))
} else {
    (u128::from(TEMPO_T0_BASE_FEE), u128::from(TEMPO_T0_BASE_FEE))
};
apply_public_fee_policy(&mut request, gas_price, max_fee_per_gas);
```

`eth_feeHistory` continues to redact activity-derived fields while preserving actual base fees after `Z1`. When a historical range ends on the final `Z0` block and its canonical `Z1` successor is already available, the trailing fee now comes from that successor instead of being derived under the `Z0` rules.

The transaction pool keeps the protocol floor at zero so `Z0` transactions remain admissible across the `Z1` transition. Once `Z1` is active, the pool's current block base fee prevents underpriced transactions from being selected for blocks.

### DIFF
--- a/crates/chainspec/src/lib.rs
+++ b/crates/chainspec/src/lib.rs
@@ -214,8 +214,8 @@ impl EthChainSpec for ZoneChainSpec {
     }
 
     fn next_block_base_fee(&self, parent: &TempoHeader, target_timestamp: u64) -> Option<u64> {
-        let target_fork = self.zone_hardfork_at(target_timestamp);
-        if target_fork.is_z1() {
+        let fork = self.zone_hardfork_at(target_timestamp);
+        if fork.is_z1() {
             let parent_base_fee = parent
                 .inner
                 .base_fee_per_gas
--- a/crates/chainspec/src/lib.rs
+++ b/crates/chainspec/src/lib.rs
@@ -16,7 +16,10 @@ use std::{fmt::Display, sync::Arc};
 use tempo_chainspec::{
     TempoChainSpec, TempoConsensusSpec,
     hardfork::TempoHardfork,
-    spec::{DEV, TempoHardforks, chainspec_from_chain_id},
+    spec::{
+        DEV, TEMPO_T7_BASE_FEE_CAP, TempoHardforks, chainspec_from_chain_id,
+        tempo_t7_next_block_base_fee,
+    },
 };
 use tempo_primitives::TempoHeader;
 pub use zone_hardfork::ZoneHardfork;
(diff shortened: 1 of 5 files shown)


crates/node/src/rpc.rs
```diff
@@ -920,7 +939,15 @@ where
     }
 
     fn gas_price(&self) -> BoxFut<'_> {
-        Box::pin(async move { to_raw(&U256::from(TEMPO_T1_BASE_FEE)) })
+        Box::pin(async move {
+            let fork = self.zone_fork()?;
+            if fork.is_z1() {
+                let gas_price = EthFees::gas_price(&self.eth.api).await.map_err(internal)?;
```
0xKitsune: IIUC, the zero-priority-fee policy should still apply at Z1. `EthFees::gas_price` adds Reth's oracle-sampled tip, and the filler below uses the same oracle when `max_priority_fee_per_gas` is missing, even though this RPC returns zero from `eth_maxPriorityFeePerGas` and redacts fee-history rewards. That lets tipped history change client fees and expose recent tip activity. Could we return the live base fee directly and default a missing priority fee to zero, while still letting the filler derive the dynamic max fee? A Z1 redacted-RPC test with tipped history would cover both paths.

crates/node/src/rpc.rs
```diff
@@ -939,7 +966,7 @@ where
                     .await
                     .map_err(internal)?;
             // Redact gas fields (like `gas_used_ratio`) that can be used to guess tx counts
-            redact_fee_history(&mut history);
+            redact_fee_history(&mut history, self.zone_fork()?);
```
0xKitsune: At the activation boundary this preserves an incorrect trailing fee. Reth builds fee history's extra “next block” entry with the requested header's own timestamp; for a historical range ending at the last Z0 block, that selects Z0 and appends zero even though the canonical successor is Z1 with a nonzero fee. Could we use the successor timestamp/base fee when it exists and cover a range ending at the final Z0 block?

0xKitsune:
cyclops audit

crates/chainspec/src/lib.rs
```diff
@@ -213,8 +213,20 @@ impl EthChainSpec for ZoneChainSpec {
         self.inner.final_paris_total_difficulty()
     }
 
-    fn next_block_base_fee(&self, _parent: &TempoHeader, _target_timestamp: u64) -> Option<u64> {
-        Some(0)
+    fn next_block_base_fee(&self, parent: &TempoHeader, target_timestamp: u64) -> Option<u64> {
+        let fork = self.tempo_hardfork_at(target_timestamp);
+        if fork.is_t13() {
```
jenpaff: @0xKitsune should we push this back to a follow-up fork to keep separate from prover upgrade?

PR #1435, opened 2026-09-11
fix(fees): gate dynamic Zone fees at Tempo T13

decofe:
Gate dynamic Zone execution fees and the public RPC fee policy at Tempo T13 instead of Z1. Before T13, blocks keep a zero base fee and the RPC keeps its legacy defaults; at T13, the existing T7 fee controller and zero-tip RPC policy activate together using the inherited Tempo schedule.

Update the Tempo pin to [346c22e](https://github.com/tempoxyz/tempo/commit/346c22eb4293ac1f5edd27d4afd1f99323bc527c), which introduces T13 and moves TIP-1096 to that boundary. This moves off the v1.14 release pin onto the mainline T13 commit; the lower package version labels in Cargo.lock reflect mainline's version metadata. Public-network T13 activation remains unscheduled. Update the reference ZoneFactory installer and its existing runtime validation tests to use the renamed T13 contract runtimes.

Targets the branch for [#1433](https://github.com/tempoxyz/zones/pull/1433). Tests cover the T13 boundary, inherited activation, Z1 remaining fee-free before T13, RPC redaction, and sponsored fee settlement. The existing public-RPC test follows the DEV schedule's T13-at-genesis behavior.

Validation: all four CI test partitions (929 tests), the Docker build, nightly formatting, Clippy, and dependency checks pass. The network-chaos test `test_incoming_leader_recovers_after_l1_disconnect` passed on CI's automatic retry. All 13 chainspec tests and both reference ZoneFactory installer tests also pass locally. Automated review and Cyclops audit checks remain pending.

Prompted by: @0xKitsune

### DIFF
--- a/crates/node/src/rpc.rs
+++ b/crates/node/src/rpc.rs
@@ -43,7 +43,10 @@ use tempo_alloy::{
     provider::ext::TempoProviderExt as _,
     rpc::{TempoCallBuilderExt as _, TempoHeaderResponse, TempoTransactionRequest},
 };
-use tempo_chainspec::spec::{TEMPO_T0_BASE_FEE, TEMPO_T1_BASE_FEE};
+use tempo_chainspec::{
+    hardfork::TempoHardfork,
+    spec::{TEMPO_T0_BASE_FEE, TEMPO_T1_BASE_FEE, TempoHardforks},
+};
 use tempo_contracts::precompiles::{
     ACCOUNT_KEYCHAIN_ADDRESS,
     account_keychain::IAccountKeychain::{self, KeyInfo, getKeyCall},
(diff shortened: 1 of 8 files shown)


crates/node/src/rpc.rs
```diff
@@ -990,7 +993,7 @@ where
                 *trailing_fee = u128::from(successor_fee);
             }
             // Redact gas fields (like `gas_used_ratio`) that can be used to guess tx counts
-            redact_fee_history(&mut history, self.zone_fork()?);
+            redact_fee_history(&mut history, self.tempo_fork()?);
```
0xKitsune: IIUC, `eth_feeHistory` should keep returning the legacy public fee for pre-T13 blocks after the tip crosses T13. This checks the latest header once and then preserves the whole range, so the same historical query starts returning the stored zero fee after activation, and a range spanning T13 treats its pre-fork entries as post-fork. Could we select the policy per returned block, including the trailing successor fee, and add a cross-fork test?

PR #1437, opened 2026-09-14
fix(engine): carry zone timestamp throughout the import

shekhirin:
ref: SIGP-330

We should not re-invent the block timestamp when advancing the chain, because it can be T12 in `next_available()` and T13 in `advance_one` which will render the import incorrect https://github.com/tempoxyz/zones/blob/9f1fecd7974de924f970fd63dc7d443d40058680/crates/node/src/engine.rs#L150-L156

This PR fixes it by passing the timestamp that we used in `next_available` all the way down to `advance_one`.


klkvr (approved):

0xrusowsky (approved):
makes sense

PR #1443, opened 2026-09-15
ci: secure workflows with Aegis

grandizzy:
Adds Aegis Secure Runner to dependency-building workflows.

Hardens manual and operational builds with locked dependency resolution and pinned tool revisions.

PR #1444, opened 2026-09-15
fix(ci): restore Foundry builds in checkout and Docker

decofe:
## Summary

Foundry's dependency checks block CI in two places: the tracked lockfile points at an old `tempo-std` revision, and Docker copies that lockfile into a source snapshot without the Git metadata needed to validate it.

- Align `crates/contracts/foundry.lock` with the existing `tempo-std` submodule revision, `96f882963ce3b0e0abc7ec2b41d03d42457877bb`.
- Exclude that lockfile from Docker's build context alongside Git metadata. Docker still compiles the checked-out, pinned submodule sources; regular contract CI continues validating the tracked lockfile against Git.

This unblocks contract and Docker builds affecting [#1429](https://github.com/tempoxyz/zones/pull/1429) and [#1443](https://github.com/tempoxyz/zones/pull/1443).

## Validation

- Verified both Foundry lockfile entries match their committed submodule revisions.
- Reproduced Docker's `fatal: not a git repository` failure with its exact Foundry nightly (`ba4700106baf7c633ec664b27db381a3401cd097`) and a source snapshot without Git metadata.
- Excluding the lockfile from that snapshot makes the same `forge build --skip test` command pass (33 Solidity files).
- `git diff --check` passed.
- Full Docker validation runs in CI; the local sandbox has no Docker daemon.

Prompted by: @0xrusowsky


grandizzy (dismissed):

0xrusowsky (dismissed):

0xrusowsky:
cyclops audit super-fast

0xrusowsky (approved):

grandizzy (approved):

PR #1447, opened 2026-09-15
tests: Fix flaky network chaos test

adityapk00:
- Fix the flaky network chaos p2p tests


adityapk00:
cyclops audit super-fast

0xKitsune (approved):

crates/node/tests/it/network_chaos_p2p_e2e.rs
```diff
@@ -333,49 +333,214 @@ async fn resume_leadership_case(
...
+    // proposal reaching L1.
+    let verified_pre_handoff_height = closing_tips
+        .iter()
+        .map(|tip| tip.height)
+        .max()
+        .ok_or_else(|| eyre::eyre!("no connected nodes to observe during the outage"))?;
     eyre::ensure!(
-        settled_height <= a_fenced_height.max(follower_fenced_height),
-        "L1 settled isolated B's private proposal at zone height {settled_height}"
+        settled_height <= verified_pre_handoff_height,
```
0xKitsune: Optional: could we pin `blockHash()` alongside `zoneHeight()` and compare it with a connected node’s canonical hash at `settled_height`? This height-only check would still pass if the Portal accepted a different fork at an allowed height, while the later recovery check only establishes the eventual settled tip. Checking the hash here would make the outage safety assertion self-contained.

PR #1453, opened 2026-09-16
feat(node): derive L1 portal addresses from genesis

decofe:
Derive the L1 portal address from the zone ID encoded in genesis using Tempo's `zone_factory::portal_address`, for node services (including the checker), HTTP-backed re-execution, and the offline EVM fallback. Node startup no longer requires `--l1.portal-address` / `L1_PORTAL_ADDRESS`; supplied values are deprecated consistency checks and cannot override genesis.

Remove the explicit address from dev, Justfile, and benchmark node launchers. This enables infra to remove `l1PortalAddress` after rolling out the updated node; operator schemas and infra manifests are outside this PR. Direct deposit/admin tooling still accepts a target portal, and legacy EOA-deployed portals cannot be selected through node configuration.

Validation: [all four CI test shards passed](https://github.com/tempoxyz/zones/actions/runs/35096964821) (920 tests), including the new mainnet/Moderato/dev derivation and offline EVM regressions. The network-chaos high-L1-latency test passed on retry. [Clippy and Rust formatting passed](https://github.com/tempoxyz/zones/actions/runs/35096965369); local whitespace and benchmark shell syntax checks also passed.

Prompted by: @shekhirin


crates/node/src/cli.rs
```diff
@@ -104,19 +105,7 @@ fn run_node(mut cli: Cli<ZoneChainSpecParser, ZoneArgs>) -> eyre::Result<()> {
...
-                    )
-                })?
-                .parse()
-                .map_err(|error| eyre::eyre!("invalid L1_PORTAL_ADDRESS: {error}"))?;
-            eyre::ensure!(
-                !portal_address.is_zero(),
-                "L1_PORTAL_ADDRESS must be nonzero"
-            );
-            Some((url, portal_address))
+            Some(url)
```
0xKitsune: IIUC, `L1_PORTAL_ADDRESS` remains a consistency check during the migration. This removes the only command-independent read of that env var; `resolve_portal_address()` runs only in the node launcher, while `import`, `stage`, and `re-execute` invoke the components path directly. Those commands will now silently ignore a stale value and execute against the derived portal. Could we validate the optional legacy value in the component path too? A non-node regression test would make this contract explicit.

PR #1479, opened 2026-09-20
fix(prover): always witness the inbox token cursor

klkvr:
SPF reads `PROCESSED_ENABLED_TOKEN_COUNT` for every batch, but pre-T13 execution does not access it. If the settlement prover replays an unsettled pre-T13 batch after activation, the generated witness can lack the slot's non-membership proof and fail with `IncompleteStorageProof`.

Always include the Inbox token cursor in the explicit storage witness targets alongside EIP-2935 history reads. Execution-recorded values take precedence over the zero placeholder.

Validation: scoped `cargo +nightly clippy -p zone-node --lib --offline` passed on the original local checkout with existing warnings; formatting and `git diff --check` passed after applying the change to the latest `prover` branch. Tests were not run.

Addresses https://github.com/tempoxyz/zones/pull/1320#discussion_r4049686677


klkvr:
cyclops audit fast

0xrusowsky (approved):

shekhirin (approved):

